PR #1, opened 2019-01-02
Implement cairo stroke styles

cbrewster:
These aren't really usable yet as there is no way to get a `StrokeStyle` from the piet `RenderContext` generically yet.

I also had to add a context save/restore so that stroke styles for one path do not affect the styles of another.


raphlinus:
Thanks!

PR #4, opened 2019-01-04
WIP CoreGraphics backend

jrmuizel:
This isn't ready to land but I through it together to see what needed to be added to core-graphics crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,9 +2,10 @@
 
 members = [
     "piet",
-    "piet-cairo",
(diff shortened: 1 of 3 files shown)


raphlinus:
Very cool! It's early to start getting serious about new back-ends right now because I expect to iterate somewhat on the core API, but it's also excellent to get feedback early about whether the design has issues that would cause problems.

cmyr:
This has been merged as part of #176.

PR #5, opened 2019-01-05
Add basic font and text support

raphlinus:
This patch adds basic ability to draw text, supported by all three back-ends. More work is needed to specify font options, measure text, etc., but it's a start.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -1,10 +1,12 @@
 //! The Cairo backend for the Piet 2D graphics abstraction.
 
-use cairo::{Context, LineCap, LineJoin};
+use cairo::{
+    Context, FontFace, FontOptions, FontSlant, FontWeight, LineCap, LineJoin, Matrix, ScaledFont,
+};
 
 use kurbo::{PathEl, QuadBez, Shape, Vec2};
 
-use piet::{FillRule, RenderContext, RoundInto};
+use piet::{FillRule, Font, FontBuilder, RenderContext, RoundInto, TextLayout, TextLayoutBuilder};
 
 pub struct CairoRenderContext<'a> {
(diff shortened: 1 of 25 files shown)


piet-cairo/src/lib.rs
```diff
@@ -113,6 +131,40 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
...
+        &mut self,
+        layout: &Self::TL,
+        pos: impl RoundInto<Self::Point>,
+        brush: &Self::Brush,
+    ) {
+        self.ctx.set_scaled_font(&layout.font);
+        self.set_brush(brush);
+        let pos = pos.round_into();
+        self.ctx.move_to(pos.x, pos.y);
+        self.ctx.show_text(&layout.text);
```
jrmuizel: Worth adding a comment that this doesn't shape the text while the other backends do shape the text.

raphlinus: Indeed. At some point I'd like to blend in proper text layout, which is then likely to be useful with a number of back-ends, particularly lower level GPU-tuned ones. This is a bit of a hard problem, and I'm not sure what solution is best. But you're certainly right; as long as I'm using the "toy" text API of Cairo, that should be documented.

rylev:
@raphlinus Do you need assistance converting the progress here in piet-web to wasm-bindgen?

raphlinus:
@rylev Thanks for the offer, but the merge was pretty easy. You can double-check my work if you like.

piet-cairo/src/lib.rs
raphlinus: @jrmuizel I added a README, do you think that addresses your concern?

jrmuizel: Yeah. That looks good. You may want to mention the option of using font-kit to handle font discovery.

PR #6, opened 2019-01-06
Use wasm-bindgen instead of stdweb

rylev:
This is still a work in progress, but I wanted to see what you thought...

stdweb is fairly heavy handed. wasm-bindgen and the related web-sys crate provide the bare minimum for integrating Rust in a JavaScript and browser environment. These are the crates that the Rust Wasm working group is focused on. There is an [issue](https://github.com/rustwasm/team/issues/226) calrifying the differences. The stdweb maintainer hints at interest in eventually having stdweb be built on top of wasm-bindgen.  

The changes are fairly minimal, but it should set the project on a path of being as minimal as possible. The downside to this approach is that it makes the example a bit more involved to run since `cargo web` does not currently work with wasm-bindgen.

If you're interested in this approach I will get the example running.

### DIFF
--- a/piet-web/Cargo.toml
+++ b/piet-web/Cargo.toml
@@ -12,4 +12,8 @@ categories = ["rendering::graphics-api", "wasm"]
 kurbo = "0.1.2"
 piet = { path = "../piet" }
 
-stdweb = "0.4.12"
+wasm-bindgen = "0.2.29"
+
+[dependencies.web-sys]
+version = "0.3.6"
+features = ["CanvasRenderingContext2d", "CanvasWindingRule"]
--- a/piet-web/src/lib.rs
+++ b/piet-web/src/lib.rs
@@ -1,6 +1,7 @@
 //! The Web Canvas backend for the Piet 2D graphics abstraction.
 
-use stdweb::web::{CanvasRenderingContext2d, FillRule};
+use wasm_bindgen::JsValue;
+use web_sys::{CanvasRenderingContext2d, CanvasWindingRule};
 
 use kurbo::{PathEl, Shape, Vec2};
 
@@ -25,15 +26,15 @@ pub enum StrokeStyle {
     Default,
 }
 
-fn convert_fill_rule(fill_rule: piet::FillRule) -> FillRule {
+fn convert_fill_rule(fill_rule: piet::FillRule) -> CanvasWindingRule {
     match fill_rule {
-        piet::FillRule::NonZero => FillRule::NonZero,
-        piet::FillRule::EvenOdd => FillRule::EvenOdd,
+        piet::FillRule::NonZero => CanvasWindingRule::Nonzero,
+        piet::FillRule::EvenOdd => CanvasWindingRule::Evenodd,
     }
 }
 
 impl<'a> RenderContext for WebRenderContext<'a> {
-    /// stdweb doesn't have a native Point type, so use kurbo's.
+    /// wasm-bindgen doesn't have a native Point type, so use kurbo's.
     type Point = Vec2;
     type Coord = f64;
     type Brush = Brush;
@@ -47,15 +48,11 @@ impl<'a> RenderContext for WebRenderContext<'a> {
         Brush::Solid(rgba)
     }
 
-    fn fill(
-        &mut self,
-        shape: &impl Shape,
-        brush: &Self::Brush,
-        fill_rule: piet::FillRule,
-    ) {
+    fn fill(&mut self, shape: &impl Shape, brush: &Self::Brush, fill_rule: piet::FillRule) {
         self.set_path(shape);
         self.set_brush(brush, true);
-        self.ctx.fill(convert_fill_rule(fill_rule));
+        self.ctx
+            .fill_with_canvas_winding_rule(convert_fill_rule(fill_rule));
     }
 
     fn stroke(
@@ -94,9 +91,9 @@ impl<'a> WebRenderContext<'a> {
                     )
                 };
                 if is_fill {
-                    self.ctx.set_fill_style_color(&color_str);
+                    self.ctx.set_fill_style(&JsValue::from_str(&color_str));
                 } else {
-                    self.ctx.set_stroke_style_color(&color_str);
+                    self.ctx.set_stroke_style(&JsValue::from_str(&color_str));
                 }
             }
         }


raphlinus:
Very interesting! Yes, after reading the linked issue I agree that wasm-bindgen is a better fit for the goals of piet, though the current convenience of stdweb also has its value. I noticed that my hello example was 130k, which is not horrific but also not great. It seems likely that wasm-bindgen will push this firmly into the "great" category, which is exciting.

Basically the main thing I ask is that the instructions for building the demo are super clear. But you probably understand the importance of that. Thanks for working on this!

raphlinus:
One more thing might be worth mentioning: it's not yet clear to me whether web 2d canvas or svg is ultimately going to be more useful. Of course, an svg backend would make sense both in web and non-web contexts, in the latter case mostly spitting out a string (though some kind of dom-like tree would also be an option). But I don't want to start _too_ many backends until the core traits have stabilized :)

rylev:
@raphlinus Awesome! I pushed up an example. It not _quite_ as easy as before to get running but if you have npm already installed it's basically just as many steps. A possible advantage for some people is that it mirrors a bit more closely a flow you'd be used to if you're coming from the web world. 

I don't love that the cargo workspaces exception I had to put in the Cargo.toml file, but unfortunately I don't have a ton of experience with workspaces so I couldn't figure out a better way to do it in the limited time I had. 

The wasm file is 90kb smaller on my machine when compared to the previous example. I'm sure we could do better with some additional steps. 

Thoughts?

rylev:
This should be good to go. There's more we can do to make the lib a bit more flexible for users. With regards to code size, there's much more we can do depending on how willing we are to sacrifice some performance especially if there is little to no need for dynamic allocation. You can read more about that here: https://rustwasm.github.io/book/reference/code-size.html

piet-web/README.md
```diff
@@ -1,11 +1,11 @@
 # Running the examples
 
-Generally follow the directions at [stdweb].
+Ensure both cargo and [npm] are installed.
 
-`$ cargo install -f cargo-web`
+`$ cargo install -f wasm-bindgen`
```
raphlinus: I get this:

```
$ cargo install -f wasm-bindgen
    Updating crates.io index
  Installing wasm-bindgen v0.2.30
error: specified package has no binaries
```

I can diagnose this, but I have on my "hapless dev" hat. This is on a Windows 10 box in a git bash shell, and node (10.15.0) and npm are installed.

rylev: Whoops! It's actually `wasm-bindgen-cli`. Once it's installed it's just called `wasm-bindgen`. I will update this.

piet-web/examples/basic/build.sh
```diff
@@ -0,0 +1,15 @@
...
+# Build the `hello_world.wasm` file using Cargo/rustc
+cargo build --target wasm32-unknown-unknown
+
+# Run the `wasm-bindgen` CLI tool to postprocess the wasm file emitted by the
+# Rust compiler to emit the JS support glue that's necessary
+wasm-bindgen ../../../target/wasm32-unknown-unknown/debug/piet_web_example.wasm --out-dir basic-web-static/dist
+
+# Finally, package everything up using Webpack and start a server so we can
+# browse the result
+npm install --prefix basic-web-static
```
raphlinus: This also has a paper-cut error: works fine if I cd to basic-web-static and do `npm install` and serve from there, but not from the dir as shown in the batch file. But then it works!

rylev: Hmm... the --prefix option should take care of that. I don't need to cd into that directory on my machine. Maybe different versions of npm are causing this? 

I can change the script to cd into that directory instead. Do you think that will be a more robust solution? (ah... the joys of bash scripting ❤️)

raphlinus: The world of npm is foreign and strange to me, though I have played with it some. Looking at the [docs](https://docs.npmjs.com/misc/config#prefix), it looks like prefix sets the path where the artifacts get stored, but not where the package.json is read from. Did you try this from a fresh clone? (my inexperienced experience is that npm tends to litter a lot of files in your working directory)

I'm fine with it cd'ing into the directory. In general I would want to follow "best practices" but have no idea what those are. Not to be too nitpicky, but I think this is kinda important because a *lot* of people start projects by copying a hello example.

rylev: No worries on being nitpicky. I 100% that this should be painless for people. 

What you said about `--prefix` is true for global items (installed with `-g`) but for local cli commands this tells npm in which directory to run. From the docs: "If set on the command line, then it forces non-global commands to run in the specified folder."

I tried from a fresh clone and it worked fine for me. 

If you run `npm install --prefix basic-web-static` from that directory (the $PROJECT_ROOT/piet-web/examples/basic` directory), it warns that you don't have a `package.json`?

I try to avoid using cd in my bash scripts because I've ran into weird issues before, but if that's what's needed to get it working, we can do it.

raphlinus:
```
raphl@DESKTOP-6JFLCBE MINGW64 ~/github/piet/piet-web/examples/basic (rylev-wasm-bindgen)
$ npm -v
6.4.1
$ npm install --prefix basic-web-static
npm ERR! code ENOLOCAL
npm ERR! Could not install from "" as it does not contain a package.json file.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\raphl\AppData\Roaming\npm-cache\_logs\2019-01-07T17_39_10_724Z-debug.log
```

I'm not sure what's going on here, and totally understand wanting to minimize cd'ing.

rylev: Looks like a bug in the windows version of npm: https://npm.community/t/npm-prefix-flag-does-not-work-as-expected-on-windows/3147

I think moving to just using `cd` is the best choice.

raphlinus: Sounds good - I was just about to try it on my Mac, but seems like you've isolated it. That bug is scary, being auto-closed even though it seems valid and with the potential to affect lots of people.

rylev: That's unfortunately how the web world works. It's pretty fast and loose. The latest commit _should_ fix it (though I just ran into another way things could break that the readme now references).

rylev:
Awesome. FYI: I have _very_ little experience in graphics, but I hope I can learn through this repo. I'll try to be as helpful as possible where I can be.

PR #7, opened 2019-01-07
Text width

raphlinus:
This adds text width measurement, and is a reasonably quick followup to #5, but I thought it might be useful to have it as a separate PR.

### DIFF
--- a/piet-web/examples/basic/src/lib.rs
+++ b/piet-web/examples/basic/src/lib.rs
@@ -6,7 +6,7 @@ use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 use web_sys::{window, HtmlCanvasElement};
 
-use piet::{FillRule, FontBuilder, RenderContext, TextLayoutBuilder};
+use piet::{FillRule, FontBuilder, RenderContext, TextLayout, TextLayoutBuilder};
 use piet_web::WebRenderContext;
 
 fn draw_pretty_picture<R: RenderContext>(rc: &mut R) {
@@ -28,8 +28,16 @@ fn draw_pretty_picture<R: RenderContext>(rc: &mut R) {
 
     let font = rc.new_font_by_name("Segoe UI", 12.0).build();
     let layout = rc.new_text_layout(&font, "Hello piet-web!").build();
+    let w: f64 = layout.width().into();
     let brush = rc.solid_brush(0x80_00_00_C0);
     rc.draw_text(&layout, (80.0, 10.0), &brush);
+
+    rc.stroke(
+        &Line::new((80.0, 12.0), (80.0 + w, 12.0)),
+        &brush,
+        1.0,
+        None,
+    );
 }
 
 #[wasm_bindgen]
--- a/piet-web/examples/basic/src/lib.rs
+++ b/piet-web/examples/basic/src/lib.rs
@@ -57,7 +57,7 @@ pub fn run() {
         .dyn_into::<web_sys::CanvasRenderingContext2d>()
         .unwrap();
 
-    let dpr = window().map(|w| w.device_pixel_ratio()).unwrap_or(1.0);
+    let dpr = window().unwrap().device_pixel_ratio();
     canvas.set_width((canvas.offset_width() as f64 * dpr) as u32);
     canvas.set_height((canvas.offset_height() as f64 * dpr) as u32);
     let _ = context.scale(dpr, dpr);
--- a/piet-web/src/lib.rs
+++ b/piet-web/src/lib.rs
@@ -224,7 +224,7 @@ impl TextLayoutBuilder for WebTextLayoutBuilder {
         let width = self
             .ctx
             .measure_text(&self.text)
-            .map(|m| m.get_width())
+            .map(|m| m.width())
             .unwrap_or(0.0);
         WebTextLayout {
             font: self.font,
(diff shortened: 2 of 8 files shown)


raphlinus:
This one is annoying to review because of merges (I'm *still* not used to Github workflow, with something like Gerrit you'd just rebase and it'd look clean). The consequential commits are 
8b4a6c5 (main API) and ea4390b (web_sys fix).

rylev:
@raphlinus If you rebase against master and force push to this branch it should be clean.

piet-cairo/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Toy text API
+
+For simplicity, the back-end currently uses the [toy text API] in Cairo. Essentially, this means there is no shaping, so complex scripts won't render correctly at all, and Latin will be missing kerning, ligatures, and other refinements. According to the docs, "Any serious application should avoid them."
+
+Fairly soon, I hope to have some type of higher-level text in place. One possibility is [pango]. From what I can tell, this should work well on Linux, but since it has a non-optional glib dependency, it might be non-trivial to get it building portably. It's also not clear to me how well this approach handles discovering system fonts.
+
+Another possibility is to use HarfBuzz more directly, using the [rust-harfbuzz] bindings. This will require more work for font discovery, but has the possibility to be considerably more native.
+
+A third possibility is to adapt [libTXT] from Flutter. This is a state of the art text layout library, with considerable investment in making it work well on mobile. However, it is in C++ and thus at the very least will need nontrivial work to make good Rust bindings.
```
rylev: Should an issue be opened for discussing which text layout engine to use? I have questions about this, but this PR doesn't seem like the right place for this discussion.

piet-cairo/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+The need for text shaping will be common to many low-level renderers that are not supported by system text services, not just Cairo.
+
+## Building on non-Linux
+
+Cairo is quite portable, and it is quite feasible to build on other systems. However, the [cairo-rs] crate seems to expect a library to be provided, rather than building it from sources. On Windows, I've been using prebuilt binary releases from [cairo-windows].
+
+TODO: nicer installation instructions (contributions welcome)
```
rylev: This an the issue of cairo-rs portability also seem like good items to make issues out of. I'd be happy to help, but I want to make sure that that's how you'd like to run the project.

piet-web/examples/basic/src/lib.rs
```diff
@@ -44,8 +57,10 @@ pub fn run() {
         .dyn_into::<web_sys::CanvasRenderingContext2d>()
         .unwrap();
 
-    canvas.set_width(canvas.offset_width() as u32);
-    canvas.set_height(canvas.offset_height() as u32);
+    let dpr = window().map(|w| w.device_pixel_ratio()).unwrap_or(1.0);
```
rylev: nit: we already grab the window above and crash if it doesn't exist. We could just use that window and avoid having to `map`

piet-web/src/lib.rs
```diff
@@ -26,6 +26,29 @@ pub enum StrokeStyle {
     Default,
 }
 
+#[derive(Clone)]
+pub struct WebFont {
+    family: String,
+    weight: u32,
+    // Maybe this should be enum? Everything is stringly typed...
+    style: &'static str,
```
rylev: Is this meant to model this? https://developer.mozilla.org/en-US/docs/Web/CSS/font-style If so, I think an enum is better. Since JavaScript doesn't have first class enums, generally what would be modeled as an enum in Rust (or even C) is modeled through strings. I think having a policy of exposing these as enums is a good idea.

piet/src/render_context.rs
```diff
@@ -34,6 +45,12 @@ pub trait RenderContext {
     /// Parameters for the style of stroke operations.
     type StrokeStyle;
 
+    type FBuilder: FontBuilder<Out = Self::F>;
+    type F: Font;
```
rylev: Any reason this type is just `F` instead of `Font`? I know `Font: Font` looks weird, but I imagine having just `F` will be confusing down the line and will make reading code harder than it needs to be.

raphlinus: I think you're right. I wasn't sure whether there would be namespace collisions, but calling both Font does make sense. I think that's probably out of scope for this PR though, happy to take the rename in another one.

piet-web/src/lib.rs
raphlinus: Agreed. Would it be better to do an `as_string()` method on the enum, or impl std::fmt::Display on it?

piet-cairo/README.md
raphlinus: Feel more than free to open issues, especially for things you want to help with.

piet-cairo/README.md
raphlinus: Yes, another issue for that is a great idea. Let's discuss it there.

piet-web/src/lib.rs
rylev: The docs say "Display is for user-facing output" which I would not classify this as. `as_string` could work. Another option, since I believe this is only meant to be used in one place is to just have a function `font_str` which does a match on the style.

raphlinus: I'll go with that then.

PR #13, opened 2019-01-08
Travis

rylev:


rylev:
@raphlinus Windows doesn't currently build because (I believe) of the issue with cairo-rs on Windows. Travis Windows VM start times are also incredibly slow (upwards of 10 minutes) which is unfortunate. If the Linux and OSX versions pass, let's merge and we can figure out Windows in another PR.

Issue #17, opened 2019-01-10
Basic Testing


rylev:
What would a basic testing infrastructure look like. Due to the nature of the code and the need for rapid iteration, unit testing won't get us very far. 

Can we set up some sort of testing framework where simple graphics programs built against the latest version of piet and then diffed with known good images. This could work with the direct-2d and cairo backends but is problematic with a web based backend. How robust would this testing be? Would it easily miss real issues or report failures on things that aren't actually failures?

jrmuizel:
I'd recommend trying to avoid this kind of reference image style testing. This style of testing tends to be very sensitive to the environment that it's running in. i.e. different versions of Direct2D can produce different images. Updating the reference images can also be quite tedious when things change especially if there are lots of different platforms. 

I've had more success with reftest style tests where you draw things two different ways and make sure they're equal, not equal, or equal with some threshold. We've used this with Gecko and WebRender with pretty good success. 

Here's the webrender implementation: https://github.com/servo/webrender/blob/master/wrench/src/reftest.rs

PR #21, opened 2019-01-12
Refine Shape ownership; refactor test pictures

raphlinus:
Allow both owned and borrowed Shape parameters, following changes in kurbo.

Reduce duplication of test picture code by putting it in a separate crate. Over time, it is expected that crate will grow test and benchmark functionality, in addition to just being a repository of pictures.

Partially addresses #17


rylev:
Looks great!

Issue #20, opened 2019-01-12
Switch to Result based error handling


rylev:
Currently the codebase is full of unwraps and only assumes a happy path. We should soon switch to a real error handling using `Result`.

PR #25, opened 2019-01-15
Add images

raphlinus:
This patch adds image capability, including support in all back-ends. Three image formats are intially supported; RGB 24 bits per pixel, and RGBA with both premultiplied and separate alpha. We will likely want to revisit this set, as some combinations of format and back-end may reduce the amount of conversion needed. In addition, there should be some mechanism (whether compile time or run time) to signal to drawing code which formats are most efficient. For example, premultiplied alpha is the most efficient on Cairo and Direct2D back ends, but separate alpha is dramatically more efficient on Web, as unpremultiplying requires 3 divides per pixel, and SIMD is not (at present) available.

The web implementation uses a scratch HTML Canvas to hold the image. That's most compatible when the source is an RGBA buffer and for general transform/clip, but direct ImageData could be more efficient for unity scale and clip to hard rectangle, and ImageBitmap could be useful for letting the browser do the decoding (though ImageBitmap has poorer compatibility and is async).

Images have many error conditions, so the need for #20 was strongly felt. That will happen soon.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -1,12 +1,16 @@
 //! The Cairo backend for the Piet 2D graphics abstraction.
 
 use cairo::{
-    Context, FontFace, FontOptions, FontSlant, FontWeight, LineCap, LineJoin, Matrix, ScaledFont,
+    Context, Filter, FontFace, FontOptions, FontSlant, FontWeight, Format, ImageSurface, LineCap,
+    LineJoin, Matrix, Pattern, PatternTrait, ScaledFont, SurfacePattern,
 };
 
-use kurbo::{Affine, PathEl, QuadBez, Shape, Vec2};
+use kurbo::{Affine, PathEl, QuadBez, Rect, Shape, Vec2};
 
-use piet::{FillRule, Font, FontBuilder, RenderContext, RoundInto, TextLayout, TextLayoutBuilder};
+use piet::{
+    FillRule, Font, FontBuilder, InterpolationMode, RenderContext, RoundInto, TextLayout,
+    TextLayoutBuilder,
+};
 
 pub struct CairoRenderContext<'a> {
     // Cairo has this as Clone and with &self methods, but we do this to avoid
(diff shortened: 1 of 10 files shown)


piet-cairo/src/lib.rs
```diff
@@ -185,6 +191,92 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
...
+        image
+    }
+
+    fn draw_image(
+        &mut self,
+        image: &Self::Image,
+        rect: impl Into<Rect>,
+        interp: InterpolationMode,
+    ) {
+        self.ctx.save();
```
rylev: Any reason to not use `with_save`?

piet-direct2d/src/lib.rs
```diff
@@ -361,6 +387,79 @@ impl<'a> RenderContext for D2DRenderContext<'a> {
...
+                for i in 0..width * height {
+                    new_buf[i * 4 + 0] = buf[i * 3 + 0];
+                    new_buf[i * 4 + 1] = buf[i * 3 + 1];
+                    new_buf[i * 4 + 2] = buf[i * 3 + 2];
+                }
+                Cow::from(new_buf)
+            }
+            ImageFormat::RgbaSeparate => {
+                let mut new_buf = vec![255; width * height * 4];
+                // TODO (performance): this would be soooo much faster with SIMD
```
rylev: I imagine this is not the only place in this PR where this is true.

piet-web/src/lib.rs
```diff
@@ -2,20 +2,25 @@
...
-use piet::{Font, FontBuilder, RenderContext, RoundInto, TextLayout, TextLayoutBuilder};
+use piet::{
+    Font, FontBuilder, ImageFormat, InterpolationMode, RenderContext, RoundInto, TextLayout,
+    TextLayoutBuilder,
+};
 
 pub struct WebRenderContext<'a> {
     ctx: &'a mut CanvasRenderingContext2d,
+    /// Used for creating image bitmaps and possibly other resources.
+    window: &'a Window,
```
rylev: My gut reaction to this was "wow that's taking a large dependency". I think this is something to keep an eye on for the future to see if there's a better way

raphlinus: 100%. I'm willing to consider making this a feature, and having some image operations fail. I also feel that this is one of many weaknesses in the web canvas spec - it can't decide whether it's a lightweight JS API (its origins from Apple) or deeply bound into the whole web stack (certainly the new ImageBitmap mechanism).

PR #28, opened 2019-01-19
Add dash handling to web

rylev:


piet-web/Cargo.toml
```diff
@@ -16,8 +16,8 @@ kurbo = "0.2.0"
 piet = { path = "../piet" }
 
 wasm-bindgen = "0.2.30"
-
+js-sys = "0.3.10"
 [dependencies.web-sys]
 version = "0.3.6"
```
SimonSapin: This needs to be 0.3.10, right?

piet-web/src/lib.rs
```diff
@@ -366,7 +367,25 @@ impl<'a> WebRenderContext<'a> {
         let miter_limit = style.and_then(|style| style.miter_limit).unwrap_or(10.0);
         self.ctx.set_miter_limit(miter_limit);
 
-        // TODO: line dashes. Waiting on set_line_dash in web_sys.
+        let (dash_segs, dash_offset) = style
+            .and_then(|style| style.dash.as_ref())
+            .map(|dash| {
+                let len = dash.0.len() as u32;
+                let array = Float64Array::new_with_length(len);
+                for (i, elem) in dash.0.iter().enumerate() {
```
SimonSapin: I think this view can be avoided entirely by using the `Float64Array::view` constructor.

piet-web/Cargo.toml
rylev: hm probably, but it seems to work without the upgrade. Strange...

piet-web/src/lib.rs
rylev: Yea although I wanted to avoid unsafe and `Float64Array::view` is unsafe since it's simply a view into unmanaged memory without any guarantees that that memory will stay valid. I think in this case we can assume it will stay valid, but I decided to go with the safe route first. I could leave a comment saying it's something we might want to look into.

piet-web/Cargo.toml
SimonSapin: If you don’t have a `Cargo.lock` (like on CI), Cargo will pick the latest. It’s in a larger project that keeps its lock file in the repository that accurate dependencies are more useful.

piet-web/src/lib.rs
SimonSapin: That’s fair.

My reasoning is that `setLineDash` is [defined in WebIDL](https://html.spec.whatwg.org/multipage/canvas.html#2dcontext:dom-context-2d-setlinedash) to take a parameter of type `sequence<unrestricted double>`. This type is [defined](https://heycam.github.io/webidl/#es-sequence) to be created from any iterable ES/JS value by iterating it.

So as far I understand, the `Float64Array` value only needs to stay valid for the duration of the `setLineDash` call. It won’t be stashed away somewhere (moved, in Rust terms) to be iterated accessed later.

It’s true that this Rust code needs to be careful to avoid something like:

1. Call `Float64Array::view` with a `&[u64]` view of the contents of a `Vec<f64>`.
2. Then drop the `Vec`
3. Then iterate or otherwise access the `Float64Array`. This is a use-after-free.

However we’re in a function that takes a `style: Option<&StrokeStyle>` parameter, so the `StrokeStyle` value is known to outlive the call.

raphlinus: I would prefer to avoid unsafe, but am not dogmatic about it. If we do adopt unsafe, I would ask for an empirical measurement of the gain - in this case, I would strongly suspect it would be in the noise compared with the rest of the FFI overhead and the cost of the gfx operation itself.

I should add, thanks for adding this and sweating the details! I'm happy with this going in as soon as there's consensus.

rylev: I think we have consensus that there is a more performant way to do this, but that that performance gain is probably not worth the overhead. We can revisit this in the future once we look into performance optimizations more closely.

PR #35, opened 2019-01-29
New crate for selecting default back-end

raphlinus:
This patch creates a new piet-common crate that selects a default piet back-end, and also provides much of the boilerplate for setting it up so that a raw bitmap can be extracted.

The common crate currently works for direct2d and cairo back-ends. I might try to get a web back-end into this PR, or do it as a follow-on.

This PR also sets versions to 0.0.1 in preparation for the first crates.io release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@
 members = [
     "piet",
     "piet-cairo",
+    "piet-common",
     "piet-direct2d",
     "piet-test",
     "piet-web",
--- a/piet-cairo/Cargo.toml
+++ b/piet-cairo/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "piet-cairo"
-version = "0.1.0"
+version = "0.0.1"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 description = "Cairo backend for piet 2D graphics abstraction."
 license = "MIT/Apache-2.0"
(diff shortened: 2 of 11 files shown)


SimonSapin:
Out of curiosity (not at all an objection to the PR), what is the scope of this new crate? I see that it introduces the concept of "device" which doesn’t exist in the `piet` crate, but so far this can only be used to obtain a buffer of rendered pixels. Should there eventually be in `piet_common` a way to render "on screen"? Is window management (on platforms where that’s relevant) in scope?

Is code that uses Piet generally expected to be generic over the backend type? What is the way to deal with backend-specific details? For example, I’m considering a PDF backend that would have multiple pages (drawing surfaces), each with a given size.

raphlinus:
@SimonSapin This is a tricky question, discussed a bit in https://github.com/xi-editor/druid/issues/8. I think the short answer is this: I really like the idea of generic drawing code, but because the RenderContext is not object-safe, the `Box<Widget>` type in druid must choose a single concrete type. That's basically what this does.

I think if there's a library to, say, plot a graph, that should be generic, and it loses its generic-ness at the point where it's boxed.

If we get to the point where we really need multiple back-ends from boxed drawing code, I envision an enum.

This is not set in stone - we're doing advanced Rust typery here, and regularly running into its limits, so we want to be flexible. But it represents what I think is the best compromise.

raphlinus:
To answer more of your question. I think a PDF back-end is a great idea. Of course, things are a bit fluid right now so there's some risk of churn, but hopefully they are settling down soon.

The question of how to access back-end specific features is a very very good one. What I'm most likely to do is have a [`to_any`](https://github.com/raphlinus/synthesizer-io/blob/master/synthesizer-io-core/src/module.rs#L84) method like I've done in the synthesizer `Module` trait. Then you can downcast to the PDF renderer and call any methods on that you like. It's clear there needs to be some mechanisms for loose coupling, but I want to be deliberate about introducing them.

Issue #36, opened 2019-01-29
Text::Coord v.s. RenderContext::Coord?


SimonSapin:
The `Text` and `RenderContext` traits both have `Coord` and `TextLayout` associated types. The former also has a `type Text: Text<TextLayout = Self::TextLayout>;` associated type, with an equality that ensures that the two `TextLayout` types are the same.

But there is no similar constraint to ensure that the two `Coord` types are the same. Is this deliberate?

raphlinus:
I thought about this, and think it's ok. For one, all the text stuff is probably going to be split off into separate modules, so looser coupling is a good thing. For two, right now I'm not seeing a use case where you really need to bridge the two. The associated type exists largely as a carrier for an `Into` to make it convenient to use either `f32` or `f64`, and avoid double conversions when both the caller and callee use the same type.

raphlinus:
Closing this as obsolete, we've basically moved off associated types for coordinates.

PR #37, opened 2019-01-29
piet-common: Add type aliases for the associated types

SimonSapin:
… which are otherwise verbose to name. This hopefully makes code like https://github.com/xi-editor/druid/pull/11#discussion_r251832951 easier to write.

The use of `'static` is somewhat dubious, but I think that picking *an* arbitrary lifetime might be necessary (unless the type aliases themselves gain a lifetime parameter) because, on principle, the associated types in `impl<'a> RenderContext for SomeType<'a>` *could* use `'a`. But they don’t, for backends that `piet-common` supports, so picking *any* lifetime sounds ok.

Another question is: should there be a `TextCoord` type alias? CC https://github.com/linebender/piet/issues/36

### DIFF
--- a/piet-common/src/lib.rs
+++ b/piet-common/src/lib.rs
@@ -3,13 +3,13 @@
 pub type Point = <Piet<'static> as piet::RenderContext>::Point;
 pub type Coord = <Piet<'static> as piet::RenderContext>::Coord;
 pub type Brush = <Piet<'static> as piet::RenderContext>::Brush;
-pub type Text = <Piet<'static> as piet::RenderContext>::Text;
+pub type Text<'a> = <Piet<'a> as piet::RenderContext>::Text;
 pub type TextLayout = <Piet<'static> as piet::RenderContext>::TextLayout;
 pub type Image = <Piet<'static> as piet::RenderContext>::Image;
 
-pub type FontBuilder = <Text as piet::Text>::FontBuilder;
-pub type Font = <Text as piet::Text>::Font;
-pub type TextLayoutBuilder = <Text as piet::Text>::TextLayoutBuilder;
+pub type FontBuilder<'a> = <Text<'a> as piet::Text>::FontBuilder;
+pub type Font = <Text<'static> as piet::Text>::Font;
+pub type TextLayoutBuilder<'a> = <Text<'a> as piet::Text>::TextLayoutBuilder;
 
 #[cfg(any(
     feature = "cairo",
--- a/piet-common/src/lib.rs
+++ b/piet-common/src/lib.rs
@@ -1,5 +1,16 @@
 //! Selection of a common back-end for piet.
 
+pub type Point = <Piet<'static> as piet::RenderContext>::Point;
+pub type Coord = <Piet<'static> as piet::RenderContext>::Coord;
+pub type Brush = <Piet<'static> as piet::RenderContext>::Brush;
+pub type Text = <Piet<'static> as piet::RenderContext>::Text;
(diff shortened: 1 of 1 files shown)


raphlinus:
So this gets into interesting territory. If were not actually using the lifetime in the associated types, maybe they should be removed in piet. I'm not clear yet on whether we are or not. The lifetime in `Piet<'a>` represents the lifetime of the "factory" (concretely direct2d::Factory and also directwrite), which is assumed to be borrowed for the lifetime of the drawing operations. I'm trying to encode the constraint that a resource that's created can't usefully be used past the lifetime of its factory, but as you note I'm not really enforcing that. I'm not sure whether it's useful to try to express. If not, it would probably be best to take the lifetime out of the associated types. If so, it's probably best to add a `PhantomData`, at which point having `'static` here wouldn't work well.

I'll think about this a bit.

SimonSapin:
Ah, so you might want to change the concrete associated types to also have a lifetime parameter in the future? In that case the type aliases should have a parameter as well, and their definition use it instead of `'static`.

raphlinus:
I think more experimentation and thought is needed, but my current thinking is that the builders will need the lifetime parameter (because in general they contain references to factories), but the resources won't.

raphlinus:
And to be clear what the "experimentatation" I have in mind is: writing functions that take these associated types and do something with them (ie build a text layout, then draw to a RenderContext). As written above, I suspect the builder types will need lifetimes. Either way, any such lifetimes are added and the experiment succeeds, I'd be more than happy to merge.

SimonSapin:
I think that whether the aliases have a lifetime parameter is more about the implementations of the traits than its users. For example in this code:

https://github.com/linebender/piet/blob/ae79ff1f71b02d1341d96255fb5b3627952114bf/piet-cairo/src/lib.rs#L107-L113

The `Text` associated type does not use the `'a` lifetime that parameterizes the trait, but it could. Another impl might look like:

```rust
impl<'a> RenderContext for CairoRenderContext<'a> {
    type Text = &'a mut CairoText;
}
```

Or:

```rust
impl<'a> RenderContext for CairoRenderContext<'a> {
    type Text = CairoText<'a>;
}
```

But last I looked none of the backends supported by `piet-common` do that, so `<CairoRenderContext<'static> as RenderContext>::Text` is the same as `<CairoRenderContext<'a> as RenderContext>::Text` for any `'a`.

raphlinus:
I didn't publish my test code, but when I had the lifetime in `Text` as `'static`, it caused borrowing errors, and it works fine with the current state of the PR. I'd have to redo the example, I lost it when I was doing some merging.

SimonSapin:
Oh, hmm, maybe there’s a bug/limitation in the compiler where it fails to unify `<CairoRenderContext<'static> as RenderContext>::Text` and `<CairoRenderContext<'a> as RenderContext>::Text` even if they happen to be the same type?

raphlinus:
Ok, I tested this again, and I think I was mistaken - I think my original error was trying to use `piet::Text` instead of `piet_common::Text`, and the former was being interpreted as a dyn Trait. That gave lifetime errors among other things. So I think I want to revert my changes.

That is bringing up another issue - you can't import both `piet::Text` and `piet_common::Text` without namespace collisions, and you do need both to use the trait methods. Same for the builder types, but it's less likely you need to explicitly name those. I'm not sure what the best approach is.

raphlinus:
Closing this as it's subsumed by #74. Thanks for raising the issue, the discussion was very useful!

PR #39, opened 2019-02-06
Add gradients

raphlinus:
Implementations for all 3 back-ends, with very basic testing in picture_4.

### DIFF
--- a/piet/src/gradient.rs
+++ b/piet/src/gradient.rs
@@ -1,0 +1,50 @@
+//! Gradient specifications.
+
+use kurbo::Vec2;
+
+/// Specification of a gradient.
+#[derive(Clone)]
+pub enum Gradient {
+    /// A linear gradient.
+    Linear(LinearGradient),
+    /// A radial gradient.
+    Radial(RadialGradient),
+}
+
+/// Specification of a linear gradient.
+#[derive(Clone)]
+pub struct LinearGradient {
+    /// The start point (corresponding to pos 0.0).
+    pub start: Vec2,
+    /// The end point (corresponding to pos 1.0).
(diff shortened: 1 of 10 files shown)


raphlinus:
The gradient functionality is very much a common denominator of the back-ends, and there are features in each that could be added (possibly with difficulty in the less capable back-ends). So one of the main things I'm looking for in this PR is feedback on whether the functionality meets the needs of clients, particularly SVG.

piet/src/gradient.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub stops: Vec<GradientStop>,
+}
+
+/// Specification of a gradient stop.
+#[derive(Clone)]
+pub struct GradientStop {
+    /// The coordinate of the stop.
+    pub pos: f32,
+    /// The color at that stop.
+    pub rgba: u32,
```
RazrFalcon: Do we have a piet-specific color type?

raphlinus: No, but I'm open to adding one. It's a bit of a tricky area, because colors can become *very* complicated (CMYK and higher dimensional color spaces for print, HDR, etc). But I want the default path to be super-simple, and standardizing on 32-bit rgba does that for me.

RazrFalcon: I'm not sure we need something more complicated than a simple RGBA. `cairo` doesn't support this for sure.

But a simple number is too error-prone.

raphlinus: Fair enough. I think that should be a separate issue, as it affects creation of a solid brush, etc. I'll file an issue.

raphlinus:
The CI failures are spurious :/

PR #48, opened 2019-06-16
Use Color to represent colors

### LINKED ISSUE #40: A simple color type
Discussion branched from review of #39.

Use of a raw `u32` is problematic to represent color. We should have a simple color type.

One model is [`ColorF`](https://docs.rs/direct2d/0.2.0/direct2d/math/struct.ColorF.html) from the direct2d crate, which wraps [D2D_COLOR_F](https://docs.microsoft.com/en-us/windows/desktop/direct2d/d2d-color-f) but has its own conversions. I think having a `From` conversion from `u32` interpreted as rgba (note the ColorF conversion is from rgb) would be good. I'd also be open to a conversion from (u32, f32), interpreted as (rgb, a); this latter conversion would be the same as ColorF. And APIs that directly accept color (`solid_brush`) is the main thing, should accept `impl Into` so any of these variants can be used.

I'll note that the concept of color is extremely deep water. There are questions of colorspace ([sRGB](https://en.wikipedia.org/wiki/SRGB) vs other standard RGB color spaces vs the idea of "device color"), gamut, depth (including HDR), use of color for print (CMYK and other spaces that include spot color), other appearance measurements such as gloss, etc. I don't want to get into any of that right now. The new color type would be, as much as specified, an 8 bit sRGB value with separate alpha. I can imagine adding more complex color types later as *optional* extensions.

Bikeshedding regarding naming and other similar questions are welcome.

raphlinus:
Fixes #40

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -52,12 +52,8 @@ impl Color {
     }
 
     /// Opaque white.
-    pub const fn white() -> Color {
-        Color::Rgba32(0xff_ff_ff_ff)
-    }
+    pub const WHITE: Color = Color::Rgba32(0xff_ff_ff_ff);
 
     /// Opaque black.
-    pub const fn black() -> Color {
-        Color::Rgba32(0x00_00_00_ff)
-    }
+    pub const BLACK: Color = Color::Rgba32(0x00_00_00_ff);
 }
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -1,0 +1,63 @@
+//! A simple representation of color
+
+/// A datatype representing color.
+///
+/// Currently this is only a 32 bit RGBA value, but it will likely
+/// extend to some form of wide-gamut colorspace, and in the meantime
+/// is useful for giving programs proper type.
+#[derive(Clone)]
+pub enum Color {
+    Rgba32(u32),
+}
+
+impl Color {
+    /// Create a color from a 32-bit rgba value (alpha as least significant byte).
+    pub fn rgba32(rgba: u32) -> Color {
+        Color::Rgba32(rgba)
+    }
+
+    /// Create a color from a 24-bit rgb value (red most significant, blue least).
+    pub fn rgb24(rgb: u32) -> Color {
+        Color::rgba32((rgb << 8) | 0xff)
+    }
+
+    /// Create a color from four floating point values, each in the range 0.0 to 1.0.
+    ///
+    /// The interpretation is the same as rgba32, and no greater precision is
+    /// (currently) assumed.
+    pub fn rgba<F: Into<f64>>(r: F, g: F, b: F, a: F) -> Color {
+        let r = (r.into().max(0.0).min(1.0) * 255.0).round() as u32;
+        let g = (g.into().max(0.0).min(1.0) * 255.0).round() as u32;
+        let b = (b.into().max(0.0).min(1.0) * 255.0).round() as u32;
+        let a = (a.into().max(0.0).min(1.0) * 255.0).round() as u32;
+        Color::rgba32((r << 24) | (g << 16) | (b << 8) | a)
+    }
+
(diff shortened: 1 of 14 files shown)


raphlinus:
A discussion question: should I make solid brush creation non-failable while I'm breaking the API, or is that a separate change? (I'd also want to make text format creation significantly less result-oriented)

RazrFalcon:
>should I make solid brush creation non-failable

I think you should.

piet/src/color.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Convert a color value to a 32-bit rgba value.
+    pub fn as_rgba32(&self) -> u32 {
+        match *self {
+            Color::Rgba32(rgba) => rgba,
+        }
+    }
+
+    /// Opaque white.
+    pub const fn white() -> Color {
```
cmyr: glad that these are `const fns`. I might go a bit further, though, and add them ass associated constants on the struct itself; `Color::WHITE`, Color::Black`, etc.

piet/src/render_context.rs
```diff
@@ -87,13 +89,15 @@ pub trait RenderContext {
     /// responsiblity? We could have a cache that is flushed when the Direct2D
     /// render target is rebuilt. Solid brushes are super lightweight, but
     /// other potentially retained objects will be heavier.
-    fn solid_brush(&mut self, rgba: u32) -> Result<Self::Brush, Error>;
+    fn solid_brush(&mut self, color: Color) -> Self::Brush;
```
cmyr: What if we removed this API, and just had `Impl Into<Brush> for Color`? I could imagine needing to expose something for gradients (although possibly also not) but I'm not sure that this API is pulling its weight. Is brush creation expensive? Even in that case, we're rarely reusing brushes between draw calls or anything...

.travis.yml
```diff
@@ -6,11 +6,16 @@ os:
...
+  - 1.35.0
   - stable
   - beta
   - nightly
 
+# There's a build problem on macOS, we hope this will be fixed.
+# See https://github.com/gtk-rs/cairo/issues/263
+env:
+  - PKG_CONFIG_PATH=/usr/local/opt/libffi/lib/pkgconfig
+
```
cmyr: needs a rebase?

piet/src/color.rs
raphlinus: Cool, that's been stable since 1.20, it's hard to keep track :)

piet/src/render_context.rs
raphlinus: I have a branch where I'm trying to do this, but it turns out to be not anywhere nearly so easy; the [`SolidColorBrush::create`](https://docs.rs/direct2d/0.2.0/direct2d/brush/struct.SolidColorBrush.html#method.create) method requires a RenderTarget (and there's some funny stuff about the brush being bound to this render target, but I think I can ease that by using a [`DeviceContext`](https://docs.rs/direct2d/0.2.0/direct2d/device_context/struct.DeviceContext.html) everywhere - I tested and this does work on Windows 7 SP1).

I did try making an IntoBrush associated trait that would take either an `&Brush` or a `Color`, but the type tetris got a bit much for me (the result should be a Cow to handle both cases, and needs to encode the lifetime).

Brush creation in D2D is a COM call that does an allocation, so it's not free. It does feel clunky though; a completely reasonable alternative is for the Rust type to be cheap and then have a cache to D2D brush objects. This goes a bit beyond my original intent of being a thin wrapper over D2D, but maybe in this case the simplification for users is worth it.

.travis.yml
raphlinus: Probably should have done a rebase instead of a merge, but then it needs a force-push. In any case, it should turn out the same. Ah, git.

piet/src/render_context.rs
cmyr: Okay, sounds like future work.

PR #52, opened 2019-06-18
Reexport kurbo from piet crate

cmyr:
This ensures that all of the piet family crates are depending on
a single version of kurbo, which is listed as a dependency in
exactly one `Cargo.toml`.

This was causing me a major headache when I wanted to test out some local changes I'd made to kurbo.

If there are any problems with this approach I'm curious to hear what they are, I couldn't think of anything.

### DIFF
--- a/piet-cairo/Cargo.toml
+++ b/piet-cairo/Cargo.toml
@@ -9,7 +9,6 @@ keywords = ["graphics", "2d"]
 categories = ["rendering::graphics-api"]
 
 [dependencies]
-kurbo = "0.2.1"
 piet = { version = "0.0.2", path = "../piet" }
 
 [dependencies.cairo-rs]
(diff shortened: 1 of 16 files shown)


cmyr:
I also don't understand the distinction between the `piet` and `piet-common` crates. Afaict `piet-common` could just as easily be a submodule, like `piet::backend` or something. 

If there is something I'm missing, and these really should be distinct, then what if `piet` were renamed `piet-core`, and was reexported  by `piet-common` (which (I know, I know) could be renamed `piet`)? 

My thinking is that I don't see any situation where people want one of these crates without the other, and the various different imports and dependencies makes everything very fragile.

raphlinus:
The short answer to the comment above (for posterity) is that piet-common is much simpler to use but is limited to the case where there's only one backend, selected at compile time. Using piet directly allows the use of multiple backends but requires significantly more type tetris. As discussed privately, I'm open to making this more robust and less confusing, but it's not immediately obvious to me how to do so. There's discussion on the zulip in the "Platform independence" thread.

PR #56, opened 2019-06-19
Setup azure pipelines for ci

cmyr:
I couldn't get cairo installed on windows, but otherwise this is working? 🤷‍♂

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,32 @@
+pr: ["master"]
+
+jobs:
+  - job:
+    strategy:
+      matrix:
+        mac:
+          imageName: 'macos-10.13'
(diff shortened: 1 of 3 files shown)


.azure-pipelines.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  - job:
+    strategy:
+      matrix:
+        mac:
+          imageName: 'macos-10.13'
+        windows:
+          imageName: 'vs2017-win2016'
+    pool:
```
raphlinus: Any special reason there's no Linux build?

cmyr: because I copied this from druid and we don't build druid on linux. But yea, should add that I suppose.

PR #64, opened 2019-07-10
Make clip path closed in direct2d

raphlinus:


raphlinus:
This will be necessary for scrolling to work on Windows (it should still be possible to test on macOS without doing a release here).

raphlinus:
It's not really important to understand deeply, as that will get reworked with #65.

PR #67, opened 2019-07-20
Add a `new` method to d2d text

raphlinus:
This will be necessary to plumb a text layout context.

WIP, a similar change should be made for cairo and web impls.


raphlinus:
I think a web impl is not necessary, because WebRenderContext is used to satisfy the `Text` trait as well. It's possible we'll want to clone the underlying canvas into a separate struct for more flexibility, but I'm not sure and suggest we defer that work until we take the canvas port of druid out of the icebox.

PR #69, opened 2019-07-23
Report x advance for cairo text width

raphlinus:
For layout purposes, we want the x advance of the text, rather than
the width of the "inked" rectangle, which is what we had before.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -496,6 +496,6 @@ impl TextLayoutBuilder for CairoTextLayoutBuilder {
 
 impl TextLayout for CairoTextLayout {
     fn width(&self) -> f64 {
-        self.font.text_extents(&self.text).width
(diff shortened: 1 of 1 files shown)


raphlinus:
The bounding area of the glyphs is perhaps also useful to report, but the return value would be a `Rect` I think, not some scalars. Given that we are actually reporting x advance, it might make sense to rename our method, "width" is pretty ambiguous.

If the bounding area is useful, I'll add a method, but the web impl (as with all the other measurement stuff beyond the very basics) is going to be very rough.

cmyr:
I think maybe the main thing we're missing is just clarity around what `width` is supposed to be returning. I think folks naive guess would be that it is the width of the bounding rect, which it isn't?

raphlinus:
I'll improve either naming or docs or both when I take a serious pass through text measurement.

PR #74, opened 2019-08-06
Expose associated types

raphlinus:
This patch exposes the backend-specific associated types for brushes,
text layout, and images.

It has similar goals as #37, but actually exposes all types as concrete
types, not just as type aliases for associated types. When these types
are hidden (as was the case for `Brush` in the d2d backend) then doing
an impl of a trait on the type causes coherence problems.

Also some updates to the documentation, making sure the new types (as
well as the `Piet` type alias) show up in documentation.

### DIFF
--- a/piet-common/src/cairo_back.rs
+++ b/piet-common/src/cairo_back.rs
@@ -6,10 +6,46 @@ use cairo::{Context, Format, ImageSurface};
 
 use piet::{ErrorKind, ImageFormat};
 
+#[doc(hidden)]
 pub use piet_cairo::*;
 
 /// The `RenderContext` for the Cairo backend, which is selected.
 pub type Piet<'a> = CairoRenderContext<'a>;
+
+/// The associated brush type for this backend.
+///
+/// This type matches `RenderContext::Brush`
+pub type Brush = piet_cairo::Brush;
+
+/// The associated text factory for this backend.
+///
+/// This type matches `RenderContext::Text`
+pub type PietText<'a> = CairoText;
+
+/// The associated font type for this backend.
+///
+/// This type matches `RenderContext::Text::Font`
+pub type PietFont = CairoFont;
+
+/// The associated font builder for this backend.
+///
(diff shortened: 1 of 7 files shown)


raphlinus:
Some discussion on this. It's very much a continuation of #37, but with a slightly different approach. The motivation was wanting to write a `Brush` trait (in druid) that would get an impl from both `piet::Color` and `<Piet as RenderTarget>::Brush`. This caused the compiler to complain about coherence (E0119), which I now understand, as the compiler had no way of knowing that the `Brush` associated type of the backend was not indeed `piet::Color`, as the concrete type was hidden. So one of the things accomplished is to expose the actual type, and this lets the compiler prove that the two types are different.

Another issue for discussion is the namespace collision between traits like (`Font`) and the concrete type implementing the trait; this came up in #37 as well. Here I've resolved it by prefixing "Piet" to the new type aliases. I'm not sure this is the best solution, but I think it's workable.

I reviewed the lifetimes and am reasonably satisfied they're correct. Note that the text factory in the d2d backend has a lifetime (concretely it's a reference to a `directwrite::Factory`) but the cairo one doesn't. There's still some unpleasantry regarding lifetimes (the builder can't stash references to, say, a string), but I *think* those are blocked on [GAT](https://github.com/rust-lang/rust/issues/44265); worth re-reviewing some time but out of scope for this patch.

piet-common/src/cairo_back.rs
```diff
@@ -6,11 +6,47 @@ use cairo::{Context, Format, ImageSurface};
...
 
+/// The associated brush type for this backend.
+///
+/// This type matches `RenderContext::Brush`
+pub type Brush = piet_cairo::Brush;
+
+/// The associated text factory for this backend.
+///
+/// This type matches `RenderContext::Text`
+pub type PietText<'a> = CairoText;
```
cmyr: Interesting, I didn't know you could introduce arbitrary lifetime annotations in an alias.

cmyr:
@raphlinus can we delete this branch?

Issue #71, opened 2019-07-26
Color ergonomics improvements


raphlinus:
This is split off from discussion in #70. The current `Color::rgb24(0xrr_gg_bb)` is deemed not very rust-like. Let the bikeshedding begin!

I think the most appealing proposal on the table is to let `Color::rgb(r: u8, g: u8, b: u8)` be the primary constructor. We can rename the `u32` constructors `from_rgba32` and the like. The existing `Color::rgb` that takes three float values would have to be renamed. What to? Perhaps we keep that as-is and use `Color::rgb8` for the new constructor?

Dealing with the alpha value is tricky. A `Color::rgba` that takes an additional `a: u8` feels consistent but that does mismatch CSS, which uses a float for alpha.

There are other proposals in the thread.

Another consideration is the use of `Into<Trait>`. This prevents the constructor from being const, so I'd prefer the primary constructors not to use it, but to favor concrete types.

Lastly, we might want to clarify our thinking on whether the RGB values are defined to be interpreted as sRGB. That's more of a documentation issue than one involving constructors though.

SimonSapin:
I would say don’t worry too much about alpha having different syntax in CSS. In CSS Color Module Level 4 you can have `rgb(…)` functions with percentages (which are sort of closer to `0.0 .. 1.0` floats than `0 .. 255` integers), and `#RGBA` / `#RRGGBBAA` with four-digits or height-digits hex values.

So pick a naming convention and have four combinations:
 * `r: u8, g: u8, b: u8`
 * `r: u8, g: u8, b: u8, a: u8`
 * `r: f32, g: f32, b: f32`
 * `r: f32, g: f32, b: f32, a: f32`

… and clarify in docs what the type actually stores. If it’s 32 bits, the constructors taking f32 are doing some rounding.

cmyr:
So I think the only thing that needs clarity here are the names.

I think that doing `rgb` and `rgb8` is okay, but does feel just ever so slightly bad. The alternative is something like this, which would allow you to write `rgb(0x33, 0x44, 0xFF)` or `rgb(0.5, 1.0, 0.3)`, and which has (I think) very clear semantics in both cases, which could be clearly documented:

```rust
trait RgbaScalar {
    fn to_u8(self) -> u8;
}

// by design, all arguments must be the same concrete type; no `rgb(0xFF, 0.5, 0x23)`.
fn rgb<T: RgbaScalar>(r: T, g: T, b: T) -> Color {
[...]

PR #75, opened 2019-08-06
More ergonomic brushes

### LINKED ISSUE #72: Gradient::new concept
Now that I've created a few gradients, I believe it would be possible to make them slightly nicer to use with a constructor.

My thinking is something like this:
```rust
Gradient::new_linear(shape: impl Shape, stops: Vec<GradientStop>, rotation: f64)

Gradient::new_radial(shape: impl Shape, stops: Vec<GradientStop>)
```

For linear you could use the top and bottom of the shape's bounding_box for start and end points if rotation is 0. Not sure how to calculate it otherwise, **so maybe this won't actually work**. But it would be nice! :)

This gives you all the expressivity of CSS afaik while saving the user from needing to calculate the specific positions for the start and end points.

If it can't be done without having to define explicit start and end points (the main convenience I'm hoping to gain) then it might not be worth the hassle.

raphlinus:
Allow Color as brush argument to stroke and fill. Provide a new, more
ergonomic Gradient that uses UnitPoint, and let that be the argument to
stroke and fill as well.

Rolls in a few Color ergonomic improvements (#71).

Also adds a new `NullRenderContext` which is useful for doctests.

Fixes #72

### DIFF
--- a/piet/src/gradient.rs
+++ b/piet/src/gradient.rs
@@ -174,7 +174,7 @@ impl LinearGradient {
     /// # Examples
     ///
     /// ```
-    /// use piet::{Color, FillRule, RenderContext, LinearGradient, UnitPoint};
+    /// use piet::{Color, RenderContext, LinearGradient, UnitPoint};
     /// use piet::kurbo::{Circle, Point};
     ///
     /// # let mut render_ctx = piet::NullRenderContext::new();
@@ -184,7 +184,7 @@ impl LinearGradient {
     ///     UnitPoint::BOTTOM,
     ///     (Color::WHITE, Color::BLACK)
     /// );
-    /// render_ctx.fill(circle, &gradient, FillRule::NonZero);
+    /// render_ctx.fill(circle, &gradient);
     /// ```
     ///
     /// [`UnitPoint`]: struct.UnitPoint.html
--- a/piet/src/gradient.rs
+++ b/piet/src/gradient.rs
@@ -75,7 +75,7 @@ pub trait GradientStops {
 }
 
 /// A description of a linear gradient in the unit rect, which can be resolved
-/// to a concrete gradient.
+/// to a fixed gradient.
 ///
 /// The start and end points in the gradient are given in [`UnitPoint`] coordinates,
 /// which are then resolved to image-space coordinates for any given concrete `Rect`.
@@ -90,7 +90,7 @@ pub struct LinearGradient {
     stops: Vec<GradientStop>,
 }
 
-/// A representation of a point relative to a rectangle.
+/// A representation of a point relative to a unit rectangle.
 pub struct UnitPoint {
     u: f64,
     v: f64,
@@ -174,8 +174,8 @@ impl UnitPoint {
 
     /// Create a new UnitPoint.
     ///
-    /// The `u` and `v` coordinates describe the point, with 0.0 being
-    /// left and top, and 1.0 being right and bottom, respectively.
+    /// The `u` and `v` coordinates describe the point, with (0.0, 0.0) being
+    /// the top-left, and (1.0, 1.0) being the bottom-right.
     pub const fn new(u: f64, v: f64) -> UnitPoint {
         UnitPoint { u, v }
     }
--- a/piet/src/gradient.rs
+++ b/piet/src/gradient.rs
@@ -9,17 +9,31 @@ use crate::{IBrush, RenderContext};
 use crate::Color;
 
 /// Specification of a gradient.
+///
+/// This specification is in terms of image-space coordinates. For linear
+/// gradients, in many cases, it is better to specify coordinates relative
+/// to the `Rect` of the item being drawn; for these, use [`LinearGradient`]
+/// instead. A similar feature for radial gradients doesn't currently exist,
+/// but may be added later if needed.
+///
+/// [`LinearGradient`]: struct.LinearGradient.html
 #[derive(Clone)]
-pub enum RawGradient {
+pub enum FixedGradient {
     /// A linear gradient.
-    Linear(RawLinearGradient),
+    Linear(FixedLinearGradient),
     /// A radial gradient.
-    Radial(RawRadialGradient),
+    Radial(FixedRadialGradient),
 }
 
 /// Specification of a linear gradient.
+///
+/// This specification is in terms of image-space coordinates. In many
+/// cases, it is better to specify coordinates relative to the `Rect`
+/// of the item being drawn; for these, use [`LinearGradient`] instead.
+///
+/// [`LinearGradient`]: struct.LinearGradient.html
 #[derive(Clone)]
-pub struct RawLinearGradient {
+pub struct FixedLinearGradient {
     /// The start point (corresponding to pos 0.0).
     pub start: Point,
     /// The end point (corresponding to pos 1.0).
@@ -32,7 +46,7 @@ pub struct RawLinearGradient {
 
 /// Specification of a radial gradient.
 #[derive(Clone)]
-pub struct RawRadialGradient {
+pub struct FixedRadialGradient {
     /// The center.
     pub center: Point,
     /// The offset of the origin relative to the center.
@@ -60,7 +74,16 @@ pub trait GradientStops {
     fn to_vec(self) -> Vec<GradientStop>;
 }
 
-/// A "gradient description" that depends on the geometry being drawn.
+/// A description of a linear gradient in the unit rect, which can be resolved
+/// to a concrete gradient.
+///
+/// The start and end points in the gradient are given in [`UnitPoint`] coordinates,
+/// which are then resolved to image-space coordinates for any given concrete `Rect`.
+///
+/// When the fixed coordinates are known, use [`FixedGradient`] instead.
+///
+/// [`UnitPoint`]: struct.UnitPoint.html
+/// [`FixedGradient`]: struct.FixedGradient.html
 pub struct LinearGradient {
     start: UnitPoint,
     end: UnitPoint,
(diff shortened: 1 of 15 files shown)


piet/src/gradient.rs
```diff
@@ -1,25 +1,29 @@
...
+use kurbo::{Point, Rect, Vec2};
+
+use crate::{IBrush, RenderContext};
 
 use crate::Color;
 
 /// Specification of a gradient.
 #[derive(Clone)]
-pub enum Gradient {
+pub enum RawGradient {
```
cmyr: With the added complexity of the new 'GradientDescription' type, we want to make sure that the different between the description and the specific gradient is clear, both in naming and documentation.

I think that `Gradient` is the best name for the GradientDescription type, presuming it will be used far more frequently.

I'm less sure about the name for this type. I assume the reasoning here is that `RawGradient` is less abstract/lower level, but `Raw` also has the implication of being unprocessed, which is sort of the opposite of how we're using it.

What we're really saying here is that we have taken a gradient description and resolved it to a specific gradient, based on some concrete geometry. This gradient is (generally) no longer suitable to be reused for shapes other than the shape it was resolved for.

Names that come to mind that capture this more fully are `ResolvedGradient`, `CookedGradient`, `ConcreteGradient`, or `LiteralGradient`. Maybe `FixedGradient`.

The only major argument I could see for `RawGradient` is concision; if we expect this type to be used frequently elsewhere then it might be worth using a shorter name.

In either case, the documentation needs to clearly state the difference between the two forms.

piet/src/gradient.rs
```diff
@@ -50,3 +54,158 @@ pub struct GradientStop {
     /// The color at that stop.
     pub color: Color,
 }
+
+/// A flexible, ergonomic way to describe gradient stops.
+pub trait GradientStops {
```
cmyr: You mentioned considering `Into<Vec<GradientStop>>`, was there a particular reason to prefer a new trait?

piet/src/gradient.rs
```diff
@@ -50,3 +54,158 @@ pub struct GradientStop {
...
+    fn to_vec(self) -> Vec<GradientStop>;
+}
+
+/// A "gradient description" that depends on the geometry being drawn.
```
cmyr: I would say something like, "A description of a linear gradient in the unit rect, which can be resolved to a concrete gradient for any given concrete `Rect`."

piet/src/gradient.rs
```diff
@@ -50,3 +54,158 @@ pub struct GradientStop {
...
+    /// let gradient = LinearGradient::new(
+    ///     UnitPoint::TOP,
+    ///     UnitPoint::BOTTOM,
+    ///     (Color::WHITE, Color::BLACK)
+    /// );
+    /// render_ctx.fill(circle, &gradient, FillRule::NonZero);
+    /// ```
+    ///
+    /// [`UnitPoint`]: struct.UnitPoint.html
+    pub fn new(start: UnitPoint, end: UnitPoint, stops: impl GradientStops) -> LinearGradient {
```
cmyr: that this function is essentially variadic makes me wonder if life might not be easier if we had a `linear_grad!` macro; took a quick look at this though and it isn't totally trivial, especially if we want some of the nice stuff like handling both `Color`s and `GradientStop`s.

piet/src/render_context.rs
```diff
@@ -48,11 +50,14 @@ impl ImageFormat {
 /// can implement this trait.
 ///
 /// Code that draws graphics will in general take `&mut impl RenderContext`.
-pub trait RenderContext {
+pub trait RenderContext
+where
+    Self::Brush: IBrush<Self>,
```
cmyr: 🤨

piet/src/render_context.rs
```diff
@@ -170,3 +180,17 @@ pub trait RenderContext {
     /// aspect ratios don't match.
     fn draw_image(&mut self, image: &Self::Image, rect: impl Into<Rect>, interp: InterpolationMode);
 }
+
+/// A brush that can be dependent on the geometry being drawn.
+pub trait IBrush<P: RenderContext>
+where
+    P: ?Sized,
+{
+    fn make_brush<'a>(&'a self, piet: &mut P, bbox: impl FnOnce() -> Rect) -> Cow<'a, P::Brush>;
```
cmyr: Would it be possible for this function to take ownership of self, and return an owned brush? I'm not sure if this makes sense, but it might make for a cleaner signature, and possibly get rid of the `Cow` business.

piet/src/gradient.rs
raphlinus: Of these, I like FixedGradient best, as it actually says what the difference is. I agree on the documentation and will address that.

piet/src/gradient.rs
raphlinus: Yes, you can't do `Into` on existing types, so the conversion from a slice into a Vec won't fly. With a new trait, no problem.

piet/src/render_context.rs
cmyr: Thinking a bit more I guess not; this would prevent reusing cooked brushes.

piet/src/gradient.rs
raphlinus: I think what's there is a pretty good compromise. I was trying to make some form of `&[color1, color2]` work, but that syntax actually gives you a `&[Color; 2]` type, not a slice. It's likely Rust will get better at this sort of thing, with integer generics or whatnot.

I'm against the use of macros for minor syntax sugaring; it makes the code a *lot* harder to follow. So my strong inclination is to play with types first, and only use macros if there's a lot of repetitive boilerplate, or if there's an invariant that's easier to enforce.

piet/src/render_context.rs
raphlinus: Can you expand? It turns out we need to write a fair amount of Chalk to get all these types to work together.

piet/src/render_context.rs
raphlinus: A goal was for it to take an existing `&Brush`. Taking ownership of self seems problematic to me, as a lot of the time you're going to get these resources from somewhere (env, internal widget state).

This trait has a complex type, for sure, but it's not something users are going to have to deal with much; I think of it as similar to the hashmap types.

piet/src/render_context.rs
cmyr: This is just another instance of "I did not know this would compile". Cool that we can express these sorts of circular relationships though!

piet/src/gradient.rs
```diff
@@ -50,3 +68,168 @@ pub struct GradientStop {
...
     pub color: Color,
 }
+
+/// A flexible, ergonomic way to describe gradient stops.
+pub trait GradientStops {
+    fn to_vec(self) -> Vec<GradientStop>;
+}
+
+/// A description of a linear gradient in the unit rect, which can be resolved
+/// to a concrete gradient.
```
cmyr:
```suggestion
/// to a fixed gradient.
```

piet/src/gradient.rs
```diff
@@ -50,3 +68,168 @@ pub struct GradientStop {
...
+///
+/// [`UnitPoint`]: struct.UnitPoint.html
+/// [`FixedGradient`]: struct.FixedGradient.html
+pub struct LinearGradient {
+    start: UnitPoint,
+    end: UnitPoint,
+    stops: Vec<GradientStop>,
+}
+
+/// A representation of a point relative to a rectangle.
```
cmyr: Maybe,

```suggestion
/// A representation of a point relative to a unit rectangle.
```

piet/src/gradient.rs
```diff
@@ -50,3 +68,168 @@ pub struct GradientStop {
...
+pub struct UnitPoint {
+    u: f64,
```
cmyr: I'm curious why this is `u, v` and not `x, y`, assume there's some mathy precedent?

piet/src/gradient.rs
```diff
@@ -50,3 +68,168 @@ pub struct GradientStop {
...
+    pub const LEFT: UnitPoint = UnitPoint::new(0.0, 0.5);
+    pub const CENTER: UnitPoint = UnitPoint::new(0.5, 0.5);
+    pub const RIGHT: UnitPoint = UnitPoint::new(1.0, 0.5);
+    pub const BOTTOM_LEFT: UnitPoint = UnitPoint::new(0.0, 1.0);
+    pub const BOTTOM: UnitPoint = UnitPoint::new(0.5, 1.0);
+    pub const BOTTOM_RIGHT: UnitPoint = UnitPoint::new(1.0, 1.0);
+
+    /// Create a new UnitPoint.
+    ///
+    /// The `u` and `v` coordinates describe the point, with 0.0 being
```
cmyr:
```suggestion
    /// The `u` and `v` coordinates describe the point, with (0.0, 0.0) being
```

piet/src/gradient.rs
```diff
@@ -50,3 +68,168 @@ pub struct GradientStop {
...
+    /// left and top, and 1.0 being right and bottom, respectively.
```
cmyr:
```suggestion
    /// the top-left, and (1.0, 1.0) the bottom-right.
```

piet/src/shapes.rs
```diff
@@ -1,14 +1,5 @@
 //! Options for drawing paths.
 
-/// A fill rule for resolving winding numbers.
```
cmyr: 🏆 my hero! 🥇

piet/src/gradient.rs
raphlinus: Both would work, but u, v is common for texture coords normalized to the unit square.

PR #77, opened 2019-08-08
Add IBrush impl for FixedGradient

raphlinus:
There's no reason these can't be used as well, as opposed to explicitly
creating a gradient. There's a bit of extra cloning which we could
probably avoid, but unlikely to be a serious performance problem.

### DIFF
--- a/piet/src/gradient.rs
+++ b/piet/src/gradient.rs
@@ -233,3 +233,14 @@ impl<P: RenderContext> IBrush<P> for LinearGradient {
         Cow::Owned(piet.gradient(gradient).expect("error creating gradient"))
     }
 }
+
+impl<P: RenderContext> IBrush<P> for FixedGradient {
+    fn make_brush<'a>(&'a self, piet: &mut P, _bbox: impl FnOnce() -> Rect) -> Cow<'a, P::Brush> {
+        // Perhaps the make_brush method should be fallible instead of panicking.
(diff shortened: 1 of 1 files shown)


piet/src/gradient.rs
```diff
@@ -233,3 +233,14 @@ impl<P: RenderContext> IBrush<P> for LinearGradient {
         Cow::Owned(piet.gradient(gradient).expect("error creating gradient"))
     }
 }
+
+impl<P: RenderContext> IBrush<P> for FixedGradient {
+    fn make_brush<'a>(&'a self, piet: &mut P, _bbox: impl FnOnce() -> Rect) -> Cow<'a, P::Brush> {
+        // Perhaps the make_brush method should be fallible instead of panicking.
```
cmyr: Another option would be to have a 'null' brush that backends can return in the case of failure? This could either be a solid color, if that is not fallible, or some sentinel value otherwise.

I'm probably over thinking this, but we don't really have a way to handle failure, so if we can just plough through safely that seems reasonable.

raphlinus: I'm personally a bit of a fan of using a ridiculously garish color (like magenta) because it's very visually prominent, but there are lots of ways to think about this.

cmyr: yes, I'm totally on board with that.

PR #79, opened 2019-08-08
Add RadialGradient

cmyr:
This fills the same role as LinearGradient; it is a gradient
description that can b reused for different shapes of different
sizes.

### DIFF
--- a/piet/src/gradient.rs
+++ b/piet/src/gradient.rs
@@ -17,7 +17,7 @@ use crate::Color;
 /// but may be added later if needed.
 ///
 /// [`LinearGradient`]: struct.LinearGradient.html
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub enum FixedGradient {
     /// A linear gradient.
     Linear(FixedLinearGradient),
@@ -32,7 +32,7 @@ pub enum FixedGradient {
 /// of the item being drawn; for these, use [`LinearGradient`] instead.
 ///
 /// [`LinearGradient`]: struct.LinearGradient.html
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct FixedLinearGradient {
     /// The start point (corresponding to pos 0.0).
     pub start: Point,
@@ -45,7 +45,7 @@ pub struct FixedLinearGradient {
 }
 
 /// Specification of a radial gradient.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct FixedRadialGradient {
     /// The center.
     pub center: Point,
@@ -61,7 +61,7 @@ pub struct FixedRadialGradient {
 }
 
 /// Specification of a gradient stop.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct GradientStop {
     /// The coordinate of the stop.
     pub pos: f32,
@@ -84,13 +84,56 @@ pub trait GradientStops {
 ///
 /// [`UnitPoint`]: struct.UnitPoint.html
 /// [`FixedGradient`]: struct.FixedGradient.html
+#[derive(Debug, Clone)]
 pub struct LinearGradient {
     start: UnitPoint,
     end: UnitPoint,
     stops: Vec<GradientStop>,
 }
 
+/// A description of a radial gradient in the unit rect, which can be resolved
+/// to a fixed gradient.
+///
+/// The center is given in `UnitPoint` coordinates.
+///
+/// The `center` and `radius` describe a sphere. `origin` describes the angle
+/// of the gradient on that sphere, useful for simulating lighting effects.
+/// see [configuring a radial gradient][config] for a fuller explanation.
+///
+/// By default, `origin` is equal to `center`. This can be changed during construction
+/// with the [`with_origin`] builder method.
+///
+/// The [`ScaleMode`] describes how the gradient is mapped to a non-square
+/// rectangle; by default this will expand on the longest axis, but this can
+/// be changed with the [`with_scale_mode`] builder method.
+///
+/// [config]: https://docs.microsoft.com/en-us/windows/win32/direct2d/direct2d-brushes-overview#configuring-a-radial-gradient
+/// [`ScaleMode`]: enum.ScaleMode.html
+/// [`with_origin`]: struct.RadialGradient.html#method.with_origin
+/// [`with_scale_mode`]: struct.RadialGradient.html#method.with_scale_mode
+#[derive(Debug, Clone)]
+pub struct RadialGradient {
+    center: UnitPoint,
+    origin: UnitPoint,
+    radius: f64,
+    stops: Vec<GradientStop>,
+    scale_mode: ScaleMode,
+}
+
+/// Mappings from the unit square into a non-square rectangle.
+#[derive(Debug, Clone)]
+pub enum ScaleMode {
+    /// The unit 1.0 is mapped to the smaller of width & height. All
+    /// values will be mapped without clipping, but the mapped item will
+    /// not cover the entire reect.
+    Fit,
+    /// The unit 1.0 is mapped to the larger of width & height; some
+    /// values on the other axis will be clipped.
+    Fill,
+}
+
 /// A representation of a point relative to a unit rectangle.
+#[derive(Debug, Clone, Copy)]
 pub struct UnitPoint {
     u: f64,
     v: f64,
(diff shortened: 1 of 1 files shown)


piet/src/gradient.rs
```diff
@@ -61,7 +61,7 @@ pub struct FixedRadialGradient {
 }
 
 /// Specification of a gradient stop.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
raphlinus: Minor nit, but order is not consistent. I'm not sure what "best practice" is here and it doesn't really matter, but for types nearby in the source it would be nice to be consistent.

piet/src/gradient.rs
```diff
@@ -84,13 +84,56 @@ pub trait GradientStops {
...
+    origin: UnitPoint,
+    radius: f64,
+    stops: Vec<GradientStop>,
+    scale_mode: ScaleMode,
+}
+
+/// Mappings from the unit square into a non-square rectangle.
+#[derive(Debug, Clone)]
+pub enum ScaleMode {
+    /// The unit 1.0 is mapped to the smaller of width & height. All
```
raphlinus: I think talking about "clipping" here might be confusing, as there's no additional clipping logic based on the rect. You might well see behavior that could be described as clipping when the radius is 0.5.

piet/src/gradient.rs
```diff
@@ -84,13 +84,56 @@ pub trait GradientStops {
...
+    /// values will be mapped without clipping, but the mapped item will
+    /// not cover the entire reect.
```
raphlinus: Probably "rectangle".

piet/src/gradient.rs
```diff
@@ -221,6 +264,43 @@ impl LinearGradient {
     }
 }
 
+impl RadialGradient {
+    /// Creates a simple `RadialGradient`. This gradient has the same `origin`
```
raphlinus: Would it make sense to have origin and center default to center? I think that's the most common use case, and makes sense as a default, but this is also ok.

piet/src/gradient.rs
```diff
@@ -233,3 +313,48 @@ impl<P: RenderContext> IBrush<P> for LinearGradient {
...
+
+impl<P: RenderContext> IBrush<P> for RadialGradient {
+    fn make_brush<'a>(&'a self, piet: &mut P, bbox: impl FnOnce() -> Rect) -> Cow<'a, P::Brush> {
+        let rect = bbox();
+        let scale_len = match self.scale_mode {
+            ScaleMode::Fill => rect.width().max(rect.height()),
+            ScaleMode::Fit => rect.width().min(rect.height()),
+        };
+
+        let rect =  equalize_sides_preserving_center(rect, scale_len);
```
raphlinus: The extra space is why it's failing CI :/

piet/src/gradient.rs
```diff
@@ -233,3 +313,48 @@ impl<P: RenderContext> IBrush<P> for LinearGradient {
...
+            radius,
+            stops: self.stops.clone(),
+        });
+        // Perhaps the make_brush method should be fallible instead of panicking.
+        Cow::Owned(piet.gradient(gradient).expect("error creating gradient"))
+    }
+}
+
+fn equalize_sides_preserving_center(rect: Rect, new_len: f64) -> Rect {
+    let (x, width) = if new_len != rect.width() {
```
raphlinus: I think the "if" here doesn't really help - the conditional branch is likely to be slower than just plowing through the math.

piet/src/gradient.rs
```diff
@@ -233,3 +313,48 @@ impl<P: RenderContext> IBrush<P> for LinearGradient {
...
+            origin_offset,
+            radius,
+            stops: self.stops.clone(),
+        });
+        // Perhaps the make_brush method should be fallible instead of panicking.
+        Cow::Owned(piet.gradient(gradient).expect("error creating gradient"))
+    }
+}
+
+fn equalize_sides_preserving_center(rect: Rect, new_len: f64) -> Rect {
```
raphlinus: Here's a stab at writing it in a higher level style:

```rust
    let size = Size::new(new_len, new_len);
    let origin = rect.center() - size.as_vec2() / 2;
    Rect::from_origin_size(origin, size)
```

piet/src/gradient.rs
cmyr: Good idea, I think that's worth the simplification.

piet/src/gradient.rs
cmyr: we need a :nerdsnipe: emoji. 🤓💧🔫 ?

cmyr: (what I mean to say is, this is a huge improvement!)

piet/src/gradient.rs
cmyr: interesting, didn't even notice that that would've worked.

piet/src/gradient.rs
```diff
@@ -1,22 +1,44 @@
 //! Gradient specifications.
+//!
+//! We provide both linear and radial gradients; and for each flavor
+//! we provide two representations, a 'generic' representation that uses
+//! points in the [unit square], and a 'fixed' representation that uses
+//! image-space coordinates.
+//!
+//! The generic representations ([`LinearGradient`] and [`RadialGradient`])
+//! are strictly more flexible than the fixed representations
```
raphlinus: I'm not 100% sure I agree with this "strictly more flexible." I consider them sibling use cases. You'll use the unit square for UI, but the fixed ones for dealing with content in existing formats such as SVG. I think this doc is clear enough without the editorial. Probably what I'd say is "a convenience for use cases such as UI." The part after the semicolon is fine.

piet/src/gradient.rs
```diff
@@ -123,12 +150,11 @@ pub struct RadialGradient {
...
 pub enum ScaleMode {
-    /// The unit 1.0 is mapped to the smaller of width & height. All
-    /// values will be mapped without clipping, but the mapped item will
-    /// not cover the entire reect.
+    /// The unit 1.0 is mapped to the smaller of width & height, but the mapped
+    /// item may not cover the entire rectangle.
     Fit,
     /// The unit 1.0 is mapped to the larger of width & height; some
-    /// values on the other axis will be clipped.
+    /// values on the other axis may extend beyond the bounding rectangle.
```
raphlinus: I think I'd say "unit" rather than "bounding" here. It doesn't actively bound (as in clipping or clamping the gradient values).

piet/src/gradient.rs
```diff
@@ -262,26 +288,52 @@ impl LinearGradient {
...
+            origin: UnitPoint::CENTER,
             radius,
             stops: stops.to_vec(),
             scale_mode: ScaleMode::Fill,
         }
     }
 
+    /// A builder-style method for changing the center of the gradient.
+    ///
+    /// See the main [`RadialGradient`] docs for an explanation of center vs. origin.
```
raphlinus: Might be worth mentioning that it leaves the origin unchanged. (In the common case of no offset, now we'll have to call both methods)

piet/src/gradient.rs
```diff
@@ -262,26 +288,52 @@ impl LinearGradient {
...
+            stops: self.stops.clone(),
+        }
+    }
 }
 
 impl RadialGradient {
-    /// Creates a simple `RadialGradient`. This gradient has the same `origin`
-    /// and `center`, and uses the `Fill` [`ScaleMode`]. These attributes can be
-    /// modified with the [`with_origin`] and [`with_scale_mode`] builder methods.
+    /// Creates a simple `RadialGradient`. This gradient has `origin` and `center`
```
raphlinus: I think a short explanation of origin and center would be handy, as I don't think they're universally understood terms. Here's what I suggest:

The `center` parameter specifies the center of the circle, so that points of distance `radius` from the center map to 1.0 in the gradient stops. The `origin` parameter specifies the point that maps to 0.0 in the gradient stops. Using the same origin and center gives a radially symmetric gradient; using different points is useful for simulating lighting effects.

cmyr: Thanks, I struggled to communicate that distinction clearly.

piet/src/gradient.rs
cmyr: that seems confusing, since the values as defined are within the unit; it's just that when we inflate them they may end up outside the target.

piet/src/gradient.rs
cmyr: although I'll put this in the type-level docstring, since neither of these fields are exposed in `new`.

PR #84, opened 2019-08-09
Add PaintBrush enum

cmyr:
This is a concrete type that can be any color or gradient. (And
perhaps in the future an image?)

### DIFF
--- a/piet/src/render_context.rs
+++ b/piet/src/render_context.rs
@@ -4,7 +4,10 @@ use std::borrow::Cow;
 
 use kurbo::{Affine, Point, Rect, Shape};
 
-use crate::{Color, Error, FixedGradient, StrokeStyle, Text, TextLayout};
+use crate::{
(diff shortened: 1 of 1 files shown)


piet/src/render_context.rs
```diff
@@ -201,3 +204,86 @@ impl<P: RenderContext> IntoBrush<P> for Color {
...
+
+impl From<FixedGradient> for PaintBrush {
+    fn from(src: FixedGradient) -> PaintBrush {
+        PaintBrush::Fixed(src)
+    }
+}
+
+impl From<FixedLinearGradient> for PaintBrush {
+    fn from(src: FixedLinearGradient) -> PaintBrush {
+        PaintBrush::Fixed(src.into())
```
raphlinus: A curiosity: this can probably be written `src.into().into()`. But writing it out more explicitly is probably better.

cmyr: my hunch is no, double inference falls over even in places where we'd expect it to be totally clear.

(tested out of curiosity, and yea this doesn't work 😢)

PR #85, opened 2019-08-10
Add text example to piet-test

hwchen:
Basic example that's text-only. In preparation for testing api changes
and new features.

### DIFF
--- a/piet-test/src/picture_5.rs
+++ b/piet-test/src/picture_5.rs
@@ -1,0 +1,25 @@
+//! Basic example of just text
+
+use piet::kurbo::Line;
+use piet::{Color, Error, FontBuilder, RenderContext, Text, TextLayout, TextLayoutBuilder};
+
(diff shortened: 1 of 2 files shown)


piet-test/src/picture_5.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn draw(rc: &mut impl RenderContext) -> Result<(), Error> {
+    // Black background
+    rc.clear(Color::BLACK);
+
+    // do something texty
+    let font = rc.text().new_font_by_name("Segoe UI", 12.0)?.build()?;
+
+    let layout = rc.text().new_text_layout(&font, "piet text!")?.build()?;
+
+    let width: f64 = layout.width().into();
```
raphlinus: The `into` is not needed here, nor is the type annotation, as we got rid of the associated type (which was designed to be either f32 or f64).

hwchen: I cleaned this up and amended the commit.

PR #88, opened 2019-08-14
Remove `Result` when creating Font and TextLayout builders

### LINKED ISSUE #76: Remove double result in text methods
Doing anything with text requires a lot of result unwrapping. Bascially, creating the builder can fail, as can the result of building. This is true for both `Font` and `TextLayout`. As part of our ergonomics push, we should make the first infallible. If there's a platform error in constructing the lower level builder object, we could either panic (if it's basically a "totally out of resources" or "configuration is irrecoverably borked" situation), or retain an error indication in the builder which would get reported on the call to the final builder.

hwchen:
Creating Font and TextLayout builders should not fail, so to avoid
unnecessary unwrapping for users, this commit removes the Result from
the `new_font_by_name` and `new_text_layout` methods in the `Text`
trait.

This is a user API breaking change.

Also updated implementers of `Text` (all existing backends):
- null_renderer
- direct2d
- cairo
- web

Also updated user of `Text`, `piet-crate`.

As an ergonomic change for devs, this commit also:
- adds `#![cfg(windows)]` to the `direct-2d` backend so that it does
not attempt to compile on linux.
- adds `.cargo` to .gitignore; it's used for cross compilation config

Closes #76

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ target/
 **/*.rs.bk
 Cargo.lock
 .DS_Store
+.cargo
 
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -362,25 +362,21 @@ impl Text for CairoText {
     type TextLayout = CairoTextLayout;
     type TextLayoutBuilder = CairoTextLayoutBuilder;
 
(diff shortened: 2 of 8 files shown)


hwchen:
AzurePipelines run

Issue #65, opened 2019-07-10
Migrate to newer direct2d


raphlinus:
Currently we're using [direct2d 0.2.0](https://crates.io/crates/direct2d). There are a number of problems with this, including excess debug logging on paths, not being able to set the closed/open status of a path in a single pass when iterating from path elements. I'd like to get us onto the newer 0.3 versions of the direct2d and related crates, but they're still in alpha. I'm wondering what it will take to get those released. It might be worth developing the patch based on alpha so we can shake out any problems.

I've been figuring that we would update the direct2d deps before taking on #60, but if it's going to take a while maybe we should special case `Rect` in particular, as that's the biggest pain point.

@Connicpu who owns the direct2d crates. I might ask for admin privs on those crates so we can move those forward. Also, this is a great "help wanted" issue for someone who wants to dive in.

raphlinus:
A quick note. The new 0.3 alpha crates are missing some methods. In one case, this should force a change in our usage: we should migrate from GenericRenderContext to DeviceContext. The latter is strictly more powerful, and the compatibility requirement is ok (it works on Windows 7 with platform update).

Otherwise, if it has everything we need and works, we can do a release of the directx crates.

It's possible there's something we need that's missing, in which case we should add that.

Kethku:
I'm taking a look at making this change currently as I am running into weird performance issues with the current druid-shell and want to get a better understanding of how things work

raphlinus:
I think @hwchen currently has the token on this - we're going with the raw_d2d branch, as it gives us the most flexibility going forward.

Kethku:
Gotcha. I will hold off then. Is that branch in a semi usable state for me to play around? Or is it in progress?

Kethku:
Nevermind. I see the branch is pretty old. I will hold off until something further happens in this area.

hwchen:
I'm doing some work on it on my own branch, but yes it's pretty far behind master at this point.

msiglreith:
With https://github.com/linebender/piet/pull/111 is this done?

raphlinus:
Yes, closing now, thanks for the reminder.

PR #91, opened 2019-09-09
Fix closed paths on stroke (d2d)

raphlinus:
Do a slightly more inefficient conversion to bezier paths so we can
tell in advance whether a subpath needs to be closed. This is a bit of
a workaround, as there are plans to improve the direct2d bindings so
we can set the is_closed at the end of the subpath, enabling a single
pass of iteration (tracked in #65).

### DIFF
--- a/piet-direct2d/src/lib.rs
+++ b/piet-direct2d/src/lib.rs
@@ -132,6 +132,12 @@ impl<'a> PathBuilder<'a> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


raphlinus:
The CI failure is due to rustfmt on nightly, which is [failing again](https://rust-lang.github.io/rustup-components-history/); we probably want to do something similar to https://github.com/xi-editor/druid/pull/93 here.

PR #93, opened 2019-10-13
Text Layout: Hit Testing

hwchen:
Cross-platform hit testing, both for point and for text-position

Very basic implementation, which does not take into account line breaking, bidi, etc.

- directwrite: a thin wrapper around directwrite api.
- web/cairo: uses `unicode-segmentation` to find grapheme boundaries.


hwchen:
@raphlinus no new work since last check-in, just getting myself organized.

hwchen:
@raphlinus the following is now reviewable for cairo:

- `hit_test_text_position`
- `get_grapheme_boundaries` in graphemes module (may be moved around soon)
- some tests for both

Generally speaking, this is the beginning of work to have hit testing work correctly with graphemes. So I've tried to put in the beginning of a test suite to ensure correctness.

Would also be helpful to know if my testing strategy is reasonable.

Question: should `text_position` parameter be usize or u32 (in `hit_test_text_position`)?

Please don't mind the `point_x` and `point_y` stuff, I'm about to change to `kurbo::Point` :sweat_smile: Also just realized I have an `expect` I need to remove.

hwchen:
`get_grapheme_boundaries` has logic to use only trailing edges for the boundaries. But it calls `hit_test_text_position` which also only uses trailing edges. So perhaps redundant, unless there's a reason it shouldn't rely on the underlying implementation. Haven't thought this through completely yet.

hwchen:
@raphlinus `hit_test_point` now works on cairo, with a basic test.

I have a question about expected behavior. Is `hit_test_point` a place where we should treat all edges as trailing edges (for now, and with exception of leading edge of text layout)?

This way, when rounding to the nearest edge, it would treat that edge as a trailing edge and return the associated text position.
- for hit on "leading half" of a grapheme, would return text position of previous grapheme.
- for hit on "trailing half" of a grapheme, would return text position of current grapheme.

Current behavior is to not round, any hit on a grapheme returns that grapheme. And I didn't really deal with leading/trailing edge yet.

raphlinus:
A few responses.

Offsets within strings should be `usize`. The fact that DWrite uses `u32` is arguably an artifact from the 32 bit days. (That said, an actual string larger than 8GB or whatever is not going to happen in practice - I suggest `try_into` with panic so it doesn't create a soundness hole from a malicious attack).

Regarding leading/trailing, the distinction is mostly useful for BiDi. I don't think it actually implements "round up / round down" behavior in single direction text, but I could be wrong about this - it might be worth experimenting with. If so, it might be a handier implementation strategy.

But my hunch is that, even if this is so, with BiDi text it won't be easy to get "round up / round down" behavior, because there are two relevant cursors at direction change points. That said, now that I think about it, that might be reasonable UI - click slightly to the left of the split point to get one result, slightly to the right to get another.

It might be worth trying to figure out if there's another way to coax "round to nearest" behavior from DirectWrite.

I talked with Andrew Glass of MS and some other people who know this stuff, at the latest Unicode Conference, and they pointed out that MS Word doesn't use the Unicode concept of grapheme cluster boundary for things like cursor positioning in complex script cases, but rather queries the font. I'm actually not 100% sure how this works, because the "fi" ligature clearly has a grapheme cluster boundary at its interior.

If I get a little time I might experiment with this some more. But for the time being I think it's ok to do something that doesn't take into account BiDi or the more subtle complex script cases.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        //assert_eq!(input.len(), 17);
+
+        let input = "\u{0023}\u{FE0F}\u{20E3}"; // #️⃣
+
+        let mut text_layout = D2DText::new(&dwrite);
+        let font = text_layout.new_font_by_name("Segoe UI", 12.0).build().unwrap();
+        let layout = text_layout.new_text_layout(&font, input).build().unwrap();
+
+        assert_eq!(layout.hit_test_text_position(0, true).map(|p| p.point.x), Some(layout.width()));
+        assert_eq!(input.len(), 7);
```
raphlinus: For style, I think these would make sense right after the def of `input` and before the querying - they seem to be sanity-checks rather than something that actually tests the lib.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        // 10 utf-8 code units (2/1/3/3/1)
+        // 3 graphemes
+        let input = "é\u{0023}\u{FE0F}\u{20E3}1"; // #️⃣
+
+        let mut text_layout = D2DText::new(&dwrite);
+        let font = text_layout.new_font_by_name("Segoe UI", 12.0).build().unwrap();
+        let layout = text_layout.new_text_layout(&font, input).build().unwrap();
+
+        let test_layout_0 = text_layout.new_text_layout(&font, "é").build().unwrap();
+        let test_layout_1 = text_layout.new_text_layout(&font, "é\u{0023}\u{FE0F}\u{20E3}").build().unwrap();
```
raphlinus: Would it be better to slice the source string, rather than use a new literal?

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        //assert_eq!(input.graphemes(true).count(), 3);
+        assert_eq!(input.len(), 10);
+
+        // Note: text position is in terms of utf8 code units
+        assert_eq!(layout.hit_test_text_position(0, true).map(|p| p.point.x), Some(test_layout_0.width()));
+        assert_eq!(layout.hit_test_text_position(2, true).map(|p| p.point.x), Some(test_layout_1.width()));
+        assert_eq!(layout.hit_test_text_position(9, true).map(|p| p.point.x), Some(layout.width()));
+
+        assert_eq!(layout.hit_test_text_position(2, false).map(|p| p.point.x), Some(test_layout_0.width()));
+        assert_eq!(layout.hit_test_text_position(9, false).map(|p| p.point.x), Some(test_layout_1.width()));
```
raphlinus: Similar considerations as above (I didn't find any new problems).

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        let mut text_layout = D2DText::new(&dwrite);
+
+        let font = text_layout.new_font_by_name("Segoe UI", 12.0).build().unwrap();
+        let layout = text_layout.new_text_layout(&font, "piet text!").build().unwrap();
+        println!("text pos 4 leading: {:?}", layout.hit_test_text_position(4, false)); // 20.302734375
+        println!("text pos 4 trailing: {:?}", layout.hit_test_text_position(4, true)); // 23.58984375
+
+        // test hit test point
+        // all inside
+        let pt = layout.hit_test_point(Point::new(21.0, 0.0));
```
raphlinus: Urgh. Using metrics from the font is fragile. Otoh, there's a built-in "slop" here because of the snap-to-closest semantics. So I'm ok with this staying as-is and fixing if/when tests fail. I'm also curious to run it on Windows 7.

Based on my reading of the tests, it implements "round to nearest" behavior, then `is_trailing_hit` indicates whether the rounding was up or down.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        assert_eq!(pt.is_trailing_hit, true);
+        let pt = layout.hit_test_point(Point::new(18.0, 0.0));
+        assert_eq!(pt.metrics.text_position, 9);
+        assert_eq!(pt.is_trailing_hit, false);
+        let pt = layout.hit_test_point(Point::new(19.0, 0.0));
+        assert_eq!(pt.metrics.text_position, 9);
+        assert_eq!(pt.is_trailing_hit, false);
+    }
+
+    #[test]
```
raphlinus: I think this should test an astral plane codepoint.

piet/src/text.rs
```diff
@@ -31,4 +32,38 @@ pub trait TextLayoutBuilder {
...
+
+    /// Given a Point, determine the corresponding text position
+    fn hit_test_point(&self, point: Point) -> HitTestPoint;
+
+    /// Given a text position, determine the corresponding pixel location
+    /// (currently consider the text layout just one line)
+    fn hit_test_text_position(&self, text_position: usize, trailing: bool) -> Option<HitTestTextPosition>;
 }
+
+#[derive(Debug, Default)]
```
raphlinus: Would prefer at least a 1-line docstring for all public types.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+                is_text: htp.metrics.is_text(),
+            },
+            is_inside: htp.is_inside,
+            is_trailing_hit: htp.is_trailing_hit,
+        }
+    }
+
+    fn hit_test_text_position(&self, text_position: usize, trailing: bool) -> Option<HitTestTextPosition> {
+        // now convert the ut8 index to utf16
+        // TODO this can panic; should try next text_position
```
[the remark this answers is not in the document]
hwchen: ok, I'll leave it as is then.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        let text_position = htp.metrics.text_position() as usize;
+
+        // Convert text position from utf-16 code units to
+        // utf-8 code units.
+        // Strategy: count up in utf16 and utf8 simultaneously, stop when
+        // utf-16 text position reached.
+        //
+        // TODO ask about text_position, it looks like windows returns last index;
+        let idx_8 = count_until_utf16(&self.text, text_position)
+            .unwrap_or(self.text.len() - 1);
```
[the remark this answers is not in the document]
hwchen: Discussed on zulip; this will be modified (and `text.len()` is a valid position) to fit directwrite semantics to piet semantics.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        // TODO this can panic; should try next text_position
+        let idx_16 = count_utf16(&self.text[0..text_position]);
+
+        // panic or Result are also fine options for dealing with overflow. Using Option here
+        // because it's already present and convenient.
+        // TODO this should probably go before convertin to utf16, since that's relatively slow
+        let idx_16 = idx_16.try_into().ok()?;
+
+
+        // Check for idx out of bounds.
```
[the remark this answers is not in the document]
hwchen: Sure. Directwrite will return the line width if you try a textPosition that's beyond the end, so I'm doing an explicit check instead to get back None.

However, I now realize that this branch is taken only when `text_position = text.len()`, since that's the one time text position is both out of bounds and can be used as a boundary for the slice for counting utf16 code units. Any higher, and that slice will panic first.

Let me know if you have any thoughts on this, otherwise I'll figure out something consistent.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        self.layout.get_line_metrics(&mut line_metrics);
+        let line_len = line_metrics.iter()
+            .map(|l| l.length())
+            //.nth(0).unwrap() as usize;
+            .sum::<u32>()
+            // TODO directwrite wrapper adds an extra \u{0}, so subtract 1 from len
+            // https://docs.rs/directwrite/0.1.4/src/directwrite/text_layout/builder.rs.html#34
+            - 1;
+
+        if idx_16 >= line_len {
```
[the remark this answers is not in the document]
hwchen: Oh, I see, this is a followup from the above comment. If we should always report a position, even if text position is out of bounds, I'll just get rid of the explicit check.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+    println!("");
+    for &b in s.as_bytes() {
+        if (b as i8) >= -0x40 {
+            utf16_count += 1;
+        }
+        if b >= 0xf0 {
+            utf16_count += 1;
+        }
+
+        utf8_count += 1;
```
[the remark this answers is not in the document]
hwchen: Thanks! changed.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+
+        let layout = text_layout.new_text_layout(&font, "p").build().unwrap();
+        let p_width = layout.width();
+
+        let layout = text_layout.new_text_layout(&font, "").build().unwrap();
+        let null_width = layout.width();
+
+        let full_layout = text_layout.new_text_layout(&font, "piet text!").build().unwrap();
+
+        assert_eq!(full_layout.hit_test_text_position(3, true).map(|p| p.point.x as f64), Some(piet_width));
```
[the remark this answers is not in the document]
hwchen: Just to confirm my understanding; because of kerning, substrings may not have the same width when inside a string and when separate from a string.

To implement your suggestion, I'd probably make a macro `assert_kinda_eq` with adjustable tolerance. Let me know if this sounds ok.

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        assert_eq!(full_layout.hit_test_text_position(2, true).map(|p| p.point.x as f64), Some(pie_width));
+        assert_eq!(full_layout.hit_test_text_position(1, true).map(|p| p.point.x as f64), Some(pi_width));
+        assert_eq!(full_layout.hit_test_text_position(0, true).map(|p| p.point.x as f64), Some(p_width));
+
```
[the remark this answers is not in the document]
hwchen: I think this is addressed above; my current plan is to return an x offset even if text position is out of bounds. Plus, `10` would be in bounds in this case, as the last text position.

piet-direct2d/src/lib.rs
hwchen: agreed

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,358 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+        assert_eq!(full_layout.hit_test_text_position(0, false).map(|p| p.point.x as f64), Some(null_width));
+        assert_eq!(full_layout.hit_test_text_position(9, true).map(|p| p.point.x as f64), Some(full_layout.width()));
+        assert_eq!(full_layout.hit_test_text_position(10, true).map(|p| p.point.x as f64), None);
```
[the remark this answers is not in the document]
hwchen: The behavior you mention for `getPrimaryHorizontal` is similar to what I was thinking for cairo. For directwrite, I've taken their api more at face value so far.

For now, in order to reduce confusion would I just want to return `None` if `trailing = true`?

piet-direct2d/src/lib.rs
hwchen: Perhaps. This was for my own clarity while reading the tests, but I'm happy to change.

piet-direct2d/src/lib.rs
hwchen: great, I'll find one.

piet-direct2d/src/lib.rs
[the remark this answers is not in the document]
hwchen: Implemented: removed explicit check.

piet-direct2d/src/lib.rs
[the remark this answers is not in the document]
hwchen: Implemented: remove explicit check, which will return the advance width.

piet-direct2d/src/lib.rs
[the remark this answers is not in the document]
hwchen: implemented: use leading edge of grapheme, and return `None` if `trailing = true` with notes in docstrings and comments indicating that it's not supported until BIDI.

piet-direct2d/src/lib.rs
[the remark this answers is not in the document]
hwchen: implemented with an `assert_close_to` test helper fn.

piet-cairo/src/lib.rs
```diff
@@ -524,4 +532,359 @@ impl TextLayout for CairoTextLayout {
     fn width(&self) -> f64 {
         self.font.text_extents(&self.text).x_advance
     }
+
+    // first assume one line.
+    // TODO do with lines
+    /// Hit Test for Point
```
raphlinus: The authoritative docstring should be on the trait (in piet crate). If there are impl-specific things to add, they can go here, but I think discoverability will be lower. (this comment applies to most of the new trait impls)

piet-direct2d/src/lib.rs
```diff
@@ -555,20 +559,372 @@ impl<'a> TextLayoutBuilder for D2DTextLayoutBuilder<'a> {
...
+    let mut utf16_count = 0;
+    for &b in s.as_bytes() {
+        if (b as i8) >= -0x40 {
+            utf16_count += 1;
+        }
+        if b >= 0xf0 {
+            utf16_count += 1;
+        }
+
+        // When count goes beyond text position, it means the start boundary of the utf16 code unit is passed.
```
raphlinus: I think this comment can be removed; it no longer accurately reflects the code, and didn't help me greatly with understanding.

piet-test/src/picture_5.rs
```diff
@@ -21,5 +21,19 @@ pub fn draw(rc: &mut impl RenderContext) -> Result<(), Error> {
     // underline text
     rc.stroke(Line::new((100.0, 52.0), (100.0 + width, 52.0)), &brush, 1.0);
 
+    // get hit test text position and draw a curos
```
raphlinus: "cursor"?

piet/src/text.rs
```diff
@@ -31,4 +32,46 @@ pub trait TextLayoutBuilder {
...
+pub struct HitTestTextPosition {
+    pub point: Point,
+    pub metrics: HitTestMetrics,
+}
+
+#[derive(Debug, Default, PartialEq)]
+pub struct HitTestMetrics {
+    pub text_position: usize,
+    //length: u32,
```
raphlinus: Probably should clean up these comments, possibly replace with something like "TODO: consider adding other metrics as needed, such as those provided in [DWRITE_HIT_TEST_METRICS](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/ns-dwrite-dwrite_hit_test_metrics)".

piet/src/text.rs
```diff
@@ -31,4 +32,46 @@ pub trait TextLayoutBuilder {
...
+/// - `is_trailing_hit` will always return `false` for now, until BIDI support is implemented.
+pub struct HitTestPoint {
+    pub metrics: HitTestMetrics,
+    pub is_inside: bool,
+
+    // removing until needed for BIDI or other.
+    //pub is_trailing_hit: bool,
 }
+
+/// return values for `hit_test_text_position`.
```
raphlinus: Ideally, link the function. Also, second line doesn't seem super useful.

piet/src/text.rs
```diff
@@ -31,4 +32,46 @@ pub trait TextLayoutBuilder {
...
+    /// Given a text position, determine the corresponding pixel location
+    /// (currently consider the text layout just one line)
+    fn hit_test_text_position(&self, text_position: usize, trailing: bool) -> Option<HitTestTextPosition>;
+}
+
+#[derive(Debug, Default, PartialEq)]
+/// return values for `hit_test_point`.
+/// - `metrics.text_position` will give you the text position.
+/// - `is_inside` indicates whether the hit test point landed within the text.
+/// - `is_trailing_hit` will always return `false` for now, until BIDI support is implemented.
```
raphlinus: Probably remove mention of `is_trailing_hit` given that it's commented out.

piet-test/src/picture_5.rs
```diff
@@ -21,5 +21,19 @@ pub fn draw(rc: &mut impl RenderContext) -> Result<(), Error> {
...
+    // this is for position three, not trailing
+    let hit_test_text_position = layout.hit_test_text_position(3, true);
```
raphlinus: The `true` here doesn't match the "not trailing" in the comment. I think it should be set to `false`, which would also have the side effect of making it actually draw a cursor :)

piet-test/src/picture_5.rs
```diff
@@ -21,5 +21,19 @@ pub fn draw(rc: &mut impl RenderContext) -> Result<(), Error> {
...
+
+    if let Some(http) = hit_test_text_position {
+        // TODO should the points be f64? directwrite only returns f32
+        let cursor_x = http.point.x as f64;
+        println!("cursor_x: {}", cursor_x);
```
raphlinus: Feels like maybe this should be removed, though I see why you have it during development.

piet-test/src/picture_5.rs
```diff
@@ -21,5 +21,19 @@ pub fn draw(rc: &mut impl RenderContext) -> Result<(), Error> {
...
+
+        let cursor_brush = rc.solid_brush(Color::rgba8(0x80, 0x80, 0x80, 0xF0));
```
raphlinus: The drawing operations can now take `&Color` directly, you don't need to call the solid brush constructor. Probably we didn't update the example code you adapted this from, but for new code it should use the cleaner form.

piet-cairo/src/grapheme.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        let (text_position, _) = UnicodeSegmentation::grapheme_indices(self.text.as_str(), true)
+            .nth(grapheme_position)?;
+
+        let curr_edge = self.hit_test_text_position(text_position, false)?;
+        // text_position logic will automatically round to next grapheme boundary
+        // when text position is not on a grapheme boundary
+        let next_edge = self.hit_test_text_position(text_position + 1, false)?;
+
+        res.curr_idx = curr_edge.metrics.text_position;
```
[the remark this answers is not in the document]
hwchen: agreed; I think I've reworked the logic so it no longer makes sense to first set a default.

piet-test/src/picture_5.rs
hwchen: the `trailing` arg is set to be removed.

piet/src/text.rs
hwchen: done for both HitTestPoint and HitTestTextPosition.

piet-cairo/src/lib.rs
hwchen: Improved docs

piet/src/text.rs
```diff
@@ -28,7 +29,54 @@ pub trait TextLayoutBuilder {
...
+}
+
+/// return values for [`hit_test_text_position`](../piet/trait.TextLayout.html#tymethod.hit_test_text_position).
+#[derive(Debug, Default)]
+pub struct HitTestTextPosition {
+    pub point: Point,
+    pub metrics: HitTestMetrics,
+}
+
+#[derive(Debug, Default, PartialEq)]
```
raphlinus: Still missing docstring for this (one-liner is ok).

piet/src/text.rs
```diff
@@ -28,7 +29,54 @@ pub trait TextLayoutBuilder {
...
     fn width(&self) -> f64;
+
+    /// Given a Point, determine the corresponding text position
+    fn hit_test_point(&self, point: Point) -> HitTestPoint;
+
+    /// Given a text position, determine the corresponding pixel location
+    /// (currently consider the text layout just one line)
+    ///
+    /// Note: if text position is not at grapheme boundary, `cairo` and `web` will find the next text position/grapheme
+    /// boundary. This behavior is different than `directwrite`, which will panic.
```
raphlinus: I posted this comment to Zulip also, but pasting here for posterity:

Really? I find that pretty hard to believe, and also super unfriendly behavior. Panic is OK if it's not at a *code point* boundary (btw, there's a [convention](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure) to use `# Panics` for this), but there are lots of legit ways a cursor can end up on a non-grapheme boundary. One I mentioned is pasting a lone [RIS](https://en.wikipedia.org/wiki/Regional_Indicator_Symbol) at the start of a RIS sequence.

piet/src/text.rs
```diff
@@ -28,7 +29,59 @@ pub trait TextLayoutBuilder {
     fn build(self) -> Result<Self::Out, Error>;
 }
 
+/// # Text Layout
+///
+/// A text position corresponds roughly to unicode code units. Because piet uses rust utf-8
+/// strings, this means that text position corresponods roughly with byte index.
```
raphlinus: "corresponds"

Also I think we can sharpen up the language. "Text position is defined in utf-8 code units, as is standard for Rust strings."

piet/src/text.rs
hwchen: Fixed doc comment, plus improved handling of non-grapheme-boundary text positions.

piet-cairo/src/lib.rs
```diff
@@ -524,4 +532,359 @@ impl TextLayout for CairoTextLayout {
...
+        assert_close_to(layout.hit_test_text_position(10, false).unwrap().point.x, test_layout_2.width(), 3.0);
+        assert_close_to(layout.hit_test_text_position(14, false).unwrap().point.x, layout.width(), 3.0);
+
+        // note code unit not at grapheme boundary
+        assert_close_to(layout.hit_test_text_position(1, false).unwrap().point.x, test_layout_0.width(), 3.0);
+        assert_eq!(layout.hit_test_text_position(1, false).unwrap().metrics.text_position, 2);
+    }
+
+    #[test]
+    fn test_hit_test_point_basic() {
```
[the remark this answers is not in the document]
hwchen: used #[cfg]

Issue #33, opened 2019-01-26
Unresolved import `direct2d::math`


canadaduane:
I am (probably naively) trying to run "cargo build" on a fresh pull of `piet` and I'm seeing several errors in the build, e.g.:

```
error[E0432]: unresolved import `direct2d::math`
 --> piet-direct2d/src/conv.rs:3:15
  |
3 | use direct2d::math::{Matrix3x2F, Point2F, RectF};
  |               ^^^^ could not find `math` in `direct2d`

error[E0433]: failed to resolve: could not find `shared` in `winapi`
  --> piet-direct2d/src/lib.rs:11:13
   |
11 | use winapi::shared::basetsd::UINT32;
   |             ^^^^^^ could not find `shared` in `winapi`

error[E0433]: failed to resolve: could not find `um` in `winapi`
  --> piet-direct2d/src/lib.rs:12:13
   |
12 | use winapi::um::dcommon::D2D_SIZE_U;
   |             ^^ could not find `um` in `winapi`
```

I assume that as a new user of rust I just don't know what I'm doing :)

raphlinus:
This looks like trying to build windows-specific code on a non-windows platform. That shouldn't be happening. What platform are you on? Are you setting any features explicitly?

canadaduane:
I'm on Mac OS. I just checked the repo out and did "cargo build" in the
root, so I'm not aware of setting any features.

raphlinus:
Ah, got it. Building the workspace tries to build all the subcrates, including direct2d, which isn't going to work. What should work is `cargo run --example basic-cairo`, assuming the Cairo prerequisites have been installed. We should probably update the README to reflect this.

We could also update the piet-direct2d crate so that nothing gets built on non-windows platforms.

canadaduane:
Thanks, that's exactly what I was looking for.

SimonSapin:
Specifying `default-members` in the workspace’s manifest could help: https://doc.rust-lang.org/cargo/reference/manifest.html#package-selection

PR #97, opened 2019-10-31
Add instructions for running with cairo

dianarg:
- Related to #9
- Related to #33


dianarg:
Hi everyone,
I tested this on Windows 10 and Ubuntu Linux.  Getting the cairo library on Windows was slightly more complicated, but the release rylev mentioned worked fine for me (https://github.com/preshing/cairo-windows).

Cargo.toml
```diff
@@ -9,3 +9,12 @@ members = [
...
 ]
+
+default-members = [
+    "piet",
+    "piet-cairo",
+    "piet-common",
+    "piet-test",
+    "piet-web",
+    "piet-web/examples/basic"
+]
```
cmyr: What's the significance of this change?

cmyr:
Thanks for this, cleanup is good and documentation is even better. One thing I would like to make sure is that this doesn't influence what gets built in CI; additionally we might want to also be building cairo in CI on windows? I'm not sure how much of a headache that would be.

Cargo.toml
dianarg: Yes, that's where it came from.

dianarg:
I haven't used Azure before, but it looks like what runs in CI is controlled by ci/test.yml?  Since the commands already have `--all`, I think it's already building everything (except cairo) for Windows.

https://github.com/linebender/piet/blob/02207a9e74034ae3ea937f3a1b165a96424f930c/ci/test.yml#L16

For adding cairo, what do you think about just adding a curl line or similar to install-rust.yml to download the prebuilt release listed above?

Cargo.toml
dianarg: Perhaps it also makes sense to remove piet-cairo from this list and have users explicitly choose a backend with --features.

cmyr:
okay, `--all` sounds good. And yea, we would do something like download prebuilt binaries. I don't think that's critical though, and it doesn't need to be part of this PR; if you want to play around with it you're more than welcome but no pressure. :)

PR #100, opened 2019-11-05
[text] improve docs on hit test point

hwchen:
Add notes on `is_inside` and behavior of testing point outside of text bounds.


piet/src/text.rs
```diff
@@ -42,10 +42,15 @@ pub trait TextLayout {
     /// Measure the advance width of the text.
     fn width(&self) -> f64;
 
-    /// Given a Point, determine the corresponding text position
+    /// Given a Point, determine the corresponding text position.
+    ///
+    /// `HitTestPoint` field `is_inside` returns true if the tested point falls within the bounds of the text.
```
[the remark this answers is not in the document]
hwchen: Thanks, now I understand what you're going for. I'll revise using this as a starting point.

One question on documentation: Do we prefer to document on the fn, the return value, or both?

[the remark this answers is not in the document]
cmyr: My rule of thumb is "in the general case, everything public should have some docs".

Personally: I would document each item in each struct, explaining what it communicates. For the functions, I would focus on documenting _how_ to use them, what the behaviour is, what happens in less common cases, and things like that.

Thanks for being a sport about this! It's everyone's least favourite part, but it's such an important thing, especially for less-obvious API like this one. It's important to remember that most people encountering this API are going to have very little understanding of the domain.

hwchen:
About this API; I don't have particularly strong feelings at the moment, but I also don't have the experience to fully understand upcoming changes. I'd defer to you or @raphlinus, if you'd like to reduce the nesting of the return values I'm all for it.

(My one thought is that I do prefer a flatter hierarchy generally, even if that means some code duplication. But as I mentioned, I don't understand this domain enough to know if it's ok to apply that bias.)

cmyr:
What was the motivation for the current design? Is it borrowing from an existing API?

hwchen:
Yes, it’s because my starting point was directwrite, I basically copied its data structures and then started deleting unneeded fields. I was discussing with Raph, that directwrite was a good starting point for structuring the overall api, but now I understand that we don’t have to hew so closely to its details.

Anyways, long story short, I’m looking forward to discovering what api piet’s users need.

piet/src/text.rs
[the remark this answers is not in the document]
hwchen: No worries. It does take a lot of energy for me to write well, but I agree it's important. I've made a second pass, and happy to take more feedback. I think it's moving towards the amount of detail necessary.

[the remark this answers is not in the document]
hwchen: No worries! It takes a lot of energy for me to write, but it's definitely worth it.

I've done another pass, I think this is closer to the amount of detail you were demonstrating. Happy to have more feedback.

piet/src/text.rs
```diff
@@ -42,17 +45,46 @@ pub trait TextLayout {
     /// Measure the advance width of the text.
     fn width(&self) -> f64;
 
-    /// Given a Point, determine the corresponding text position
+    /// Given a `Point`, determine the corresponding text position.
+    ///
+    /// ## Return value:
+    /// Returns a `HitTestPoint` describing the results of the test.
+    ///
+    /// `HitTestPoint` field `is_inside` returns true if the tested point falls within the bounds of the text, `false` otherwise.
```
cmyr: probably 'is `true`', not 'returns true'.

piet/src/text.rs
```diff
@@ -42,17 +45,46 @@ pub trait TextLayout {
     /// Measure the advance width of the text.
     fn width(&self) -> f64;
 
-    /// Given a Point, determine the corresponding text position
+    /// Given a `Point`, determine the corresponding text position.
+    ///
+    /// ## Return value:
+    /// Returns a `HitTestPoint` describing the results of the test.
```
cmyr: It's actually probably worth making these links to the structs in question; I saw you doing this below for some other things and it's helpful.

I also find that for the purposes of reading the docs in the code itself (as opposed to the rendered HTML docs) it can be nice to put all the links down at the bottom of the comment, like:

```rust
/// link to [`HitTestPoint`]
/// 
/// at the bottom of the whole comment, after a newline, add:
///
/// [`HitTestPoint`]: struct.HitTestPoint.html
```

If we do that, maybe we can just document the fields on the struct? That might be simpler? It would save us having to document these twice for these two functions, at least.

piet/src/text.rs
```diff
@@ -42,17 +45,46 @@ pub trait TextLayout {
...
+    ///
+    /// `HitTestTextPosition` field `metrics` is a `HitTestMetrics` struct. `HitTestMetrics` field `text_position` is the original text position (unless out of bounds).
     ///
-    /// Note: in directwrite, if text position is not at code point boundary, this method will
-    /// panic. Cairo and web are more lenient and may not panic.
+    /// ## Notes:
+    /// In directwrite, if text position is not at code point boundary, this method will panic.
+    /// Cairo and web are more lenient and may not panic.
+    ///
+    /// For text position that is greater than `text.len()`, web/cairo will return the
```
cmyr: For 'a' text position, maybe?

piet/src/text.rs
```diff
@@ -68,6 +100,9 @@ pub struct HitTestPoint {
 }
 
 /// return values for [`hit_test_text_position`](../piet/trait.TextLayout.html#tymethod.hit_test_text_position).
+/// - `metrics.text_position` will give you the text position.
+/// - `point` is the width offset of the leading edge of the grapheme cluster containing the text
```
cmyr: I would say something like, "the point's `y` value is the position of the leading edge of ..."

piet/src/text.rs
```diff
@@ -68,6 +100,9 @@ pub struct HitTestPoint {
 }
 
 /// return values for [`hit_test_text_position`](../piet/trait.TextLayout.html#tymethod.hit_test_text_position).
+/// - `metrics.text_position` will give you the text position.
```
cmyr: I would put these docs on the individual fields themselves.

piet/src/text.rs
hwchen: Thanks for the example, that really cleans things up.

I'm going to add the links, but for now I'm going to keep the field documentation in both the method and the return struct. My reasoning is that it's a little harder to digest the docs if you have to flip back and forth between pages.

Open to changing in the future.

piet/src/text.rs
hwchen: thanks, that's better wording (although I think it's the `x` in this case)

hwchen:
@cmyr of course, and thanks again for the feedback! Good documentation definitely makes this feel more "real".

PR #103, opened 2019-11-07
[text] update TextLayout trait for linebreaking, (nullrender)

hwchen:
Line breaking:

- [ ] confirm trait methods and basic docs
- [ ] update directwrite
- [ ] update cairo
- [ ] update web


hwchen:
@cmyr @raphlinus just wanted to start by putting up the draft methods and docs for linebreaking. Feel free to comment on everything, including the `TODO`s I have.

piet/src/text.rs
```diff
@@ -41,10 +62,29 @@ pub trait TextLayoutBuilder {
...
 pub trait TextLayout {
     /// Measure the advance width of the text.
     fn width(&self) -> f64;
 
+
+    /// Used for changing the width of a text layout. Given a width, returns a [`TextLayout`]
+    /// struct with recalculated lines and line metrics.
+    // TODO: Should this take &self or self? I'd say maybe; perhaps the old struct should be kept around
+    // to help with caching.
+    fn update_width(&self, new_width: f64) -> Self;
```
cmyr: Interesting question.

A lot comes down to how we expect this to work under the hood. Based on the name of the function, I would expect the signature,

```rust
fn update_width(&mut self, new_width: f64);
```

That is, it would mutate the existing `TextLayout` to reflect the new width.

If we think that the caller might want to keep around multiple `TextLayout`s, then I could imagine the signature you use here, but I would call it something like `new_layout_with_width`.

I would only take `self` by value if there was expensive internal state that we wanted to reuse between the old and new layouts, or if we wanted the API to enforce some contract whereby the previous layout is unusable after you have a new one.

piet/src/text.rs
```diff
@@ -41,10 +62,29 @@ pub trait TextLayoutBuilder {
...
+
+    /// Given a line number, return a reference to that line's underlying string.
+    // TODO can I return &str without viral lifetime?
+    fn line_text(line_number: usize) -> String;
```
cmyr: assuming this has a `&self` parameter, I _think_ you should be able to do `fn line_text(&self, line_number: usize) -> &str`, without needing a lifetime in the trait itself.

In addition we _might_ want this to return a `Cow<str>`, in case we want to allow more flexible text storage at some point in the future; I'm thinking specifically if we want to support a rope. But I think we can break the API later, and &str is fine for now.

hwchen: Whoops, I knew returning a reference should have worked, not sure why I forgot `&self` :)

I'll keep the `Cow<str>` in mind.

piet/src/text.rs
hwchen: This is something I discussed with @raphlinus, having more of an immutable feel.

My main thought for using `self` was to enforce a contract, but I couldn't think of any.

So, I'll keep `&self` for now, but we'll see what the future holds.

Thanks for your thoughts, they're a great articulation of things I have a fuzzy grasp of but can't articulate very well yet.

piet/src/text.rs
```diff
@@ -94,6 +132,34 @@ pub trait TextLayout {
...
 }
 
+/// Metadata about each line in a text layout.
+#[derive(Debug, Default)]
+pub struct LineMetric {
+    /// Line length (in code units), but offset from the beginning of the text. So it's the length
+    /// of this line summed with the lengths of all the lines before it.
+    ///
+    /// Does not include trailing whitespace.
+    pub line_length_offset: usize,
```
raphlinus: I think the name here is confusing, maybe something more like `line_end_offset`.

piet/src/text.rs
```diff
@@ -94,6 +132,34 @@ pub trait TextLayout {
...
+
+    /// Line width, but offset from the beginning of the text. So it's the width
+    /// of this line summed with the widths of all the lines before it.
+    ///
+    /// Does not include trailing whitespace.
+    pub line_width_offset: f64,
```
raphlinus: Similarly, maybe `cumulative_width`. We can definitely discuss the name.

piet/src/text.rs
```diff
@@ -94,6 +132,34 @@ pub trait TextLayout {
...
+
+    /// Line length (in code units), but offset from the beginning of the text. So it's the length
+    /// of this line summed with the lengths of all the lines before it.
+    ///
+    /// Includes trailing whitespace.
+    pub line_length_trailing_whitespace_offset: usize,
```
raphlinus: This particular one, if we're following DWrite, can maybe just be the size of the trailing whitespace. Otherwise, it should be the "trimmed line end" (ie line end - whitespace size).

cmyr:
How is this coming along?

hwchen:
I got mired in updating druid and Piet to raw d2d. That should be finished up soon, then I'll be updating hit testing to raw d2d, then I'll get back to multiline

cmyr:
cool, sounds good! Just wanted to make sure it wasn't being neglected by us. :)

hwchen:
Closed because of refactor to use raw_d2d.

See #134 for next effort.

PR #104, opened 2019-11-08
Do addition and subtraction in binary search

futurepaul:
Currently we have a couple of problems with non-ASCII characters in Druid's textbox (as pointed out in [druid#256](https://github.com/xi-editor/druid/pull/256) that trace back to this loop: this can get stuck in an infinite loop, and this can overflow when subtracting.

By changing these branches to += 1 / -=1 (as proposed in the comments) I'm not seeing those problems anymore, but I'm not 100 percent sure we're out of the woods.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -594,9 +594,9 @@ impl TextLayout for CairoTextLayout {
             // since it's not a hit, check if closer to start or finish
             // and move the appropriate search boundary
             if point.x < grapheme_bounds.leading {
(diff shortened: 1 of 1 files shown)


hwchen:
I think that this makes it not a binary search anymore?

I'm also going to take a look at this and see how it's caught in a loop. I have one test already with relatively extensive hit test point testing with non-ascii chars, but it may not have been enough.

futurepaul:
Ah thanks for jumping into this and you're right I think this disqualifies it as a binary search!

PR #105, opened 2019-11-09
[text] fix hanging in hit test point binary search

hwchen:
Hit test point was hanging (infinte loop in binary search). This should fix.

The issue was mixing up grapheme index and text position when updating the binary search slice.

This should hopefully provide a complete fix for the issue.

refer: #104

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -594,9 +594,9 @@ impl TextLayout for CairoTextLayout {
             // since it's not a hit, check if closer to start or finish
             // and move the appropriate search boundary
             if point.x < grapheme_bounds.leading {
-                right = grapheme_bounds.curr_idx as usize; // should this be -1?
+                right = middle;
             } else if point.x > grapheme_bounds.trailing {
-                left = grapheme_bounds.curr_idx as usize; // should this be +1?
+                left = middle;
             }
         }
     }
@@ -1004,7 +1004,7 @@ mod test {
 
     #[test]
     #[cfg(target_os = "linux")]
-    fn test_hit_test_point_complex() {
+    fn test_hit_test_point_complex_0() {
         // Notes on this input:
         // 6 code points
         // 7 utf-16 code units (1/1/1/1/1/2)
@@ -1053,7 +1053,7 @@ mod test {
 
     #[test]
     #[cfg(target_os = "macos")]
-    fn test_hit_test_point_complex() {
+    fn test_hit_test_point_complex_0() {
         // Notes on this input:
         // 6 code points
         // 7 utf-16 code units (1/1/1/1/1/2)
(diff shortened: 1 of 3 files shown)


hwchen:
I guess I was unlucky in my previous tests; I would have expected to catch this, but didn't. I added another test, based on reports from @futurepaul and @scholtzan where they reported hanging.

futurepaul:
I really had to jam on it to get it to hang and / or crash, haven't found an easy reproduction yet (it seems like the problem was at certain multiples). My guess is the subtraction underflow is somehow `left` becoming > than `right`, but I don't know why that was happening.

I did similar "jamming" with this PR and haven't been able to get it to panic or hang.

piet-cairo/src/lib.rs
```diff
@@ -594,9 +594,9 @@ impl TextLayout for CairoTextLayout {
             // since it's not a hit, check if closer to start or finish
             // and move the appropriate search boundary
             if point.x < grapheme_bounds.leading {
-                right = grapheme_bounds.curr_idx as usize; // should this be -1?
+                right = middle;
             } else if point.x > grapheme_bounds.trailing {
-                left = grapheme_bounds.curr_idx as usize; // should this be +1?
+                left = middle;
```
cmyr: I haven't looked super closely, but just looking at this if/else it isn't clear to me that there can't be cases where we don't match either condition; that is, I can't tell if this is exhaustive, or not.

If it isn't exhaustive, we still risk an infinite loop.

If you're confident it's exhaustive, you might consider adding a final `else { unreachable!("hit_test_point conditional is exhaustive"); }` or similar, or you could consider logging an error and returning `HitTestPoint::default()`.

hwchen:
I'm fairly confident, but it does make sense to put in a branch to make it
exhaustive. I also wonder if it's worth it to put in some quick check /
fuzzing type tests.

cmyr:
I'm not totally sure how feasible quickcheck or fuzzing would be here, but that isn't a crazy idea by any means.

hwchen:
I'm going to continue thinking about what tests would be good to expose edge cases.

In the meantime, I can add a comment on why I think the conditional is exhaustive.

I think the conditions are exhaustive because:
- since outliers are already rejected by this line in the method, the tested point must exist within the width of the text.
- and since testing whether the point is within a grapheme boundaries inspects the range, from leading to trailing, inclusive of the boundaries.
- and since the conditionals for moving the search boundaries check whether the tested point is outside grapheme boundaries, exclusive of the grapheme boundaries.
- it follows that, these conditions together have full coverage of the points within the width of the text.
- therefore, it should be impossible that there is another condition that's reachable by the test point.

However, I don't think that this necessarily rules out infinite loops in general; just that it wouldn't result from not checking the location of a point.

The other place I could think of possibly causing an infinte loop would be an incorrect `middle` assignment never reaching the correct end condition. This should only apply to an almost-base case (2 chars) since the other assignments involving `middle` should just shrink the search space.

Upon further inspection, because I'm using integer division, the loop on its own could get stuck in an infinite cycle with two chars. However, because I explicitly check for `beginning` and `end` of text before the loop, I bypass :sweat_smile:  

I'm fixing this in a commit by bumping the assignment by 1, when `middle` is assigned to `left` (deals with middle always being `0` for two chars, can never reach `1`). Just in case somebody removes the early `beginning` and `end` checks.

hwchen:
@cmyr let me know what else you think should be necessary before merging this pr. I think that perhaps adding some notes on the code is ok; but don't know how detailed you'd want it, especially since "unreachable" already provides a hint.

cmyr:
Yea, I think that `unreachable` at least communicates that you believe this is exhaustive, and means we'll find out if it isn't. 😁

PR #106, opened 2019-11-11
Download cairo library in azure

dianarg:
Creating a PR to trigger azure to run; let me know if there's a way to do this on my fork.

The eventual goal is to be able to build piet-cairo for Windows in CI.


cmyr:
I think we're okay without testing cairo on windows in CI, but thanks for this and the other patch!

PR #110, opened 2019-11-18
add console_error_panic_hook to help debug panics in web example.

bdon:
The test page wasn't working for me - I needed to add this crate to diagnose via stack trace. LMK if you need any changes.

### DIFF
--- a/piet-web/examples/basic/Cargo.toml
+++ b/piet-web/examples/basic/Cargo.toml
@@ -13,6 +13,7 @@ piet-web = { path = "../.." }
 piet-test = { path = "../../../piet-test" }
 
 wasm-bindgen = "0.2.30"
+console_error_panic_hook = "0.1.6"
 [dependencies.web-sys]
 version = "0.3.10"
 features = ["console", "CanvasRenderingContext2d", "Window", "Document", "Element", "HtmlElement", "HtmlCanvasElement"]
--- a/piet-web/examples/basic/src/lib.rs
+++ b/piet-web/examples/basic/src/lib.rs
@@ -9,8 +9,12 @@ use piet_web::WebRenderContext;
 
 use piet_test::draw_test_picture;
 
+extern crate console_error_panic_hook;
+use std::panic;
+
 #[wasm_bindgen]
 pub fn run() {
+    panic::set_hook(Box::new(console_error_panic_hook::hook));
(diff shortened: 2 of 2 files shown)


cmyr:
Hmm, both master and this branch fail to show me anything on latest safari. If I go back to 02207a9 I get the little demo. What problem were you running into, that this fixed?

bdon:
piet-web/examples/basic/src/lib.rs

I was getting a panic from test::test_hit_test_point_basic_0 on master, haven't looked more deeply into it though - will investigate more

hwchen:
Those “tests” are definitely brittle. It worked for me, I wrote/tested on Firefox. I don’t yet have a better place to put tests for the web backend, @cmyr let me know if you have thoughts.

I haven’t yet had the chance to look at the behavior of this pr, but if it allows better handling of the tests that would be great.

bdon:
@hwchen it's failing for me on Firefox, what OS are you on? I wonder if it has to do with what font gets loaded by this line depending on platform:

https://github.com/linebender/piet/blob/master/piet-web/examples/basic/src/lib.rs#L88

hwchen:
yup, failing depending on font loaded could definitely be a thing. I'm on Linux.

I've talked with Raph about this, and it would be nice to be able to load font assets and use feature gates to better handle testing in CI. But at this point it's something for the future.

Suggestions for dealing with this are definitely welcome.

bdon:
What about shipping the test suite with some simple but non-trivial (variable width) WOFF font? seems like the most future proof way to build out more sophisticated tests.

hwchen:
@bdon That sounds reasonable, but I'm not sure when I'll get to it.

For now, I'd like to just figure out the best way to have the example exhibit reasonable behavior. I'm not sure this PR makes sense in the context of just making sure that the example works, since users exploring piet won't expect an example to fail at all.

Perhaps the simplest solution for now is to put in a feature flag for the "tests" block in the example to be run only for testing purposes.

Let me know what you think.

cmyr:
although I think wasm is very cool, the piet-web backend is not something I have a lot of time to spend on, unfortunately. If someone wanted to step up and take ownership of this work they would have our full support.

bdon:
I'll close this until I have a chance to look into the tests more!

PR #111, opened 2019-11-24
Raw d2d

hwchen:
WIP for finishing integration of raw d2d bindings.

- move d3d api into `piet-direct2d`
- integrate `piet-common` with `piet-direct2d`
- add raw_d2d hit testing
- add docstrings indicating that DWriteFactory, DeviceContext, and D2DFactory are to be used only for system integration in druid-shell, not for regular end-users.

Note: on my maching (windows 10 in virtualbox) the test fails for `piet-direct2d`, not sure why.

### DIFF
--- a/piet-common/src/direct2d_back.rs
+++ b/piet-common/src/direct2d_back.rs
@@ -2,14 +2,20 @@
 
 use std::fmt;
 
-use direct2d::enums::BitmapOptions;
-use direct2d::image::Bitmap;
-use direct2d::render_target::RenderTag;
-use direct2d::RenderTarget;
-use direct3d11::flags::{BindFlags, CreateDeviceFlags};
-use direct3d11::helpers::ComWrapper;
-use dxgi::flags::Format;
+// --- old block of imports
+//use direct2d::enums::BitmapOptions;
+//use direct2d::image::Bitmap;
+//use direct2d::render_target::RenderTag;
+//use direct2d::RenderTarget;
+//use direct3d11::flags::{BindFlags, CreateDeviceFlags};
+//use direct3d11::helpers::ComWrapper;
+//use dxgi::flags::Format;
 
+// --- new block of imports
+// d3d11 interface, enough for drawing d2d
+use piet_direct2d::d3d::{D3D11Device, D3D11DeviceContext, D3D11Texture2D, DXGI_MAP_READ, TextureMode};
+
+// -- end change in import block
 use piet::{ErrorKind, ImageFormat};
 
 #[doc(hidden)]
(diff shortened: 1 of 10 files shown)


piet-common/src/direct2d_back.rs
```diff
@@ -2,14 +2,20 @@
...
 
-use direct2d::enums::BitmapOptions;
-use direct2d::image::Bitmap;
-use direct2d::render_target::RenderTag;
-use direct2d::RenderTarget;
-use direct3d11::flags::{BindFlags, CreateDeviceFlags};
-use direct3d11::helpers::ComWrapper;
-use dxgi::flags::Format;
-
+// --- old block of imports
```
raphlinus: These should be deleted.

piet-common/src/direct2d_back.rs
```diff
@@ -21,7 +27,7 @@ pub type Piet<'a> = D2DRenderContext<'a>;
 /// The associated brush type for this backend.
 ///
 /// This type matches `RenderContext::Brush`
-pub type Brush = GenericBrush;
+//pub type Brush = GenericBrush;
```
raphlinus: The doc comments will apply to the next item; delete.

piet-common/src/direct2d_back.rs
```diff
@@ -165,30 +110,28 @@ impl Device {
 
     /// Create a new bitmap target.
     pub fn bitmap_target(
-        &self,
+        &mut self, //TODO this changed to mut, double check that this it ok
```
raphlinus: I think so. There's a tendency for wrapped com to be `&self` with interior mutability, but that's because of technical details. This feels like a case where `&mut` is the right type, but if doc for the underlying device was explicitly marked as thread-safe, we might relax it.

(It's worth noting that taking a `&self` and calling a non-thread-safe function is a potential safety violation. So `&mut self` is always more conservative if you're not sure)

piet-direct2d/examples/basic.rs
```diff
@@ -101,3 +106,124 @@ fn main() {
     )
     .unwrap();
 }
+
```
raphlinus: It looks like a lot of this now exists in d3d.rs, so it's duplication. I'm not sure it's worth blocking the PR to resolve that, but maybe a TODO, and this can be followup for an interested person.

piet-direct2d/src/d2d.rs
```diff
@@ -0,0 +1,638 @@
...
+impl DeviceContext {
+    /// Create a new device context from an existing COM object.
+    ///
+    /// Marked as unsafe because the device must be in a good state.
+    /// This *might* be overly conservative.
+    pub unsafe fn new(ptr: ComPtr<ID2D1DeviceContext>) -> DeviceContext {
+        DeviceContext(ptr)
+    }
+
+    /// docs
```
raphlinus: "Get the raw pointer."

I don't think this is actually unsafe. Of course, the caller wouldn't be calling it unless they wanted to do unsafe things with it, but this is a fairly clear distinction in Rust.

piet-common/src/direct2d_back.rs
hwchen: resolved, had to also update cairo backend's `bitmap_target` to take `&mut self`, as well as png example.

piet-common/src/direct2d_back.rs
hwchen: I forgot why I commented out `Brush` and `Bitmap`; I think it was a temporary thing to get stuff to compile. Perhaps when we did a rebase?

In any case, on closer inspection, I can export `Brush` and `Bitmap` from `piet_direct2d`, would that be preferred?

piet-direct2d/examples/basic.rs
hwchen: added a comment to the example.

piet-common/src/direct2d_back.rs
raphlinus: I *think* the answer here is that they're exported as associated types of `RenderContext`, and that should be considered their "true names" for the purpose of other code using them. Otherwise they can be considered internal details. So I think the right thing to do is delete them.

We will probably come back to the question of exporting platform-specific capabilities (especially the ability to create D2D compatible images from other sources than memory buffers). But I don't think that just exporting the `Brush` or `Bitmap` type helps us all that much.

hwchen: Ok, Brush and Bitmap deleted.

PR #112, opened 2019-11-24
Partial SVG backend

Ralith:
Lightly tested; produces output similar to the cairo backend for test images 1, 3 and 4. Attempts to use text and images will produce `NotSupported` error; the `unimplemented` invocations are all statically unreachable as a result.

For some reason I couldn't get SVG line elements to display, but paths work fine and are more concise anyway.

### DIFF
--- a/piet-svg/Cargo.toml
+++ b/piet-svg/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "piet-svg"
+version = "0.0.7"
+authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>"]
+edition = "2018"
+license = "MIT/Apache-2.0"
+repository = "https://github.com/linebender/piet"
+
+[dependencies]
+piet = { version = "0.0.7", path = "../piet" }
+svg = "0.6"
+
+[dev-dependencies]
+piet-test = { version = "0.0.7", path = "../piet-test" }
--- a/piet-svg/src/lib.rs
+++ b/piet-svg/src/lib.rs
@@ -1,0 +1,490 @@
+//! SVG output support for piet
+//!
+//! Text and images are unimplemented and will always return errors.
+
+use std::borrow::Cow;
+use std::{io, mem};
+
+use piet::kurbo::{Affine, Point, Rect, Shape};
+use piet::{
+    new_error, Color, Error, ErrorKind, FixedGradient, HitTestPoint, HitTestTextPosition,
+    ImageFormat, InterpolationMode, IntoBrush, LineCap, LineJoin, StrokeStyle,
+};
+use svg::node::Node;
+
+type Result<T> = std::result::Result<T, Error>;
+
+/// `piet::RenderContext` for generating SVG images
+pub struct RenderContext {
+    stack: Vec<State>,
+    state: State,
+    doc: svg::Document,
+    next_id: u64,
+    text: Text,
+}
+
+impl RenderContext {
+    /// Construct an empty `RenderContext`
+    pub fn new() -> Self {
+        Self {
+            stack: Vec::new(),
+            state: State::default(),
+            doc: svg::Document::new(),
+            next_id: 0,
+            text: Text(()),
+        }
+    }
+
+    /// Write graphics rendered so far to an `std::io::Write` impl, such as `std::fs::File`
+    ///
+    /// Additional rendering can be done afterwards.
+    pub fn write(&self, writer: impl io::Write) -> io::Result<()> {
+        svg::write(writer, &self.doc)
+    }
+
+    fn new_id(&mut self) -> Id {
+        let x = Id(self.next_id);
+        self.next_id += 1;
+        x
+    }
+}
+
+impl piet::RenderContext for RenderContext {
+    type Brush = Brush;
+
+    type Text = Text;
+    type TextLayout = TextLayout;
+
+    type Image = Image;
+
+    fn status(&mut self) -> Result<()> {
+        Ok(())
+    }
+
+    fn clear(&mut self, color: Color) {
+        let brush = color.make_brush(self, || Rect::ZERO);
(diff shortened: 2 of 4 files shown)


Ralith:
Implementing `TextLayout` faithfully would present some challenges because we're not *actually* rendering anything. At minimum, I think we'd need to pull in a text shaping engine (perhaps behind a feature flag) and disclaim any prospect of correctness if the document isn't displayed using exactly the same fonts that were specified when generating the file.

Images present a different challenge: we can embed base64'd png or jpeg (or other?) data, but calling out to an image encoder when someone calls `make_image` is pretty weird. This could also be behind a feature flag, and configured (to e.g. select codec and compression level) via inherent methods.

piet-svg/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+        if let Some(id) = self.clip {
+            node.assign("clip-path", format!("url(#{})", id.to_string()));
+        }
+        if let Some((ref brush, rule)) = self.fill {
+            node.assign("fill", brush.val());
+            if let Some(rule) = rule {
+                node.assign("fill-rule", rule);
+            }
+        } else {
+            node.assign("fill", "transparent");
```
RazrFalcon: There are no `transparent` in SVG. You should use `none`.

piet-svg/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+            .set("r", circle.radius);
+        attrs.apply_to(&mut x);
+        node.append(x);
+    } else if let Some(rect) = shape.as_rounded_rect() {
+        let mut x = svg::node::element::Rectangle::new()
+            .set("x", rect.origin().x)
+            .set("y", rect.origin().y)
+            .set("width", rect.width())
+            .set("height", rect.height())
+            .set("rx", rect.radius());
```
RazrFalcon: You should set `ry` too for compatibility.

piet-svg/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+            }
+            FixedGradient::Radial(x) => {
+                let mut gradient = svg::node::element::RadialGradient::new()
+                    .set("gradientUnits", "userSpaceOnUse")
+                    .set("id", id)
+                    .set("cx", x.center.x)
+                    .set("cy", x.center.y)
+                    .set("fx", x.center.x + x.origin_offset.x)
+                    .set("fy", x.center.y + x.origin_offset.y)
+                    .set("fr", 0)
```
RazrFalcon: There are no `fr` in SVG.

Ralith: Yes, there is: https://svgwg.org/svg2-draft/pservers.html#RadialGradientElementFRAttribute

Oddly, MDN disagrees with the spec on the default value (50% vs 0% respectively).

RazrFalcon: I see. I should clarify that I was talking about SVG 1.1. Which is a better target, since probably no one supports SVG 2 anyway.

SVG 2 is not finished yet, so everything is possible.

Ralith: Fair enough.

Ralith:
A better solution to the text issue might be to render text as paths, for which there at least exists an authoritative answer to the queries. Shame not to have working copy/paste though.

piet-svg/examples/demo.rs
```diff
@@ -0,0 +1,18 @@
+//! Basic example of rendering to a SVG
```
raphlinus: Naming doesn't follow the pattern, this should be "basic-svg.rs". That said, I agree the naming scheme is not ideal, and the d2d one also doesn't follow the pattern.

Ralith:
Thanks for the review!

> From what I can tell, images are largely a function of resource linking. But these could be done as base64 data uri's?

We can easily add backend-specific methods for linking external resources, and probably that's what we should direct users towards here. My concern is mainly about implementing `make_image`, which consumes raw bitmap data. Pulling in an image codec just to support a really ugly and inefficient embedding is unfortunate, and hardwiring any one codec would tend to give bad results for many inputs. Is there maybe a way to encode raw bitmap data in a data URI directly?

RazrFalcon:
>Is there maybe a way to encode raw bitmap data in a data URI directly?

AFAIK, SVG allows only `image/png`, `image/jpeg` and `image/svg+xml`.

raphlinus:
It's a good question. It's certainly possible to generate a PNG without actually doing compression, so if we really didn't want to depend on any other crates, it's possible. But I think perhaps the best thing to do here is pull in the png crate - that crate alone is not a super heavy dependency.

It's starting to seem like there might need to be a more general mechanism for resources that are not generated through factory methods on the render context. I can see that coming into play, for example, for images that are generated gpu-side, for the other backends. I'll continue to think about this.

Ralith:
Added (very) brief documentation and renamed the example.

cmyr:
What's the status of this?

Ralith:
I think this is ready to go.

piet-svg/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "piet-svg"
+version = "0.1.0"
```
cmyr: I struggle with what to do with version numbers for these crates. Eventually we'll all go to 0.1.0 together, but maybe just for the sake of pragmatism we can start this at `0.0.7`, just to make my search & replace job easier? If that offends your sensibilities then I'm happy with `0.0.1`, but I want to save `0.1.0`.

Ralith: Done.

Ralith:
> Just out of curiosity, what was the original motivation for this?

My current use of druid is a 2D vector graphics toy which I want to support both interactive editing and export. Using the same drawing code for both is extremely attractive, for obvious reasons.

spiral

PR #113, opened 2019-12-05
Add draw_image_area to RenderContext.

xStrom:
This PR expands the `draw_image` API call by adding a `src_rect` parameter. The drawing works as before, but now it uses the area of the image that is specified.

This is motivated by an image gallery app I'm writing using druid, where I need to show different areas of known images for every animation frame.

Previously this was possible by either having different images in memory or using `clip`. The new API makes it easier and probably more efficient.

Keeping ergonomics in mind for people who just want to draw the whole image (or upgrade to this new API) I also wanted it to be easy to do that. I tried a few ways of doing this.

First idea was to use `Option<impl Into<Rect>>` but unfortunately this ended up looking less than ideal.

```rust
Some(Rect::new(10.0, 10.0, 20.0, 20.0)) // Subimage
None as Option<Rect> // Whole image, the compiler demands the type info for None
```

Thus I ended up using just `impl Into<Rect>` with a special case. Less typing and better errors.

```rust
Rect::new(10.0, 10.0, 20.0, 20.0) // Subimage
Rect::ZERO // Whole image, can also be any rect for which .area() == 0
```

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -299,30 +299,63 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         Ok(image)
     }
 
+    #[inline]
     fn draw_image(
         &mut self,
         image: &Self::Image,
-        rect: impl Into<Rect>,
+        dst_rect: impl Into<Rect>,
         interp: InterpolationMode,
     ) {
-        let _ = self.with_save(|rc| {
-            let surface_pattern = SurfacePattern::create(image);
-            let filter = match interp {
-                InterpolationMode::NearestNeighbor => Filter::Nearest,
-                InterpolationMode::Bilinear => Filter::Bilinear,
-            };
-            surface_pattern.set_filter(filter);
-            let rect = rect.into();
-            rc.ctx.translate(rect.x0, rect.y0);
-            rc.ctx.scale(
-                rect.width() / (image.get_width() as f64),
-                rect.height() / (image.get_height() as f64),
-            );
-            rc.ctx.set_source(&surface_pattern);
-            rc.ctx.paint();
-            Ok(())
-        });
+        draw_image(self, image, None, dst_rect.into(), interp);
     }
+
+    #[inline]
+    fn draw_image_area(
+        &mut self,
+        image: &Self::Image,
+        src_rect: impl Into<Rect>,
+        dst_rect: impl Into<Rect>,
+        interp: InterpolationMode,
+    ) {
+        draw_image(self, image, Some(src_rect.into()), dst_rect.into(), interp);
+    }
+}
+
+fn draw_image<'a>(
+    ctx: &mut CairoRenderContext<'a>,
+    image: &<CairoRenderContext<'a> as RenderContext>::Image,
+    src_rect: Option<Rect>,
+    dst_rect: Rect,
+    interp: InterpolationMode,
+) {
+    let _ = ctx.with_save(|rc| {
+        let surface_pattern = SurfacePattern::create(image);
+        let filter = match interp {
+            InterpolationMode::NearestNeighbor => Filter::Nearest,
+            InterpolationMode::Bilinear => Filter::Bilinear,
+        };
+        surface_pattern.set_filter(filter);
+        let src_rect = match src_rect {
+            Some(src_rect) => src_rect,
+            None => Rect::new(
+                0.0,
+                0.0,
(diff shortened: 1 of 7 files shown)


cmyr:
Thanks for this, will try to do a batch of reviews over the weekend!

piet-cairo/src/lib.rs
```diff
@@ -329,12 +330,28 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
...
             };
             surface_pattern.set_filter(filter);
-            let rect = rect.into();
-            rc.ctx.translate(rect.x0, rect.y0);
-            rc.ctx.scale(
-                rect.width() / (image.get_width() as f64),
-                rect.height() / (image.get_height() as f64),
+            let src_rect = {
+                let src_rect = src_rect.into();
+                if src_rect.area() == 0.0 {
```
raphlinus: I'm not sure I love this. If "entire image" is desired, I think I'd prefer an Option. I'm also aware that this doesn't interact seamlessly with the `Into` magic.

cmyr: I think we could just make `src_rect` be `impl Into<Option<Rect>>`? It wouldn't work with all `Into<Rect>` types I don't think (I don't think there's a blanket `impl<T, S> Into<Option<S>> for T where T: Into<S>`?) but it might be easier than this.

cmyr: my bigger question, though, is whether or not this should be a new API, and `draw_image` should be the easy mode that just draws an image.

raphlinus: > my bigger question, though, is whether or not this should be a new API, and `draw_image` should be the easy mode that just draws an image.

That would be another choice, would avoid breakage, and would parallel our approach to fill mode. Obviously a reasonable implementation strategy is to have two thin wrappers that do the `Into` and call into a common function that has `Option<Rect>`.

Also, to address a comment above, `Into<Option<Rect>>` doesn't work the same way as `Into<Rect>`. It will take a `Rect` and a `None` value, but not an arbitrary type that has `Into<Rect>`. This feels like a sharp edge to me.

cmyr: I agree that it's a drawback, but my hunch is that in the wild most of the actual `impl Into<Rect>` arguments will be actual `Rect`s, and so my subjective hunch is that the ergonomic advantage of accepting either `Rect` or `None` is better than having `Rect::ZERO` be an alias for `Rect::EVERYTHING`.

Zarenor: I'd favor the two-thin-wrappers approach best, I think. Though I can imagine wanting a 'fast path' function that always draws the entire image, in addition to the sometimes-all (`Into<Option<Rect>>`) and always-done-the-math (`Into<Rect>`). I don't think the `Into<Rect>` papercut is *too* bad, but it does feel like the kind of edge case that would feel thoughtless to a new user. (One of those "Why would they do it that way?" moments)

cmyr: Okay, I think making this separate API is probably the best option.

xStrom:
The subimage support is now available through a new method called `draw_image_area` and it takes the source rect as `impl Into<Option<Rect>>` as discussed.

raphlinus:
Shouldn't that be `impl Into<Rect>` if it's drawing a subarea? (Sorry, haven't dug into the PR, just that stood out to me)

xStrom:
It used to be `impl Into<Rect>` (as seen in the original post) but I changed it based on the feedback in this PR.

I have no strong opinion here, because I can just pass a `Rect` in both cases.

There's a small performance difference, in that currently `draw_image` is a simple inlined shim, however in the `impl Into<Rect>` case (with no special case) the `draw_image` function would have to construct a proper `Rect` and so the inlining wouldn't be as obvious.

raphlinus:
Right, I think I see the confusion. What I was suggesting is that internally you have a function that takes an `Option<Rect>`. The public API has `draw_image` which takes nothing, calling the internal function with `None`, and `draw_image_area` which takes `impl Into<Rect>` and calls with `Some(into)`. My feeling is that this is the friendliest API, and avoids efficiency problems. In fact, having the `into` in a separate function makes for less monomorphization.

Sorry for not being clearer on this before.

raphlinus:
Yes, this is what I had in mind, thanks! On quick skim this looks good to me and I'm prepared to merge. Thoughts from anyone else?

PR #116, opened 2020-01-16
Document `StrokeStyle` a tad

MarcoPolo:


piet/src/shapes.rs
```diff
@@ -1,6 +1,21 @@
 //! Options for drawing paths.
 
 /// Options for drawing stroked lines.
+/// Most of these are self explanatory, but some aren't.
+///
+/// `dash` has two parts. The Vec<f64> pattern array and an offset. The array
+/// represents alternating lengths to be drawn and undrawn repeatedly. The offset
+/// specifes how far into the pattern it should start. If your platform does not
```
raphlinus: I think addressing the back-end author in the second person might be confusing, as some of the consumers of this doc are probably users of the library. I'd reword: "on platforms that do not support an odd number of lengths in the array, the implementation may concatenate two copies of the array to reach an even count."

MarcoPolo: Agreed! Thanks for the suggestion

PR #117, opened 2020-01-19
Expose current_transform() for contexts

scholtzan:
This is related to https://github.com/xi-editor/druid/issues/430 and exposes `current_transform()` and required for https://github.com/xi-editor/druid/pull/466

### DIFF
--- a/piet-direct2d/src/lib.rs
+++ b/piet-direct2d/src/lib.rs
@@ -109,11 +109,6 @@ impl<'b, 'a: 'b> D2DRenderContext<'a> {
         }
     }
 
-    pub fn current_transform(&self) -> Affine {
-        // This is an unwrap because we protect the invariant.
-        self.ctx_stack.last().unwrap().transform
-    }
-
     fn pop_state(&mut self) {
         // This is an unwrap because we protect the invariant.
         let old_state = self.ctx_stack.pop().unwrap();
@@ -426,7 +421,8 @@ impl<'a> RenderContext for D2DRenderContext<'a> {
     }
 
     fn current_transform(&self) -> Affine {
-        self.current_transform()
+        // This is an unwrap because we protect the invariant.
+        self.ctx_stack.last().unwrap().transform
     }
 
     fn make_image(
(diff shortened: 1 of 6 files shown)


piet-web/src/lib.rs
```diff
@@ -262,6 +262,13 @@ impl<'a> RenderContext for WebRenderContext<'a> {
         let _ = self.ctx.transform(a[0], a[1], a[2], a[3], a[4], a[5]);
     }
 
+    fn current_transform(&self) -> Affine {
+        // todo
+        // current_transform() and get_transform() currently not implemented:
+        // https://github.com/rustwasm/wasm-bindgen/blob/f8354b3a88de013845a304ea77d8b9b9286a0d7b/crates/web-sys/webidls/enabled/CanvasRenderingContext2D.webidl#L136
+        Affine::default()
```
scholtzan: I did open a PR to get this added: https://github.com/rustwasm/wasm-bindgen/pull/1966
Not sure what the best way here is, wait until it's merged or return something like `Affine::default()` for now and change it later?

cmyr: yea this looks good. I would also open an issue when this is merged.

piet-direct2d/src/lib.rs
```diff
@@ -425,6 +425,10 @@ impl<'a> RenderContext for D2DRenderContext<'a> {
             .set_transform(&affine_to_matrix3x2f(self.current_transform()));
     }
 
+    fn current_transform(&self) -> Affine {
+        self.current_transform()
```
cmyr: can we just move the other function down here, or does it need to be used in contexts where we don't have access to the trait?

cmyr:
update: just merged #112, which means there's another backend we will want to add this to.

scholtzan:
This should be ready to get merged now

piet-web/src/lib.rs
scholtzan: https://github.com/rustwasm/wasm-bindgen/pull/1966#issuecomment-577242039 got merged, now we'll just have to wait until a new version gets released to use it

PR #118, opened 2020-01-21
Update cairo-rs dependency

TheNeikos:


TheNeikos:
The errors seem to be unrelated to the changes, as they complain about a not installed tool chain. Is there anything I should do in that case?

cmyr:
Thanks, I'll look into the CI failures. In the meantime, while you're in here, maybe it's worth updating any other older dependencies as well? In particular if I run `cargo outdated` I notice that we're using an older version of the `image` crate, and this could get bumped now too?

cmyr:
Okay CI should be fixed, if you rebase this and push again it should pass?

cmyr:
In the process of doing piet reviews I realized I need to prepare for the next release, and it was easier to roll these changes into that patch, #120. Thanks for this PR though, much appreciated!

TheNeikos:
Awesome, thanks!

PR #131, opened 2020-02-02
[d2d] Fix crash on unsupported image format

raphlinus:
Do cleanup in `BitmapTarget::into_raw_pixels` on error.


cmyr:
This is a serious API eyebrow-raiser, though: we're passing in an option where only one of the possible values is legal? Maybe we need per-platform `Default` impls or something.

raphlinus:
I think it's basically a TODO to support the other options, waiting for use cases to drive it.

Issue #128, opened 2020-01-31
piet_common API to create an image file?


cmyr:
Playing around with testing, and have wanted to do something that I don't see an obvious API for: basically I would like to use the `piet_common` types (`Device`, `BitmapTarget` etc) to do some drawing, but then I would like to save the result to disk. I won't mind if the format of the image differs per-backend (although I suppose this isn't ideal) but I would like to be able to easily dump the current buffer in a way that lets me go take a look at it with an image viewer. Is this possible somehow?

raphlinus:
We currently have [image](https://crates.io/crates/image) as a dev-dependency in piet. This is a very heavyweight crate (at least by default, where lots of image codecs are turned on, and there are more dependencies for those). I would be open to changing that to png and moving the save function from the examples to the main library, as that seems like generally very useful functionality and a light enough dependency we can tolerate it.

Dmitry-Borodin:
working on it

Dmitry-Borodin:
Changed "image" crate to "png" crate dependency, now instead of 1.5-2Mb, we have ~200kb dependency according to lib.rs
Should I now implement saving to file functionality as public function of BitmapTarget for cairo backend?

PR #132, opened 2020-02-05
API to create an image file

Dmitry-Borodin:
Changed "image" crate to "png" crate dependency, now instead of 1.5-2Mb, we have ~200kb dependency according to lib.rs
Added this dependency for cairo-backend, actually can be implemented for direct2d as well (copy&paste the code?)
#128


piet-common/src/cairo_back.rs
```diff
@@ -121,4 +126,19 @@ impl<'a> BitmapTarget<'a> {
         }
         Ok(raw_data)
     }
+
```
cmyr: We try to have documentation for all public items

piet-common/src/cairo_back.rs
```diff
@@ -121,4 +126,19 @@ impl<'a> BitmapTarget<'a> {
...
+    pub fn save_to_file(self, path: &str) -> Result<(), piet::Error> {
```
cmyr: for `path`, I think the best type for the signature would be `impl AsRef<Path>`; this allows a larger number of acceptable types to be used here, and is the same type that `File::create` takes.

piet-common/src/cairo_back.rs
```diff
@@ -121,4 +126,19 @@ impl<'a> BitmapTarget<'a> {
...
+        let height = self.surface.get_height();
+        let width = self.surface.get_width();
+        let image = self.into_raw_pixels(ImageFormat::RgbaPremul)?;
+        let file = BufWriter::new(File::create(path)
+            .map_err(|e| Into::<Box<dyn std::error::Error>>::into(e))?);
```
cmyr: I was playing around trying to figure out if there was a less verbose way to do this conversion, and the best I can come up with is,

```rust
.map_err(|e| Box::new(e) as Box<_>)?;
// also works:
.map_err(|e| Into::<Box<_>>::into(e))?;
```
In any case this is fine, I was just playing around.

Dmitry-Borodin: That truly looks better

Dmitry-Borodin:
Thank you for review. FIxed comments, implemented the same functionality for d2d.
And I don't understand why CI is failing.

cmyr:
might just be `rustfmt` failing?

Dmitry-Borodin:
I see a problem, need to use different implementation for windows, will fix it soon.

cmyr:
cool, this looks okay; I'm not going to be able to test out the windows version for a couple days, but  it looks good!

cmyr:
@Dmitry-Borodin if you're interested in a project that takes advantage of this patch, I've written up some thoughts in https://github.com/xi-editor/druid/issues/527. If you end up experimenting with this, and run into a lifetime issue with the `Text` struct, change the `'static` to `'a`. (sorry if this is cryptic, there's just another recent breaking change in piet that you might encounter)

Dmitry-Borodin:
Thank you for information, will play around with it.

cmyr:
Okay, this is working great for me. Would like to get a +1 from @raphlinus to make sure the dependency change is okay, but I'm very happy.

cmyr:
Okay, one final request here which is that we make `png` a feature. This will mean we will also need to put this new method behind a feature flag. A nice thing we could do, here, is something like:

```rust
/// Save bitmap to RGBA PNG file
#[cfg(feature = "svg")]
pub fn save_to_file<P: AsRef<Path>>(self, path: P) -> Result<(), piet::Error> { ... }

#[cfg(not(feature = "svg"))]
pub fn save_to_file<P: AsRef<Path>>(self, path: P) -> Result<(), piet::Error> { 
    Err(piet::error::new_error(piet::error::ErrorKind::MissingFeature))
}
```

That is, the api would still exist if the feature was disabled, just we would return an error that would communicate the problem.

hwchen:
I've just added github actions to the CI in master (#138), I'm not exactly sure how it will affect this PR.

The new CI has clippy lints turned on, so I think you'll have to rebase. (I think I've caught and cleaned up some of those lints since this branch was started).

Let me know if you have any questions.

Dmitry-Borodin:
Will continue working on it to avoid warnings and clarify how features are working

Dmitry-Borodin:
Sample doesn't work properly on my machine, investigating. Like feature is always missing.

Dmitry-Borodin:
@cmyr implemented and tested on GNU/Linux. It's done now.
CI have some package download problems.

hwchen:
Re: CI. When on ubuntu, `sudo apt update` has the error:
```
Err:28 https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod bionic/main amd64 Packages
  File has unexpected size (94576 != 93944). Mirror sync in progress? [IP: 40.76.35.62 443]
  Hashes of expected file:
   - Filesize:93944 [weak]
```

@Dmitry-Borodin was the CI working before? Judging from the commit history in this PR, it appears it was working before the last commit.

Perhaps this will just fix itself if we wait...

Dmitry-Borodin:
Yes, it was working before and it's ok now. Somebody restarted checks?

PR #134, opened 2020-02-10
Text/multiline

hwchen:
Replaces #103 (which used dwrite and direct2d crates instead of raw_d2d).

Line breaking algorithm is naive. Iterates down text, checking width of each possible break.

- implement new methods for multiline in `TextLayout'
  - update_width
  - line_text
  - line_metric
  - line_count
  - and fix d2d width to include trailing whitespace

- update hit testing
  - account for multiline
  - if out of bounds top or bottom in y axis, will hit test closest line to boundary
  - all hit testing includes trailing whitespace

- lots of tests
  - basic multiline and hit testing for d2d and cairo and web
  - new test harness (using wasm-pack) for web


piet/src/text.rs
```diff
@@ -94,6 +135,26 @@ pub trait TextLayout {
     fn hit_test_text_position(&self, text_position: usize) -> Option<HitTestTextPosition>;
 }
 
+/// Metadata about each line in a text layout.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct LineMetric {
+    /// Index (in code units) of the start of the line, offset from the beginning of the text.
+    pub line_start_offset: usize,
```
cmyr: just `start_offset` is fine, we know from the type that we're dealing with a line? (ditto below)

piet/src/text.rs
```diff
@@ -94,6 +135,26 @@ pub trait TextLayout {
...
+
+    /// Line length (in code units), but offset from the beginning of the text. So it's the length
+    /// of this line summed with the lengths of all the lines before it.
+    ///
+    /// Does not include trailing whitespace.
+    pub line_end_offset: usize,
+
+    /// Line width.
```
cmyr: What are the units? I would expect this to be in pixels or something but it's not clear.

I'm also unsure about this name, 'cumulative' implies that this includes the widths of preceding lines or something? If that's the case we  should document it, but otherwise this can just be `width`?

piet/src/text.rs
```diff
@@ -94,6 +135,26 @@ pub trait TextLayout {
...
+    /// Does not include trailing whitespace.
+    pub cumulative_width: f64,
+
+    /// Length (in code units) of trailing whitespace.
```
cmyr: if this is code units it should be an integer type?

piet/src/text.rs
```diff
@@ -41,10 +67,25 @@ pub trait TextLayoutBuilder {
...
+    /// struct with recalculated lines and line metrics.
+    // TODO: Should this take &self or self? I'd say maybe; perhaps the old struct should be kept around
+    // to help with caching.
+    fn update_width(&self, new_width: f64) -> Self;
+
+    /// Given a line number, return a reference to that line's underlying string.
+    fn line_text(&self, line_number: usize) -> Option<&str>;
+
+    /// Given a line number, return a reference to that line's metrics.
+    fn line_metric(&self, line_number: usize) -> Option<&LineMetric>;
```
cmyr: if `LineMetric` doesn't contain anything expensive, this should return it by value, not by reference.

piet/src/text.rs
```diff
@@ -94,6 +135,26 @@ pub trait TextLayout {
...
+    /// Line length (in code units), but offset from the beginning of the text. So it's the length
+    /// of this line summed with the lengths of all the lines before it.
+    ///
+    /// Does not include trailing whitespace.
+    pub line_end_offset: usize,
+
+    /// Line width.
+    /// Does not include trailing whitespace.
+    pub cumulative_width: f64,
+
```
cmyr: we're also going to need a way to get the *height* of the line, as well as the position of the baseline relative to the height.

Ideally we would get both the nominative line height (that is, the value by which lines should be spaced) as well as the *actual* height of the line as rendered, by which I mean the height of the rect that encloses all of the glyphs in the line; these can be different in certain cases.

raphlinus: This LineMetrics struct is designed to be an platform-neutral abstraction of [DWRITE_LINE_METRICS ](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/ns-dwrite-dwrite_line_metrics), which has height and baseline fields. These are enough to do basic formatting and alignment. I suggest we add them.

In addition to that, I'm proposing that we store the cumulative sum of these. That makes binary searching efficient, and it's also efficient to get metrics for one line by subtracting two adjacent lines (we can have a query function that does this as a convenience).

cmyr: I'm not sure how well that translates to CoreText, but I haven't looked too deeply.

I'm also not sure about the cumulative sum stuff, which feels like an implementation detail, not necessarily part of the api? But I'm not too concerned about the details, I just want the basics.

piet/src/text.rs
hwchen: Hmmm... this was in the last PR for line breaking, so I carried it over. But looking at it again I'm not sure exactly why we need cumulative width. Is it better to just use width?

piet/src/text.rs
hwchen: I made some mixup here; I think trailing whitespace should be in code units (usize)

raphlinus:
I think it makes sense not to do cumulative width of lines, but it does make sense to do cumulative code unit counts. This is not a tricky thing, a "cumulative code unit count" is just an offset.

piet/src/text.rs
cmyr: I think we want just "width", in pixels. Also I think this comment was maybe in response to this?


> Raph: I think it makes sense not to do cumulative width of lines, but it does make sense to do cumulative code unit counts. This is not a tricky thing, a "cumulative code unit count" is just an offset.




https://github.com/linebender/piet/pull/134#issuecomment-595604575

hwchen:
If `LineMetric` looks good, and the new `TextLayout` methods (update_width, line_text, line_metric, line_count) look good, I'll implement the windows backend next.

piet/src/text.rs
```diff
@@ -94,6 +135,34 @@ pub trait TextLayout {
...
+    /// Does not include trailing whitespace.
+    pub end_offset: usize,
+
+    /// Length in (in code units) of current line's trailing whitespace.
+    pub trailing_whitespace: usize,
+
+    /// Line width in pixels
+    pub width: f64,
+
+    /// Distance (in pixels) of the baseline of the line from the top of the line
```
cmyr:
```suggestion
    /// Distance (in pixels) of the baseline from the top of the line
```

cmyr:
This looks reasonable to me.

hwchen:
@cmyr, it looks like dwrite line metrics do not provide width; it's done at the text layout level. Does it make sense to remove width from `LineMetric`?

https://docs.rs/winapi/0.3.8/winapi/um/dwrite/struct.DWRITE_LINE_METRICS.html

cmyr:
hm, I'm at least a bit surprised. I'd defer to @raphlinus, I would imagine there's *some* way to get the width for a line?

raphlinus:
> hm, I'm at least a bit surprised. I'd defer to @raphlinus, I would imagine there's _some_ way to get the width for a line?

There is some way, but it looks like you have to go a lot lower level in the DirectWrite API - I think what you do is iterate through the results of [GetClusterMetrics](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritetextlayout-getclustermetrics) and gather together all the clusters on a line. My *sense* is that DWrite provides high-level access where you're just painting paragraphs, and then for the things that require more fine grained querying, they provide cluster granularity.

Based on that, I think we'll remove it for now, and come back to this when we get a clearer picture of what we really need.

hwchen:
@cmyr do you anticipate needing line metrics (or string slices) for a range of lines? The api is currently for just one line at a time.

cmyr:
I'm totally happy if we just have line metrics first.

hwchen:
Thanks for the quick turnaround. I'll get this ported to piet-web asap.

hwchen:
Woohoo! I'm ready for the bug reports :)

PR #136, opened 2020-02-10
Clippy cleanup

hwchen:
- [x] Fix all clippy lints in piet

Includes a change from `&self` to `self` for `render_context::ImageFormat::bytes_per_pixel`, referenced at https://github.com/linebender/piet/pull/133#discussion_r377208811

This should have no effect on API


hwchen:
As discussed with @raphlinus, I've split the clippy stuff here and will work on the github actions in a separate review.

PR #138, opened 2020-02-11
Github actions for CI

hwchen:


hwchen:
@cmyr and @raphlinus I've modified `.github/workflows/ci.yml` from druid to work for piet. I believe rustfmt, tests, clippy and doc building are all working (with logic for platform-specific excludes). A scan of the logs appears to look good.

I believe that turning off the azure pipelines integration is a blocker. I've removed the azure configs, but the integration appears to still run.

cmyr:
@hwchen can we keep the azure configs, and we'll just run them both for a week or so?

hwchen:
Sounds good, will do

hwchen:
Should be ready to go?

PR #141, opened 2020-03-12
Restore common types for Direct2D.

xStrom:
The common types `Brush` and `Image` got removed from the Direct2D backend with #111. They weren't removed from other backends nor did the removal have a deprecation period. Thus I imagine it was a mistake.

This PR adds the types back. It also adds a test to prevent accidental removal in the future.

### DIFF
--- a/piet-common/src/direct2d_back.rs
+++ b/piet-common/src/direct2d_back.rs
@@ -11,6 +11,7 @@ use std::path::Path;
 use png::{ColorType, Encoder};
 
 use piet::{ErrorKind, ImageFormat};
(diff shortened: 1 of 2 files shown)


raphlinus:
This seems reasonable to me. @hwchen can you double-check?

Surprising this didn't break stuff.

hwchen:
Thanks @xStrom. I think I had some other rationale for removing (if it was important enough I should have noted it though), but consistency is important, and thanks for the test.

raphlinus:
I think it might be a good idea as cleanup to delete all these, and force code to access the types as associated types of `RenderContext`, which would have the benefit of making code that uses them generic. But this is a low priority.

PR #146, opened 2020-03-15
direct2d: Add swapchain support and a winit example

msiglreith:
This adds basic support for creating a swapchain froma HWND handle including a sample using winit.
I'm not sure if surface/swapchain handling is in scope of `piet` but I feel it's reasonable to have as extension to make it easier to get started.

### DIFF
--- a/piet-common/src/direct2d_back.rs
+++ b/piet-common/src/direct2d_back.rs
@@ -123,11 +123,11 @@ impl Device {
         // Bind the backing texture to a D2D Bitmap
         let target = unsafe {
             context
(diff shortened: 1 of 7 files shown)


raphlinus:
This one is certainly trickier; in our stack, swapchains are firmly in the domain of druid-shell rather than piet. But these are not hard and fast boundaries, and it might be helpful to potential non-druid users to have such a thing.

cmyr:
what's the status here?

raphlinus:
Thanks for the exploration, and sorry not to have been more responsive to it. I have a strong feeling we'll come back to this when piet-gpu becomes more real (my prototyping is already winit based).

PR #153, opened 2020-03-25
Remove refs from WebRenderContext constructor API

elrnv:
Specifically, the lifetime of `WebRenderContext` is changed to not depend on lifetimes of  `CanvasRenderingContext2d` and `Window` objects, since these lifetimes don't actually enforce the existence of web resources anyways.

The `mut `is also removed from `CanvasRenderingContext2d` since its API has no `mut` calls and this generally makes the `WebRenderContext` difficult to use in libraries like druid.

For compatibility the `WebRenderContext` maintains a phantom lifetime placeholder for compatibility as well as for anticipating a correct lifetime there.

### DIFF
--- a/piet-web/examples/basic/src/lib.rs
+++ b/piet-web/examples/basic/src/lib.rs
@@ -19,7 +19,7 @@ pub fn run() {
         .unwrap()
         .dyn_into::<HtmlCanvasElement>()
         .unwrap();
-    let mut context = canvas
+    let context = canvas
         .get_context("2d")
         .unwrap()
         .unwrap()
@@ -31,7 +31,7 @@ pub fn run() {
     canvas.set_height((canvas.offset_height() as f64 * dpr) as u32);
     let _ = context.scale(dpr, dpr);
 
-    let mut piet_context = WebRenderContext::new(&mut context, &window);
+    let mut piet_context = WebRenderContext::new(context, window);
     run_tests(&mut piet_context);
 
     // TODO: make the test picture selectable
--- a/piet-web/src/lib.rs
+++ b/piet-web/src/lib.rs
@@ -26,18 +26,20 @@ use piet::{
 pub use text::{WebFont, WebFontBuilder, WebTextLayout, WebTextLayoutBuilder};
 
 pub struct WebRenderContext<'a> {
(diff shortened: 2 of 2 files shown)


cmyr:
I'm really not sure who 'owns' this code, I've never looked at the piet-web stuff. Are you using this anywhere @elrnv?

elrnv:
I was looking for a UI library to do some prototyping preferably with web support, since it makes it easier to share the results with others. I saw that druid doesn't yet have complete support for wasm so I thought I could contribute there. This PR makes it possible to integrate piet with wasm in druid ([#759](https://github.com/xi-editor/druid/pull/759)). Hope it helps :)

cmyr:
So I think the motivation for having these signatures might be `piet-common`. At the moment I certainly can't get the tests to compile (I can't actually compile on mac period but that's a separate problem, i'll open an issue for that.)


So I guess: can you run tests in piet-common? I think it doesn't _quite_ matter how closely these APIs match, as long as they are used correctly from the particular backends; it does matter when we want to be able to treat the backends interchangeably.

xStrom:
The `piet-common` tests do seem to compile for me with these changes. I also opened #156 so that our CI here would compile tests with wasm32.

elrnv:
Can confirm the compile test in `piet-common` works for me too. I don't think I ended up changing anything in `piet-common`, just `piet-web`. So this PR should only effect `piet-web` AFAICT.

If I remember correctly (from reading through the code and issues), the goal of having a lifetime on the `Piet` context (and `PietText` I guess) is to ensure that the context doesn't outlive the "factory" that it relies on. But in this case the lifetime refers only to the `web-sys` object pointer, which may in theory outlive the resource (window/canvas/context) it points to.

I'm not sure how to tie the lifetime to the availability of the context in wasm. Getting the window initially is done through `web_sys::window()` function which returns on `Option` so the lifetime is effectively checked at runtime. We could potentially restructure the code to call `web_sys::window()` every time we need the window (followed by getting the canvas and context if those are needed), but this is awkward IMO.

I'm not sure how to test the situation where the window/canvas/context would not be available. I tried deleting the canvas from the inspector, but it did not cause any errors, as if the underlying canvas/context were not garbage collected.

Issue #115, opened 2019-12-09
A bunch of Direct2D text tests fail on Windows 7.


xStrom:
## My environment
**Windows 7 SP1** (build 7601) x64
rustc **1.39.0** (4560ea788 2019-11-04)
nVidia GTX 1060 (**397.93**)

The tests fail with both `stable-x86_64-pc-windows-msvc` and `stable-x86_64-pc-windows-gnu`.

## These two tests fail in `piet-direct2d`

```
test test::test_hit_test_point_complex ... FAILED
test test::test_hit_test_text_position_complex_1 ... FAILED

---- test::test_hit_test_point_complex stdout ----
text pos 2: Some(HitTestTextPosition { point: (6.275390625, 0.0), metrics: HitTestMetrics { text_position: 2 } })
text pos 9: Some(HitTestTextPosition { point: (13.365234375, 0.0), metrics: HitTestMetrics { text_position: 9 } })
text pos 10: Some(HitTestTextPosition { point: (19.833984375, 1.546875), metrics: HitTestMetrics { text_position: 10 } })
text pos 14: Some(HitTestTextPosition { point: (28.669921875, 0.0), metrics: HitTestMetrics { text_position: 14 } })
thread 'test::test_hit_test_point_complex' panicked at 'assertion failed: `(left == right)`
  left: `6`,
 right: `2`', piet-direct2d\src\lib.rs:1001:9

---- test::test_hit_test_text_position_complex_1 stdout ----
thread 'test::test_hit_test_text_position_complex_1' panicked at 'assertion failed: x <= max && x >= min', piet-direct2d\src\lib.rs:707:9
```

#### test_hit_test_point_basic_0
```
piet_web_example.js:506 text pos 4: Some(HitTestTextPosition { point: (17.994140625, 0.0), metrics: HitTestMetrics { text_position: 4 } })
piet_web_example.js:506 text pos 5: Some(HitTestTextPosition { point: (20.994140625, 0.0), metrics: HitTestMetrics { text_position: 5 } })
RuntimeError: unreachable
    at __rust_start_panic (wasm-function[396]:0x1ab88)
    at rust_panic (wasm-function[254]:0x1a1a6)
    at std::panicking::rust_panic_with_hook::h5e7c2dc110ae79d4 (wasm-function[142]:0x172f1)
    at std::panicking::continue_panic_fmt::hb5b3e4b5160fe2ab (wasm-function[175]:0x18978)
[...]

PR #154, opened 2020-03-25
Fix piet-web tests & add console hook for panics

elrnv:
This commit tweaks the tests to pass, and adds a useful console error hook to capture panics in the browser console.

Admittedly, the tweak in tests is perhaps dependent on my system configuration, so it will be worth for reviewers to verify whether the tests pass. To do so please check the console log in the browser.

### DIFF
--- a/piet-web/examples/basic/Cargo.toml
+++ b/piet-web/examples/basic/Cargo.toml
@@ -16,7 +16,7 @@ piet-web = { path = "../.." }
 piet-test = { path = "../../../piet-test" }
 
 wasm-bindgen = "0.2.30"
-console_error_panic_hook = { version = "0.1.1", optional = true }
+console_error_panic_hook = { version = "0.1.6", optional = true }
 
 [dependencies.web-sys]
 version = "0.3.10"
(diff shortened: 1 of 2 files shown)


xStrom:
The text tests are broken in a more deep way, as you guessed it. I created #115 previously to point out how they're even broken differently depending on the browser you use. I can also confirm that your test fixes don't work on my machine. (Neither do the unchanged ones.)

The console hook looks great though! If you would update your PR here to only contain the console hook then we can get that merged.

For the failing tests we can discuss it some more in separate issue, perhaps in #115.

piet-web/examples/basic/Cargo.toml
```diff
@@ -7,12 +7,17 @@ edition = "2018"
...
+[features]
+default = ["console_error_panic_hook"]
+
 [dependencies]
 piet = { path = "../../../piet" }
 piet-web = { path = "../.." }
 piet-test = { path = "../../../piet-test" }
 
 wasm-bindgen = "0.2.30"
+console_error_panic_hook = { version = "0.1.1", optional = true }
```
xStrom: Sorry I didn't notice this earlier, but is there a specific reason this is `0.1.1`? `0.1.6` seems to be the latest.

elrnv: No reason, just copied from the [rust game of life tutorial](https://github.com/rustwasm/wasm_game_of_life/blob/master/Cargo.toml)

PR #155, opened 2020-03-26
Fix backend choice collision

cmyr:
Previously it was possible to select both cairo and the web
backend by passing --features=web without having the target be
wasm32.

This attempts to simplfy how we choose the backend.

We will use cairo if it is selected explicitly, or if we're not on
wasm32, windows, or don't see the 'web' feature;

We will use d2d if we're on windows and don't see the 'cairo' feature;

And we will use the web backend if our target arch is wasm32 or
we see the "web" feature.

We will fail to compile if we cannot select a backend.

### DIFF
--- a/piet-common/src/lib.rs
+++ b/piet-common/src/lib.rs
@@ -29,20 +29,21 @@ pub use piet::*;
 #[doc(hidden)]
 pub use piet::kurbo;
 
-#[cfg(any(
-    feature = "cairo",
-    not(any(target_arch = "wasm32", target_os = "windows", feature = "direct2d"))
-))]
-#[path = "cairo_back.rs"]
-mod backend;
-
-#[cfg(any(feature = "d2d", all(target_os = "windows", not(feature = "cairo"))))]
-#[path = "direct2d_back.rs"]
-mod backend;
-
-#[cfg(any(feature = "web", target_arch = "wasm32"))]
-#[path = "web_back.rs"]
-mod backend;
+cfg_if::cfg_if! {
+    // if we have explicitly asked for cairo *or* we are not wasm, web, or windows:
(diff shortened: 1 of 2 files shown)


piet-common/src/lib.rs
```diff
@@ -29,20 +29,21 @@ pub use piet::*;
...
+    // if we have explicitly asked for cairo *or* we are not wasm, web, or windows:
+    if #[cfg(any(feature = "cairo", not(any(target_arch = "wasm32", feature="web", target_os = "windows"))))] {
+        #[path = "cairo_back.rs"]
+         mod backend;
+    } else if #[cfg(target_os = "windows")] {
+        #[path = "direct2d_back.rs"]
+        mod backend;
+    } else if #[cfg(any(feature = "web", target_arch = "wasm32"))] {
+    #[path = "web_back.rs"]
+    mod backend;
```
xStrom: Missing some formating spaces here.

piet-common/src/lib.rs
```diff
@@ -29,20 +29,21 @@ pub use piet::*;
...
-mod backend;
-
-#[cfg(any(feature = "web", target_arch = "wasm32"))]
-#[path = "web_back.rs"]
-mod backend;
+cfg_if::cfg_if! {
+    // if we have explicitly asked for cairo *or* we are not wasm, web, or windows:
+    if #[cfg(any(feature = "cairo", not(any(target_arch = "wasm32", feature="web", target_os = "windows"))))] {
+        #[path = "cairo_back.rs"]
+         mod backend;
```
xStrom: Whoops, didn't notice that extra formating space earlier. Fine to keep it I guess, but I just wanted to let the info be out there. Really unfortunate that rustfmt doesn't work on this.

cmyr: 😆

PR #158, opened 2020-03-27
Switch to offset_{width,height} when clearing web context

elrnv:
This enables proper browser zooming through the dpi/devicePixelRatio mechanism.

For reference this addresses the zooming issue in druid wasm support found in the [#759](https://github.com/xi-editor/druid/pull/759).

### DIFF
--- a/piet-web/src/lib.rs
+++ b/piet-web/src/lib.rs
@@ -113,7 +113,7 @@ impl<'a> RenderContext for WebRenderContext<'a> {
 
     fn clear(&mut self, color: Color) {
         let (width, height) = match self.ctx.canvas() {
-            Some(canvas) => (canvas.width(), canvas.height()),
+            Some(canvas) => (canvas.offset_width(), canvas.offset_height()),
             None => return,
             /* Canvas might be null if the dom node is not in
              * the document; do nothing. */


elrnv:
Sorry, I should have probably elaborated more. Maybe I can motivate this change for future reference:

-  Zooming in and out in the browser changes the reported `devicePixelRatio` (zooming in increases it, zooming out decreases it). This ratio is also used to handle screens with different DPI settings (e.g. retina displays have higher pixel ratio).
- When zoom is changed, `width` & `height` on the canvas do not change (because of how we are setting up the context from druid, which follows [the example from MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio)). OTOH `offset_width` and `offset_height` decrease when zoomed in, and increase when zoomed out (these are read only properties so their behaviour should be consistent).
- Zooming is achieved, by drawing to a smaller (when zoomed in) or larger (when zoomed out) context (with dimensions given by `offset_{width,height}`) and scaling it to fit the canvas, which is scaled by the device pixel ratio from the reported `offset_{width,height}`. This way we can change the size of all drawn things inside the context.
- So to make zooming out work, we should expect the visible part of the context to have the size reported by `offset_width` and `offset_height`, which are guaranteed to be increased when the view is zoomed out since they are read-only.

I think this zooming mechanism is an appropriate initial solution. However, I can imagine that it may be preferable that the rendering of some widgets is tied to screen resolution, so probably this approach will be revisited in the future.

PR #159, opened 2020-04-01
Impl From for Color

futurepaul:
I find myself copy pasting a "split_rgba" function in a few projects. By impling From we get Into for free and this will cover my use case. I could also add a From<[f64; 4]> / From<[u8; 4]> if those are interesting, but I haven't personally needed those yet.

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -171,43 +171,31 @@ impl Color {
         }
     }
 
+    /// Convert a color value to four 8-bit rgba values.
+    pub fn as_rgba_u8(&self) -> (u8, u8, u8, u8) {
+        let rgba = self.as_rgba_u32();
+        (
+            (rgba >> 24 & 255) as u8,
+            ((rgba >> 16) & 255) as u8,
+            ((rgba >> 8) & 255) as u8,
+            (rgba & 255) as u8,
+        )
+    }
+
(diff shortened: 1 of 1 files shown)


futurepaul:
I would be fine with just adding methods like `as_rgba_u8` and `as_rgba_f64` and skipping on `From` if that would be more future-proof?

futurepaul:
Going to. It's easy to make a color given an `r g b a`, but given a color it's hard to get the `r g b a` currently.

piet/src/color.rs
```diff
@@ -171,6 +171,28 @@ impl Color {
...
+        (
+            (rgba >> 24 & 255) as u8,
+            ((rgba >> 16) & 255) as u8,
+            ((rgba >> 8) & 255) as u8,
+            (rgba & 255) as u8,
+        )
+    }
+
+    /// Convert a color value to four f64 values, each in the range 0.0 to 1.0.
+    pub fn as_rgba_f64(rgba: &Color) -> (f64, f64, f64, f64) {
```
cmyr, line 186: our naming convention here seems to be that we assume `f64`; so this can just be `as_rgba`.
(I'll make this change, since I'd like to get this in before doing a new piet release)

```suggestion
    pub fn as_rgba(rgba: &Color) -> (f64, f64, f64, f64) {
```

PR #162, opened 2020-04-09
Add blurred rectangles

raphlinus:


piet-direct2d/src/d2d.rs
```diff
@@ -93,6 +98,11 @@ pub struct Brush(ComPtr<ID2D1Brush>);
 
 pub struct Bitmap(ComPtr<ID2D1Bitmap1>);
 
+pub struct Effect(ComPtr<ID2D1Effect>);
+
+// TODO: consider combining with piet-common direct2d_back version.
+pub struct BitmapRenderTarget(ComPtr<ID2D1BitmapRenderTarget>);
```
cmyr, line 104: what's the rationale here?

piet-cairo/src/blurred_rect.rs
```diff
@@ -0,0 +1,57 @@
...
+//! images.
+
+use cairo::{Format, ImageSurface};
+
+use piet::kurbo::{Point, Rect};
+
+/// Extent to which to expand the blur.
+const BLUR_EXTENT: f64 = 2.5;
+
+pub fn compute_blurred_rect(rect: Rect, radius: f64) -> (ImageSurface, Point) {
```
cmyr, line 16: Does this need to be public? Feels like it is currently an implementation detail of other public API?

piet-cairo/src/blurred_rect.rs
```diff
@@ -0,0 +1,57 @@
...
+    let yfrac = rect_padded.y0 - rect_exp.y0;
+    let width = rect_exp.width() as usize;
+    let height = rect_exp.height() as usize;
+    let strip = (0..width)
+        .map(|i| {
+            let x = ((i as f64) - (xfrac + padding)) * radius_recip;
+            (255.0 * 0.25) * (compute_erf7(x) + compute_erf7(xmax - x))
+        })
+        .collect::<Vec<_>>();
+    // TODO: don't panic on error
```
cmyr, line 33: Hard to know what to do here. if we return an error are we going to handle it any better? If this fails it's likely we're out of memory. I wish I had better guidelines for how to handle these sorts of errors. At a higher level, What's the best behaviour? Do we just skip this operation, but keep the process running?

raphlinus:
Made minor changes in response to review, thanks. Thumbs up to commit?

PR #166, opened 2020-04-10
[text] TextLayout width as Option<f64> instead of f64, fix update_width

hwchen:
Updates `Text::new_text_layout` and `TextLayout::update_width` to use
`impl Into<Option<f64>>` for width and new width parameters. Internally,
`f64` is still used.

Fixed a bug for piet-web and piet-d2d where `TextLayout::update_width` did
not recalculate new width. Should add tests for recalculating width.

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -76,9 +76,11 @@ impl<'a> Text for CairoText<'a> {
         &mut self,
         font: &Self::Font,
         text: &str,
-        width: f64,
+        width: impl Into<Option<f64>>,
     ) -> Self::TextLayoutBuilder {
         // TODO Should probably move the calculations to `build` step
+        let width = width.into().unwrap_or(std::f64::INFINITY);
+
         let line_metrics = lines::calculate_line_metrics(text, &font.0, width);
 
         let widths = line_metrics.iter().map(|lm| {
@@ -87,7 +89,6 @@ impl<'a> Text for CairoText<'a> {
                 .x_advance
         });
 
-        // TODO default width 0?
         let width = widths.fold(0.0, |a: f64, b| a.max(b));
 
         let text_layout = CairoTextLayout {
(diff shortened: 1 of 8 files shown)


piet-cairo/src/text.rs
```diff
@@ -129,9 +130,20 @@ impl TextLayout for CairoTextLayout {
...
 
-    fn update_width(&mut self, new_width: f64) -> Result<(), Error> {
-        self.width = new_width;
+    // TODO refactor this to use same code as new_text_layout
+    fn update_width(&mut self, new_width: impl Into<Option<f64>>) -> Result<(), Error> {
+        let new_width = new_width.into().unwrap_or(std::f64::INFINITY);
+
         self.line_metrics = lines::calculate_line_metrics(&self.text, &self.font, new_width);
+
+        let widths = self.line_metrics.iter().map(|lm| {
```
cmyr, line 139: is this fixing a separate bug?

piet/src/text.rs
```diff
@@ -78,7 +78,7 @@ pub trait TextLayout: Clone {
 
     /// Used for changing the width of a text layout. Given a width, returns a [`TextLayout`]
     /// struct with recalculated lines and line metrics.
```
cmyr, line 80:
```suggestion
    /// Change the width of this `TextLayout`.
    ///
    /// This may be an `f64`, or `None` if this layout is not constrained;
    /// `None` is equivalent to `std::f64::INFINITY`.
```

piet-cairo/src/text.rs
hwchen: Yes, I made a quick note but I can say a bit more here. In the text layout build step, I calculate the overall layout width by using the max of the widths of the lines in the layout. Unfortunately, on the first pass, I forgot that this also needed to be done when updating the width (I had a placeholder which simply set the width to the `width` argument).

I thought about making it cleaner by doing some refactoring; but decided to keep it simple for now, and perhaps refactor in another PR.

hwchen: I also thought about fixing in a separate PR, I can also do that if you prefer. I guess I just wanted to fix it as soon as I saw it :sweat_smile:

cmyr: I understand, separate PR is probably strictly 'better' but i'm happy to get this in, just wanted to make sure I wasn't overlooking something.

PR #167, opened 2020-04-12
Wasm bitmap target

elrnv:
This PR adds a placeholder for the bitmap target rendering mechanism present in other backends. In the current state this PR will enable druid unit tests to be built with WASM, although their execution will need to be thought out further.

I intend this PR to simply get the ball rolling toward getting testing working for the wasm backend in druid since https://github.com/xi-editor/druid/pull/759 doesn't provide any testing in wasm.

### DIFF
--- a/piet-common/Cargo.toml
+++ b/piet-common/Cargo.toml
@@ -36,3 +36,6 @@ png = { version = "0.16.1", optional = true }
 
 [target.'cfg(target_arch="wasm32")'.dependencies]
 piet-web = { version = "0.0.12", path = "../piet-web" }
+web-sys = { version = "0.3.36", features = ["console", "Window", "CanvasGradient", "CanvasRenderingContext2d", "CanvasWindingRule", "Document", "Element", "HtmlCanvasElement", "ImageBitmap", "ImageData", "TextMetrics"] }
+wasm-bindgen = "0.2.59"
+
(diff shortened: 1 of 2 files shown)


cmyr:
@elrnv any suggestions for how I should test that this is working as intended?

elrnv:
Yes! In druid, remove the `#[cfg(not(target_arch = "wasm32"))]` instances in `/src/window.rs`, `/src/lib.rs` and `/src/event.rs` and compile against this PR. The compile errors should be removed.

elrnv:
I should emphasize that this is just a placeholder for the implementation. I'm not sure yet how to actually test it.

cmyr:
this is maybe a bit daft, but what target tam I building? I can't seem to get compilation to fail even without this PR. 😕

elrnv:
Ah sorry, I should have been more clear. I guess the idea is that the test harness in druid fails to build without this bitmap target, so to trigger the build of the test harness in druid (for wasm) you need

```
cargo test --no-run --target wasm32-unknown-unknown
```

Of course on top of removing the `#[cfg(not(target_arch = "wasm32"))]` instances, you also need to add something like

```
[patch.crates-io]
piet-common = { git = "https://github.com/elrnv/piet.git", branch = "wasm-bitmap-target", features = ["png"] }
```

To the top level Cargo.toml to ensure that the right piet-common version is being pulled.

(BTW It seems like I had some things missing for some reason, but I think the issues should be fixed now).

PR #171, opened 2020-04-15
current_transform() for web backend

scholtzan:
Follow-up to https://github.com/linebender/piet/pull/117

The newest version of `wasm-bindgen` now exposes `get_transform()`

### DIFF
--- a/piet-web/src/lib.rs
+++ b/piet-web/src/lib.rs
@@ -12,8 +12,8 @@ use std::ops::Deref;
 use js_sys::{Float64Array, Reflect};
 use wasm_bindgen::{Clamped, JsCast, JsValue};
 use web_sys::{
-    CanvasGradient, CanvasRenderingContext2d, CanvasWindingRule, HtmlCanvasElement, ImageData,
-    Window,
+    CanvasGradient, CanvasRenderingContext2d, CanvasWindingRule, DomMatrix, HtmlCanvasElement,
+    ImageData, Window,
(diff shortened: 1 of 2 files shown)


piet-web/src/lib.rs
```diff
@@ -236,10 +236,7 @@ impl<'a> RenderContext for WebRenderContext<'a> {
     }
 
     fn current_transform(&self) -> Affine {
-        // todo
-        // current_transform() and get_transform() currently not implemented:
-        // https://github.com/rustwasm/wasm-bindgen/blob/f8354b3a88de013845a304ea77d8b9b9286a0d7b/crates/web-sys/webidls/enabled/CanvasRenderingContext2D.webidl#L136
-        Affine::default()
+        matrix_to_affine(self.ctx.get_transform().unwrap())
```
cmyr, line 239: what are the conditions under which this might fail? I might prefer to log this and return an identity affine? Not sure if anyone else has any opinions..

raphlinus: I don't have very strong feelings but am ok with the crash (it seems more likely to be noticed to me). From having looked at a couple more things like this (text measurement), failures here are extremely unlikely in browsers that can actually run wasm.

scholtzan: My slight preference is to let if fail rather than having some weird behaviour. But I'm also okay with logging some error message and return an identity affine

PR #173, opened 2020-04-16
Streamline CI testing and expand it to tests and examples

xStrom:
Previously clippy was only testing the main code.


hwchen:
Thanks, I didn’t catch this!

hwchen:
Why does `clippy` require `--all-targets` but the other steps do not?

xStrom:
Clippy seems to behave like `cargo build` [which defaults to](https://doc.rust-lang.org/cargo/commands/cargo-build.html) `--bins --lib`.
The `--all-targets` means `--lib --bins --tests --benches --examples`.

While `cargo test` [defaults to a lot more](https://doc.rust-lang.org/cargo/commands/cargo-test.html) though not everything.

`cargo rustc` seems to work as build so it's not building examples etc right now. I'm still learning all of this but I'm not even sure what building does differently from clippy, because doesn't clippy also build?

xStrom:
After looking more into this I've confirmed that it makes sense to remove the `cargo rustc` compiling we had. `cargo clippy` already does that.

Indeed the clippy readme notes:
> Note that adding `-D warnings` will cause your build to fail if any warnings are found in your code. That includes warnings found by rustc (e.g. `dead_code`, etc.). 

I also confirmed this in practice by testing it out.

Knowledge of this enables us to have a significantly faster CI testing phase.

xStrom:
The CI checks have completely stopped working, so I'll create a new PR with these same changes.

hwchen:
Interesting, even though I had ported over the CI from druid, my mental model was still that compiler warnings were denied somewhere else (maybe on `cargo test`?). Being able to remove a compiler pass is awesome, thanks!

PR #176, opened 2020-04-28
Core Graphics back-end

raphlinus:
This is still work in progress. It's based off #4, updated for API changes and so on. In its current state, paths work and there is code for images, but nothing on gradients or text yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,21 +59,28 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace --all-targets --exclude piet-cairo -- -D warnings
+          args: --workspace --all-targets --exclude piet-cairo --exclude piet-coregraphics -- -D warnings
         if: contains(matrix.os, 'windows')
 
-      - name: cargo clippy (not windows)
+      - name: cargo clippy (mac)
         uses: actions-rs/cargo@v1
         with:
           command: clippy
           args: --workspace --all-targets --exclude piet-direct2d -- -D warnings
-        if: contains(matrix.os, 'windows') != true
+        if: contains(matrix.os, 'macOS')
+
+      - name: cargo clippy (linux)
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --workspace --all-targets --exclude piet-direct2d --exclude piet-coregraphics -- -D warnings
+        if: contains(matrix.os, 'ubuntu')
 
       - name: cargo test (windows)
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --workspace --exclude piet-cairo
+          args: --workspace --exclude piet-cairo --exclude piet-coregraphics
         if: contains(matrix.os, 'windows')
 
       - name: cargo test (not windows)
@@ -81,7 +88,14 @@ jobs:
         with:
           command: test
           args: --workspace --exclude piet-direct2d
-        if: contains(matrix.os, 'windows') != true
+        if: contains(matrix.os, 'macOS')
+
+      - name: cargo test (not windows)
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --workspace --exclude piet-direct2d --exclude piet-coregraphics
+        if: contains(matrix.os, 'ubuntu')
 
   test-stable-wasm:
     runs-on: ${{ matrix.os }}
@@ -105,13 +119,13 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace --all-targets --exclude piet-cairo --exclude piet-direct2d --target wasm32-unknown-unknown -- -D warnings
+          args: --workspace --all-targets --exclude piet-cairo --exclude piet-direct2d --exclude piet-coregraphics --target wasm32-unknown-unknown -- -D warnings
 
       - name: cargo test compile
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --workspace --exclude piet-cairo --exclude piet-direct2d --no-run --target wasm32-unknown-unknown
+          args: --workspace --exclude piet-cairo --exclude piet-direct2d --exclude piet-coregraphics --no-run --target wasm32-unknown-unknown
 
   test-nightly:
     runs-on: ${{ matrix.os }}
@@ -143,15 +157,22 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --workspace --exclude piet-cairo
+          args: --workspace --exclude piet-cairo --exclude piet-coregraphics
         if: contains(matrix.os, 'windows')
 
-      - name: cargo test (not windows)
+      - name: cargo test (mac)
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --workspace --exclude piet-direct2d
-        if: contains(matrix.os, 'windows') != true
+        if: contains(matrix.os, 'macOS')
+
+      - name: cargo test (linux)
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --workspace --exclude piet-direct2d --exclude piet-coregraphics
+        if: contains(matrix.os, 'ubuntu')
 
   check-docs:
     name: Docs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,14 +83,14 @@ jobs:
           args: --workspace --exclude piet-cairo --exclude piet-coregraphics
         if: contains(matrix.os, 'windows')
 
-      - name: cargo test (not windows)
+      - name: cargo test (mac)
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --workspace --exclude piet-direct2d
         if: contains(matrix.os, 'macOS')
 
-      - name: cargo test (not windows)
+      - name: cargo test (linux)
         uses: actions-rs/cargo@v1
         with:
           command: test
--- a/piet-common/src/cg_back.rs
+++ b/piet-common/src/cg_back.rs
@@ -87,7 +87,7 @@ impl Device {
             core_graphics::base::kCGImageAlphaPremultipliedLast,
         );
         ctx.scale(pix_scale, pix_scale);
-        let height = height as f64 * pix_scale;
+        let height = height as f64 * pix_scale.recip();
         Ok(BitmapTarget {
             ctx,
             height,
@@ -112,27 +112,8 @@ impl<'a> BitmapTarget<'a> {
             return Err(Error::NotSupported);
         }
 
-        let width = self.ctx.width() as usize;
-        let height = self.ctx.height() as usize;
-        let stride = self.ctx.bytes_per_row() as usize;
-
         let data = self.ctx.data();
-        if stride != width {
-            let mut raw_data = vec![0; width * height * 4];
-            for y in 0..height {
-                let src_off = y * stride;
-                let dst_off = y * width * 4;
-                for x in 0..width {
-                    raw_data[dst_off + x * 4 + 0] = data[src_off + x * 4 + 2];
-                    raw_data[dst_off + x * 4 + 1] = data[src_off + x * 4 + 1];
-                    raw_data[dst_off + x * 4 + 2] = data[src_off + x * 4 + 0];
-                    raw_data[dst_off + x * 4 + 3] = data[src_off + x * 4 + 3];
-                }
-            }
-            Ok(raw_data)
-        } else {
-            Ok(data.to_owned())
-        }
+        Ok(data.to_owned())
     }
 
     /// Save bitmap to RGBA PNG file
@@ -141,7 +122,7 @@ impl<'a> BitmapTarget<'a> {
         let width = self.ctx.width() as usize;
         let height = self.ctx.height() as usize;
         let mut data = self.into_raw_pixels(ImageFormat::RgbaPremul)?;
-        unpremultiply(&mut data);
+        piet_coregraphics::unpremultiply(&mut data);
         let file = BufWriter::new(File::create(path).map_err(|e| Into::<Box<_>>::into(e))?);
         let mut encoder = Encoder::new(file, width as u32, height as u32);
         encoder.set_color(ColorType::RGBA);
@@ -158,17 +139,5 @@ impl<'a> BitmapTarget<'a> {
     #[cfg(not(feature = "png"))]
     pub fn save_to_file<P: AsRef<Path>>(self, _path: P) -> Result<(), piet::Error> {
         Err(Error::MissingFeature)
-    }
-}
-#[cfg(feature = "png")]
-fn unpremultiply(data: &mut [u8]) {
-    for i in (0..data.len()).step_by(4) {
-        let a = data[i + 3];
-        if a != 0 {
-            let scale = 255.0 / (a as f64);
-            data[i] = (scale * (data[i] as f64)).round() as u8;
-            data[i + 1] = (scale * (data[i + 1] as f64)).round() as u8;
-            data[i + 2] = (scale * (data[i + 2] as f64)).round() as u8;
-        }
     }
 }
--- a/piet-common/src/cg_back.rs
+++ b/piet-common/src/cg_back.rs
@@ -57,6 +57,7 @@ pub struct Device;
 /// A struct provides a `RenderContext` and then can have its bitmap extracted.
 pub struct BitmapTarget<'a> {
     ctx: CGContext,
+    height: f64,
     phantom: PhantomData<&'a ()>,
 }
 
@@ -83,8 +84,10 @@ impl Device {
             core_graphics::base::kCGImageAlphaPremultipliedLast,
         );
         ctx.scale(pix_scale, pix_scale);
+        let height = height as f64 * pix_scale;
         Ok(BitmapTarget {
             ctx,
+            height,
             phantom: PhantomData,
         })
     }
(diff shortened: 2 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -59,29 +59,43 @@ jobs:
...
 
       - name: cargo test (windows)
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --workspace --exclude piet-cairo
+          args: --workspace --exclude piet-cairo --exclude piet-coregraphics
         if: contains(matrix.os, 'windows')
 
       - name: cargo test (not windows)
```
xStrom, line 86: Let's rename this to `cargo test (mac)`.

.github/workflows/ci.yml
```diff
@@ -59,29 +59,43 @@ jobs:
...
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --workspace --exclude piet-direct2d
-        if: contains(matrix.os, 'windows') != true
+        if: contains(matrix.os, 'macOS')
+
+      - name: cargo test (not windows)
```
xStrom, line 93: Let's rename this to `cargo test (linux)`.

piet-coregraphics/src/gradient.rs
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+    }
+}
+
+fn new_cg_gradient(stops: &[GradientStop]) -> CGGradient {
+    unsafe {
+        //FIXME: is this expensive enough we should be reusing it?
+        let space = CGColorSpace::create_with_name(kCGColorSpaceSRGB).unwrap();
+        let space_ref: *const u8 = &*space as *const CGColorSpaceRef as *const u8;
```
raphlinus, line 100: This is a mess. Let me suggest the following. Delete this line. Several more related suggestions to follow.

piet-coregraphics/src/gradient.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[link(name = "CoreGraphics", kind = "framework")]
+extern "C" {
+    fn CGGradientGetTypeID() -> CFTypeID;
+    fn CGGradientCreateWithColors(
+        space: *const u8,
+        colors: CFArrayRef,
+        locations: *const CGFloat,
+    ) -> CGGradientRef;
+    fn CGColorCreate(space: *const u8, components: *const CGFloat) -> SysCGColorRef;
```
raphlinus, line 134:
```suggestion
    #[allow(improper_ctypes)]
    fn CGColorCreate(space: *const CGColorSpaceRef, components: *const CGFloat) -> SysCGColorRef;
```

This is valid because `CGColorSpaceRef` is in foreign_types style, that is it represents the opaque object. The `#allow` is because foreign_types is missing `#[repr(C)]` on this decl, which is a problem with upstream; probably best to include a comment pointing that out.

piet-coregraphics/src/gradient.rs
```diff
@@ -0,0 +1,151 @@
...
+fn new_cg_gradient(stops: &[GradientStop]) -> CGGradient {
+    unsafe {
+        //FIXME: is this expensive enough we should be reusing it?
+        let space = CGColorSpace::create_with_name(kCGColorSpaceSRGB).unwrap();
+        let space_ref: *const u8 = &*space as *const CGColorSpaceRef as *const u8;
+        let mut colors = Vec::<CGColor>::new();
+        let mut locations = Vec::<CGFloat>::new();
+        for GradientStop { pos, color } in stops {
+            let (r, g, b, a) = Color::as_rgba(&color);
+            let color = CGColorCreate(space_ref as *const u8, [r, g, b, a].as_ptr());
```
raphlinus, line 105:
```suggestion
            let color = CGColorCreate(&*space, [r, g, b, a].as_ptr());
```

piet-coregraphics/src/gradient.rs
```diff
@@ -0,0 +1,151 @@
...
+
+fn to_cgpoint(point: Point) -> CGPoint {
+    CGPoint::new(point.x as CGFloat, point.y as CGFloat)
+}
+
+#[link(name = "CoreGraphics", kind = "framework")]
+extern "C" {
+    fn CGGradientGetTypeID() -> CFTypeID;
+    fn CGGradientCreateWithColors(
+        space: *const u8,
```
raphlinus, line 130:
```suggestion
    #[allow(improper_ctypes)]
    fn CGGradientCreateWithColors(
        space: *const CGColorSpaceRef,
```

piet-coregraphics/src/gradient.rs
```diff
@@ -0,0 +1,151 @@
...
+            let (r, g, b, a) = Color::as_rgba(&color);
+            let color = CGColorCreate(space_ref as *const u8, [r, g, b, a].as_ptr());
+            let color = CGColor::wrap_under_create_rule(color);
+            colors.push(color);
+            locations.push(*pos as CGFloat);
+        }
+
+        let colors = CFArray::from_CFTypes(&colors);
+        let gradient = CGGradientCreateWithColors(
+            space_ref as *const u8,
```
raphlinus, line 113:
```suggestion
            &*space,
```

piet-coregraphics/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+    //type StrokeStyle = StrokeStyle;
+
+    fn clear(&mut self, color: Color) {
+        let rgba = color.as_rgba_u32();
+        self.ctx.set_rgb_fill_color(
+            byte_to_frac(rgba >> 24),
+            byte_to_frac(rgba >> 16),
+            byte_to_frac(rgba >> 8),
+            byte_to_frac(rgba),
+        );
```
raphlinus, line 105:
```suggestion
        let (r, g, b, a) = color.as_rgba();
        self.ctx.set_rgb_fill_color(r, g, b, a);
```

piet-coregraphics/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+        self.ctx.scale(1.0, -1.0);
+        match brush.as_ref() {
+            Brush::Solid(color) => {
+                self.set_fill_color(color);
+                layout.draw(self.ctx);
+            }
+            Brush::Gradient(grad) => {
+                //FIXME: there's no simple way to fill text with a gradient here.
+                //To do this correctly we need to work glyph by glyph, using
+                //CTFontCreatePathForGlyph, and I'm not doing that today.
```
raphlinus, line 229: :+1:

piet-coregraphics/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    fn save(&mut self) -> Result<(), Error> {
+        self.ctx.save();
+        let state = self.transform_stack.last().copied().unwrap_or_default();
+        self.transform_stack.push(state);
+        Ok(())
+    }
+
+    //TODO: this panics in CoreGraphics if unbalanced. We could try and track stack depth
```
raphlinus, line 245: I agree. Is it a safe panic or a crash?

piet-coregraphics/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+                PathEl::ClosePath => self.ctx.close_path(),
+            }
+        }
+    }
+}
+
+fn byte_to_frac(byte: u32) -> f64 {
+    ((byte & 255) as f64) * (1.0 / 255.0)
+}
+
```
raphlinus, line 445:
```suggestion
```

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+
+use piet::kurbo::{Point, Size};
+use piet::{
+    Error, Font, FontBuilder, HitTestMetrics, HitTestPoint, HitTestTextPosition, LineMetric, Text,
+    TextLayout, TextLayoutBuilder,
+};
+
+use crate::ct_helpers::{AttributedString, Frame, Framesetter, Line};
+
+// inner is an nsfont.
```
raphlinus, line 20: This comment is somewhat cryptic.

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+    pub(crate) line_y_positions: Vec<f64>,
+    /// offsets in utf8 of lines
+    line_offsets: Vec<usize>,
+    pub(crate) frame_size: Size,
+    width_constraint: f64,
+}
+
+pub struct CoreGraphicsTextLayoutBuilder(CoreGraphicsTextLayout);
+
+pub struct CoreGraphicsText<'a>(PhantomData<&'a ()>);
```
raphlinus, line 42: This seems strange. Is there a better type (the &self when `Text` gets created) that would document the expected lifetime nesting?

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+pub struct CoreGraphicsTextLayout {
+    string: String,
+    attr_string: AttributedString,
+    framesetter: Framesetter,
+    pub(crate) frame: Option<Frame>,
+    // distance from the top of the frame to the baseline of each line
+    pub(crate) line_y_positions: Vec<f64>,
+    /// offsets in utf8 of lines
+    line_offsets: Vec<usize>,
+    pub(crate) frame_size: Size,
```
raphlinus, line 36: Note to self while reviewing: should this be a `Rect` rather than a `Size`, or can we always assume the origin is at zero?

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    fn line_metric(&self, line_number: usize) -> Option<LineMetric> {
+        let lines = self.unwrap_frame().get_lines();
+        let line = lines.get(line_number.min(isize::max_value() as usize) as isize)?;
+        let line = Line::new(&line);
+        let typo_bounds = line.get_typographic_bounds();
+        let (start_offset, end_offset) = self.line_range(line_number)?;
+        let text = self.line_text(line_number)?;
+        //FIXME: this is just ascii whitespace
```
raphlinus, line 136: What　could possibly go wrong?

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+            .map(Into::into)
+            .unwrap();
+        let fake_y = self.line_y_positions[line_num];
+        // map that back into our inverted coordinate space
+        let fake_y = -(self.frame_size.height - fake_y);
+        let point_in_string_space = CGPoint::new(point.x, fake_y);
+        let offset_utf16 = line.get_string_index_for_position(point_in_string_space);
+        let offset = match offset_utf16 {
+            // this is 'kCFNotFound'.
+            // if nothing is found just go end of string? should this be len - 1? do we have an
```
raphlinus, line 194: Gut feeling is that `len` is reasonable. `len - 1` is almost certainly wrong as it might cut a codepoint.

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+            // implicit newline at end of file? so many mysteries
+            -1 => self.string.len(),
+            n if n >= 0 => {
+                let utf16_range = line.get_string_range();
+                let utf8_range = self.line_range(line_num).unwrap();
+                let line_txt = self.line_text(line_num).unwrap();
+                let rel_offset = (n - utf16_range.location) as usize;
+                let mut off16 = 0;
+                let mut off8 = 0;
+                for c in line_txt.chars() {
```
raphlinus, line 204: I'd recommend calling into a helper function; https://github.com/linebender/piet/blob/master/piet-direct2d/src/text.rs#L230 is reasonable though we should have better ways of avoiding cross-crate duplication.

piet-coregraphics/src/text.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn hit_test_text_position(&self, offset: usize) -> Option<HitTestTextPosition> {
+        let line_num = match self.line_offsets.binary_search(&offset) {
+            Ok(line) => line.saturating_sub(1),
+            Err(line) => line.saturating_sub(1),
+        };
+        let line: Line = self.unwrap_frame().get_line(line_num)?.into();
+        let text = self.line_text(line_num)?;
+
+        let offset_remainder = offset - self.line_offsets.get(line_num)?;
+        let off16: usize = text[..offset_remainder].chars().map(char::len_utf16).sum();
```
raphlinus, line 238: This is is a clear statement of what it's trying to compute but we also have a slightly more optimized helper function (nearby link above).

piet-coregraphics/src/lib.rs
```diff
@@ -0,0 +1,542 @@
...
+};
+
+use gradient::Gradient;
+
+pub struct CoreGraphicsContext<'a> {
+    // Cairo has this as Clone and with &self methods, but we do this to avoid
+    // concurrency problems.
+    ctx: &'a mut CGContextRef,
+    // the height of the context; we need this in order to correctly flip the coordinate space
+    text: CoreGraphicsText<'a>,
```
xStrom, line 39: The comment doesn't seem to match the field. :thinking:

piet-common/src/cg_back.rs
```diff
@@ -0,0 +1,174 @@
...
+            None,
+            width,
+            height,
+            8,
+            0,
+            &CGColorSpace::create_device_rgb(),
+            core_graphics::base::kCGImageAlphaPremultipliedLast,
+        );
+        ctx.scale(pix_scale, pix_scale);
+        let height = height as f64 * pix_scale;
```
xStrom, line 90: Doesn't `bitmap_target` take the height in pixels? Looking at other platforms it seems to. Then if you want the logical height you gotta divide.

piet-coregraphics/examples/basic-cg.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let mut encoder = png::Encoder::new(w, WIDTH as u32, HEIGHT as u32);
+    encoder.set_color(png::ColorType::RGBA);
+    encoder.set_depth(png::BitDepth::Eight);
+    let mut writer = encoder.write_header().unwrap();
+
+    writer.write_image_data(cg_ctx.data()).unwrap();
+}
+
+fn unpremultiply(data: &mut [u8]) {
```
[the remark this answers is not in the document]
cmyr: Just not sure where it would live? I guess we could have some common-utils module or something in piet...

piet-coregraphics/examples/test-picture.rs
```diff
@@ -0,0 +1,46 @@
...
+        0,
+        &CGColorSpace::create_device_rgb(),
+        core_graphics::base::kCGImageAlphaPremultipliedLast,
+    );
+    cg_ctx.scale(SCALE, SCALE);
+    let mut piet_context =
+        CoreGraphicsContext::new_y_up(&mut cg_ctx, HEIGHT as f64 * SCALE.recip());
+    piet::draw_test_picture(&mut piet_context, test_picture_number).unwrap();
+    piet_context.finish().unwrap();
+    std::mem::drop(piet_context);
```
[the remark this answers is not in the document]
cmyr: good point. What's a good way to rest the premultiplication? I don't really understand the tradeoffs here, or whether we even want to be using premultiplied alpha?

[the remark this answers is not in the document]
raphlinus: In general, premultiplied is faster for rendering, but PNG stores only "separated." What I generally do for testing is render a large object with semi-transparency. For example, a white rectangle at 50% opacity will appear as gray if rendered as premultiplied and saved as png without correction.

piet-coregraphics/src/ct_helpers.rs
```diff
@@ -0,0 +1,203 @@
...
+}
+
+impl<'a> From<CTLine> for Line<'a> {
+    fn from(src: CTLine) -> Line<'a> {
+        Line(Cow::Owned(src))
+    }
+}
+
+#[link(name = "CoreText", kind = "framework")]
+extern "C" {
```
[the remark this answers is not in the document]
cmyr: I was thinking not to, but on reflection I think it would be nice to make my use of these go through code review in those crates, so I'm going to upstream as much as I can.

piet-coregraphics/src/lib.rs
cmyr: we hit an assert in some C code.

piet-coregraphics/src/text.rs
cmyr: yes, was stale. :)

piet-coregraphics/src/text.rs
cmyr: This is the same as what we do in cairo, currently.

we could pass this in a reference to the creating context, but I'm not sure that's any more meaningful?

piet-coregraphics/src/text.rs
cmyr: this is a size. It's hard to even say what origin means here, but in some sense it is at zero. (the drawing origin is the baseline of the first line, but we aren't given it explicitly)

piet-coregraphics/src/text.rs
cmyr: actually I didn't notice that apparently rust has at least some of the unicode tables? I mean: there's a `char::is_whitespace` method I can use here.

piet-coregraphics/src/text.rs
cmyr: opened #183 to track sharing code between crates.

piet-coregraphics/src/text.rs
raphlinus: Yeah, what Rust has is really limited, mostly just some case stuff and this. I think consensus is that real Unicode logic belongs in crates rather than core language.

piet-coregraphics/src/text.rs
cmyr: can fix this when #184 lands.

piet-common/src/cg_back.rs
cmyr: I think it's the height in device points? Isn't that why we pass in the pix_scale argument?

piet-coregraphics/src/text.rs
raphlinus: I'm fine leaving it as it is, it's basically documentation.

PR #177, opened 2020-04-29
Piet implementation using pathfinder

derekdreery:
I'm just opening this to get feedback. I'm not sure if you want it to live in the piet repo or separately. The bits I haven't implemented yet are: radial gradients, all text/font, styled stroke, blurred rect, and some image formats. Seems to work well - there is an example that works.


derekdreery:
I'm gonna close this as I think it's unlikely you'd get anything from using pathfinder. I'll leave the branch in the unlikely event that anyone is interested.

PR #181, opened 2020-05-07
Remove piet-test crate

cmyr:
This moves the test drawing code into a module ('samples') of the
main piet crate; this is guarded behind the 'samples' feature.


xStrom:
Actually, wait, the CI should probably be updated to enable this feature.

PR #188, opened 2020-05-10
Fixup sample_picture_0

cmyr:
This does some cleanup, and a few tweaks:
- we now draw the area that is used for clipping
- the image is simpler, to make it easier to distinguish whether
it is oriented correctly

before:
temp-cairo
after:

new-cairo


cmyr:
The gradient wasn't a gradient, exactly, it was an image we synthesized.

PR #189, opened 2020-05-11
Move blurred rect code into piet::utils

cmyr:


piet-cairo/src/lib.rs
```diff
@@ -487,3 +486,16 @@ fn matrix_to_affine(matrix: Matrix) -> Affine {
...
+
+fn compute_blurred_rect(rect: Rect, radius: f64) -> (ImageSurface, Point) {
+    let size = piet::util::size_for_blurred_rect(rect, radius);
+    // TODO: maybe not panic on error (but likely to happen only in extreme cases such as OOM)
+    let mut image =
+        ImageSurface::create(Format::A8, size.width as i32, size.height as i32).unwrap();
+    let stride = image.get_stride() as usize;
+    let mut data = image.get_data().unwrap();
+    let rect_exp = piet::util::compute_blurred_rect(rect, radius, stride, &mut *data);
+    std::mem::drop(data);
```
xStrom, line 498: Does this not get dropped at the end of the function anyway? :thinking:

cmyr: For unclear reasons we can't return `image` unless we drop this? I would have thought NLL would sort this out but apparently not.

PR #195, opened 2020-05-12
Fix current_transform with GTK cairo.

xStrom:
As discovered in [druid#922](https://github.com/xi-editor/druid/issues/922) the cairo context given by GTK has a transform already applied to it. This PR mimics the coregraphics solution to do the bookkeeping ourselves.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -23,6 +23,11 @@ pub struct CairoRenderContext<'a> {
     // concurrency problems.
     ctx: &'a mut Context,
     text: CairoText<'a>,
+    // because of the relationship between GTK and cairo (where GTK applies a transform
+    // to adjust for menus and window borders) we cannot trust the transform returned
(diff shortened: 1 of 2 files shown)


piet-cairo/src/lib.rs
```diff
@@ -39,6 +44,7 @@ impl<'a> CairoRenderContext<'a> {
         CairoRenderContext {
             ctx,
             text: CairoText::new(),
+            transform_stack: vec![Affine::default()],
```
cmyr, line 47: wow now I'm doubting my own implementation. I think this should be an empty vec?? because a call to `restore` without a call to `save` should error.

This was expected below, in `transform`, where we check if there's an item in the stack, as well as in `current_transform` where we do an `unwrap_or_default` 🤔

Does my thinking make sense to you?

xStrom: Indeed, this should definitely be empty. I pushed the change.

PR #199, opened 2020-05-15
Clean up workspace compilation and testing.

### LINKED ISSUE #192: CI does not test 'png' feature
(I am assuming, because some breakage gated on this feature presumably passed CI. see #191)

xStrom:
This PR refactors the CI script to allow for easier expansion and testing of features. The cargo manifest files are cleaned up a bit and unified in style. I also updated `README.md` to better reflect the current state of things.

Fixes  #192.

### DIFF
--- a/piet-common/Cargo.toml
+++ b/piet-common/Cargo.toml
@@ -14,23 +14,23 @@ cairo = ["piet-cairo", "cairo-rs"]
 web = ["piet-web"]
 
 [dependencies]
-cfg-if = "0.1.10"
 piet = { version = "0.1.0", path = "../piet" }
 piet-cairo = { version = "0.1.0", path = "../piet-cairo", optional = true }
 piet-coregraphics = { version = "0.1.0", path = "../piet-coregraphics", optional = true }
 piet-direct2d = { version = "0.1.0", path = "../piet-direct2d", optional = true }
 piet-web = { version = "0.1.0", path = "../piet-web", optional = true }
-cairo-rs = { version = "0.8.1", default_features = false, optional = true}
 
-[target.'cfg(not(any(target_arch="wasm32", target_os="windows", target_os="macos")))'.dependencies]
+cfg-if = "0.1.10"
+cairo-rs = { version = "0.8.1", default_features = false, optional = true }
+png = { version = "0.16.1", optional = true }
+
+[target.'cfg(target_os="linux")'.dependencies]
 piet-cairo = { version = "0.1.0", path = "../piet-cairo" }
 cairo-rs = { version = "0.8.1", default_features = false}
-png = { version = "0.16.1", optional = true }
 
 [target.'cfg(target_os="macos")'.dependencies]
 piet-coregraphics = { version = "0.1.0", path = "../piet-coregraphics" }
 core-graphics = { version = "0.19" }
-png = { version = "0.16.1", optional = true }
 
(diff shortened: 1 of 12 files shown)


piet-common/Cargo.toml
```diff
@@ -14,34 +14,36 @@ cairo = ["piet-cairo", "cairo-rs"]
...
-cairo-rs = { version = "0.8.1", default_features = false, optional = true}
 
-[target.'cfg(not(any(target_arch="wasm32", target_os="windows", target_os="macos")))'.dependencies]
+cfg-if = "0.1.10"
+cairo-rs = { version = "0.8.1", default_features = false, optional = true }
+png = { version = "0.16.1", optional = true }
+
+[target.'cfg(target_os="linux")'.dependencies]
 piet-cairo = { version = "0.1.0", path = "../piet-cairo" }
 cairo-rs = { version = "0.8.1", default_features = false}
```
cmyr, line 29: will these get added to a wasm32 build on linux? or is that not a thing?

piet-common/Cargo.toml
```diff
@@ -14,34 +14,36 @@ cairo = ["piet-cairo", "cairo-rs"]
...
 piet-web = { version = "0.1.0", path = "../piet-web", optional = true }
-cairo-rs = { version = "0.8.1", default_features = false, optional = true}
 
-[target.'cfg(not(any(target_arch="wasm32", target_os="windows", target_os="macos")))'.dependencies]
+cfg-if = "0.1.10"
+cairo-rs = { version = "0.8.1", default_features = false, optional = true }
+png = { version = "0.16.1", optional = true }
+
+[target.'cfg(target_os="linux")'.dependencies]
 piet-cairo = { version = "0.1.0", path = "../piet-cairo" }
```
cmyr, line 28: we want piet-cairo to be built if the --features=cairo option is passed on macOS. Maybe we're being too fancy though, and could just get rid of this option?

piet-common/src/cairo_back.rs
```diff
@@ -136,14 +136,14 @@ impl<'a> BitmapTarget<'a> {
...
         let width = self.surface.get_width();
         let image = self.into_raw_pixels(ImageFormat::RgbaPremul)?;
-        let file = BufWriter::new(File::create(path).map_err(|e| Into::<Box<_>>::into(e))?);
+        let file = BufWriter::new(File::create(path).map_err(Into::<Box<_>>::into)?);
         let mut encoder = Encoder::new(file, width as u32, height as u32);
         encoder.set_color(ColorType::RGBA);
         encoder
             .write_header()
-            .map_err(|e| Into::<Box<_>>::into(e))?
+            .map_err(Into::<Box<_>>::into)?
```
cmyr, line 144: huh I tried to do this recently and couldn't get it to work

piet-common/Cargo.toml
xStrom: They will not be compiled. There are no warnings or anything either, so I think it's fine to not make the cfg guard more complex here.

piet-common/Cargo.toml
xStrom: Yes the `cairo` feature still works on every platform, including Windows when using gcc.

The CI only tests cairo on Linux. If we want to ensure that cairo support is awesome then we could test it on macOS and Windows too. However my understanding has been that cairo is more of a temporary backend, so probably fine to just test it on Linux.

PR #203, opened 2020-05-25
Do doc tests with the stable compiler.

xStrom:
The doc tests are marked as required, so they should be done with the stable compiler instead.


xStrom:
I think there's still value in running the `doc` command because it'll complain if multiple incompatible library version docs will overwrite eachother. Could also make sense to turn on more docs lints in the future.

PR #204, opened 2020-05-26
WebRender clear uses canvas pixel dimensions instead of DOM layout di…

bdon:
On a display with device pixel density > 2, the offset dimensions for a 1024x1024 pixel canvas element with layout dimensions 512x512 will be 512x512, meaning context.clear() would only clear the top left quadrant.

this just uses the width/height properties instead of offset_width/offset_height.

### DIFF
--- a/piet-web/src/lib.rs
+++ b/piet-web/src/lib.rs
@@ -115,7 +115,7 @@ impl<'a> RenderContext for WebRenderContext<'a> {
 
     fn clear(&mut self, color: Color) {
         let (width, height) = match self.ctx.canvas() {
-            Some(canvas) => (canvas.offset_width(), canvas.offset_height()),
+            Some(canvas) => (canvas.width(), canvas.height()),
             None => return,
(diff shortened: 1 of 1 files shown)


xStrom:
I wonder how this interacts with scale. Druid sets the canvas scale factor, so `clear` probably works fine. Using `width`/`height` would then be too large. Perhaps the solution is to fetch the scale factor in `clear` and divde `width`/`height` by that.

This needs more testing, I'm just theorizing here.

bdon:
Hmm, good catch. That made me realize I had a bug in my code where I was doing the clear() on a transformed RenderContext in some cases.

However, it does seem like managing the Context's transformation matrix stack correctly is a reasonable demand for the user of the API.

 I think expecting that the matrix = identity when `clear()` is called is well defined behavior, in the absence of alternatives.

xStrom:
Okay so maybe the correct behavior for the `clear` implementation is some sort of `with_save` closure which clears transform/scale before doing the fill with `width`/`height`.

PR #209, opened 2020-06-06
Remove lifetime from D2DText

ForLoveOfCats:
For now I am just experimenting and abusing CI to test stuff I cannot test locally


ForLoveOfCats:
That took a few hours but all the CI passes! I'm sure I've overlooked some important details but I think this is ready for a tentative review

cmyr:
@ForLoveOfCats good to go?

ForLoveOfCats:
I think so? As long as there are no objections

PR #210, opened 2020-06-07
Add support to drawing text with gradient brush (piet-coregraphics)

brunogouveia:
Hey guys, I was looking into adding support to Core Graphics backend a few months ago (https://github.com/brunogouveia/piet/commits/piet-core-graphics), but I never had the time to finish.

I looked into the current core graphics backend implementation and I noticed text rendering with gradient brushes was missing, so I created this PR adding gradient support, that uses the same approach I used on my branch: set `text drawing mode` to clip and render the gradient afterwards.


.gitignore
```diff
@@ -3,4 +3,9 @@ target/
 Cargo.lock
 .DS_Store
 .cargo
+*.png
 
+## Intellij Idea project configuration files
+.idea/
+## VScode project configuration files
+.vscode/
```
cmyr, line 11: we might want to add the newline here or some things will complain
```suggestion
.vscode/

```

PR #213, opened 2020-06-15
For convenience, use a shared reference to the Cairo context

siebelmann:
The reference kept by `CairoRenderContext` should be shared. The Cairo context is internally reference-counted and, as far as I can tell, thread-safe.

The fact that it now keeps an exclusive reference is an inconvenience in combination with GTK. The callback of the [`draw` signal](https://docs.rs/gtk/0.8.1/gtk/trait.WidgetExt.html#tymethod.connect_draw) is only provided a shared reference to a context which means that you have to clone it first.

### DIFF
--- a/piet-cairo/examples/test-picture.rs
+++ b/piet-cairo/examples/test-picture.rs
@@ -19,9 +19,9 @@ fn main() {
 
     let surface = ImageSurface::create(Format::ARgb32, size.width as i32, size.height as i32)
         .expect("Can't create surface");
-    let mut cr = Context::new(&surface);
+    let cr = Context::new(&surface);
     cr.scale(HIDPI, HIDPI);
-    let mut piet_context = CairoRenderContext::new(&mut cr);
+    let mut piet_context = CairoRenderContext::new(&cr);
     draw_test_picture(&mut piet_context, test_picture_number).unwrap();
     piet_context.finish().unwrap();
     surface.flush();
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -23,7 +23,7 @@ pub use crate::text::{
 pub struct CairoRenderContext<'a> {
     // Cairo has this as Clone and with &self methods, but we do this to avoid
     // concurrency problems.
-    ctx: &'a mut Context,
+    ctx: &'a Context,
     text: CairoText,
     // because of the relationship between GTK and cairo (where GTK applies a transform
     // to adjust for menus and window borders) we cannot trust the transform returned
@@ -38,7 +38,7 @@ impl<'a> CairoRenderContext<'a> {
     /// At the moment, it uses the "toy text API" for text layout, but when
     /// we change to a more sophisticated text layout approach, we'll probably
     /// need a factory for that as an additional argument.
-    pub fn new(ctx: &mut Context) -> CairoRenderContext {
+    pub fn new(ctx: &Context) -> CairoRenderContext {
         CairoRenderContext {
             ctx,
             text: CairoText::new(),
--- a/piet-common/src/cairo_back.rs
+++ b/piet-common/src/cairo_back.rs
@@ -106,7 +106,7 @@ impl<'a> BitmapTarget<'a> {
     /// Note: caller is responsible for calling `finish` on the render
     /// context at the end of rendering.
     pub fn render_context(&mut self) -> CairoRenderContext {
-        CairoRenderContext::new(&mut self.cr)
+        CairoRenderContext::new(&self.cr)
     }
 
     /// Get raw RGBA pixels from the bitmap by copying them into `buf`. If all the pixels were


cmyr:
@raphlinus this needs your input.

raphlinus:
Hmm, I'd appreciate a more detailed explanation of the problem, though I get a sense of it. The problem with a change like this is that we're trying to keep the signatures consistent across the different back-ends. The specific risk that I'm worried about is developing against piet-cairo, assuming that a non-mutable reference is available, and then finding that switching to a different back-end fails.

All that said, I'm open to solving integration problems with gtk, I just want a clearer picture.

siebelmann:
Sure, I’ll give an example. Let's say I've written a great little function that paints a picture with piet:
```rust
fn paint_picture_with_piet(ctx: &mut impl RenderContext, width: f64, height: f64) {
    // ... here be awesome drawing
}
```
I’ve used the SVG back-end so far, but now I want to show my pretty picture in a Gtk app.

I create a drawing area widget. Any time that widget needs to be drawn (at application start, on window resize, etc.) a draw signal is emitted. So I connect to the draw signal and try to turn the *shared/immutable* reference to a Cairo context that I am given by Gtk into a `CairoRenderContext`.
```rust
// ... set up the app window, etc.
let canvas = gtk::DrawingArea::new();
canvas.connect_draw(|widget, cairo_ctx: &cairo::Context| {
    let mut piet_ctx = piet_cairo::CairoRenderContext::new(cairo_ctx);     
    //                   error: types differ in mutability ^^^^^^^^^

    let width = widget.get_allocated_width() as f64;
    let height = widget.get_allocated_height() as f64;
    paint_picture_with_piet(&mut piet_ctx, width, height);
});
```
I look into the Cairo docs to find out how to get a mutable reference from an immutable one and the only solution that I see is to clone it. But is that a good idea? I don't know that much about Cairo–that’s why I chose to use an abstraction in the first place. Will that clone also create a new surface? How expensive is it? After all the callback might be called many times per second. So I have to do research and look into the code. After I've assured myself that it’s cheap and safe my code looks like this:
```rust
canvas.connect_draw(|widget, cairo_ctx: &cairo::Context| {
    let mut my_ctx = cairo_ctx.clone();
    let mut piet_ctx = piet_cairo::CairoRenderContext::new(&mut my_ctx);
    // ...
}
```
It's not terrible and it works. But it’s inconvenient.

With the changes in this pull request the first example will work and save the headache. Existing code won’t break apart from warnings about unnecessary mutability.

Keeping the signatures of all back-ends in line would be nice, but it's hard when you rely on external libraries with different conventions. The current signatures of the other back-end creation functions in this repository show that:

```rust
pub fn new_y_up(ctx: &mut CGContextRef, height: f64) -> CoreGraphicsContext {}
pub fn new_y_down(ctx: &mut CGContextRef) -> CoreGraphicsContext {}

pub fn new(
        factory: &'a D2DFactory,
        dwrite: DwriteFactory,
        rt: &'b mut DeviceContext,
    ) -> D2DRenderContext<'b> {}

pub fn new() -> piet_svg::RenderContext {}

pub fn new(ctx: CanvasRenderingContext2d, window: Window) -> WebRenderContext {}
```

I think it's the right move to follow Cairo’s convention and accept a shared reference for `CairoRenderContext`.

If you choose not to go this way I would suggest to at least add documentation to `CairoRenderContext::new` that states that it’s okay to clone the Cairo context first. I’d be willing to write it.

cmyr:
I believe this makes sense. @raphlinus?

psychon:
> The Cairo context is internally reference-counted and, as far as I can tell, thread-safe.

I'm not sure what exactly you mean with "thread-safe", but a cairo context is certainly not thread-safe. It is `!Send` and `!Sync` in cairo-rs: https://docs.rs/cairo-rs/0.8.1/cairo/struct.Context.html#impl-Send
The underlying C code is also not thread-safe in the sense that you must not use a single `cairo_t*` from multiple threads (but may use different cairo contexts from different threads). For example, `cairo_set_dash` looks like this (after passing through two layers of abstraction that do not acquire any locks):
```C
cairo_status_t
_cairo_gstate_set_dash (cairo_gstate_t *gstate, const double *dash, int num_dashes, double offset)
{
    double dash_total, on_total, off_total;
    int i, j;

    free (gstate->stroke_style.dash);

    gstate->stroke_style.num_dashes = num_dashes;

[more code follows]
```
The above will certainly "go bad" if called from multiple threads simultaneously.

Everything else said here makes sense to me, but I am just a bit worried about this *`cairo::Context` is thread-safe*. Sorry if I misunderstood.

siebelmann:
Thanks for the clarification. I guess I misused the term thread-safe. The point I was trying to make is this: You don’t need a `&mut` reference to prevent things from going bad in combination with threads.

cmyr:
@siebelmann this needs a rebase it looks like?

siebelmann:
@cmyr rebase is done. Do you need me to do anything else before this can be merged?

cmyr:
looks good, thanks!

Issue #212, opened 2020-06-15
Copying pixels out of a BitmapTarget?


jneem:
The only current way to get pixel data out of a `BitmapTarget` involves consuming it. Would you accept a patch that adds a function with a signature like this?
```rust
impl BitmapTarget {
  fn copy_raw_pixels(&self, fmt: ImageFormat, buf: &mut [u8]) -> Result<usize, Error>
}
```
It would return the number of bytes written, or an error if the buf isn't big enough.

My use-case for this is encoding video: I want to render to a `BitmapTarget` and then pass the data to gstreamer for encoding. My data is mostly unchanged from frame to frame; I want to reuse the bitmap so that I only need to render the changed parts.

raphlinus:
Yes, this is reasonable, as long as it works reliably across back-ends. I understand the usecase.

PR #214, opened 2020-06-16
Add a BitmapTarget::copy_raw_pixels function

jneem:
This adds the functions that I suggested in #212. There are two sad things about this PR that I couldn't figure out how to improve:

- on web, I couldn't find a way to access the borrowed pixels, so `copy_raw_pixels` has to copy them twice :|
- cairo-rs has some safety checks that made it a pain to get the data out without destroying the context first. I ended up needing to use unsafe.

### DIFF
--- a/piet-common/src/cairo_back.rs
+++ b/piet-common/src/cairo_back.rs
@@ -126,35 +126,51 @@ impl<'a> BitmapTarget<'a> {
             // get the data from an ImageSurface that's still referenced by a context. The C docs
             // don't seem to think that's a problem, as long as we call flush (which we already
             // did), and promise not to mutate anything.
+            // https://www.cairographics.org/manual/cairo-Image-Surfaces.html#cairo-image-surface-get-data
             let data = {
                 let data_ptr = cairo_sys::cairo_image_surface_get_data(self.surface.to_raw_none());
                 if data_ptr.is_null() {
                     let err = cairo::BorrowError::from(cairo::Status::SurfaceFinished);
                     return Err((Box::new(err) as Box<dyn std::error::Error>).into());
                 }
-                std::slice::from_raw_parts(data_ptr, size)
+                std::slice::from_raw_parts(data_ptr, height.saturating_sub(1) * stride + width * 4)
             };
             for y in 0..height {
                 let src_off = y * stride;
                 let dst_off = y * width * 4;
                 for x in 0..width {
-                    buf[dst_off + x * 4 + 0] = data[src_off + x * 4 + 2];
-                    buf[dst_off + x * 4 + 1] = data[src_off + x * 4 + 1];
-                    buf[dst_off + x * 4 + 2] = data[src_off + x * 4 + 0];
-                    buf[dst_off + x * 4 + 3] = data[src_off + x * 4 + 3];
+                    // These unchecked indexes allow the autovectorizer to shine.
+                    // Note that dst_off maxes out at (height - 1) * width * 4, and so
+                    // dst_off + x * 4 + 3 maxes out at height * width * 4 - 1, which is size - 1.
+                    // Also, src_off maxes out at (height - 1) * stride, and so
+                    // src_off + x * 4 + 3 maxes out at (height - 1) * stride + width * 4 - 1.
+                    *buf.get_unchecked_mut(dst_off + x * 4 + 0) =
+                        *data.get_unchecked(src_off + x * 4 + 2);
+                    *buf.get_unchecked_mut(dst_off + x * 4 + 1) =
+                        *data.get_unchecked(src_off + x * 4 + 1);
+                    *buf.get_unchecked_mut(dst_off + x * 4 + 2) =
+                        *data.get_unchecked(src_off + x * 4 + 0);
+                    *buf.get_unchecked_mut(dst_off + x * 4 + 3) =
+                        *data.get_unchecked(src_off + x * 4 + 3);
                 }
             }
-        };
+        }
         Ok(size)
     }
 
     /// Get raw RGBA pixels from the bitmap.
-    pub fn into_raw_pixels(mut self, fmt: ImageFormat) -> Result<Vec<u8>, piet::Error> {
+    pub fn to_raw_pixels(&mut self, fmt: ImageFormat) -> Result<Vec<u8>, piet::Error> {
         let width = self.surface.get_width() as usize;
         let height = self.surface.get_height() as usize;
         let mut buf = vec![0; width * height * 4];
         self.copy_raw_pixels(fmt, &mut buf)?;
         Ok(buf)
+    }
+
+    /// Get raw RGBA pixels from the bitmap.
+    #[deprecated(since = "0.2.0", note = "use to_raw_pixels")]
+    pub fn into_raw_pixels(mut self, fmt: ImageFormat) -> Result<Vec<u8>, piet::Error> {
+        self.to_raw_pixels(fmt)
     }
 
     /// Save bitmap to RGBA PNG file
--- a/piet-common/src/cairo_back.rs
+++ b/piet-common/src/cairo_back.rs
@@ -135,6 +135,14 @@ impl<'a> BitmapTarget<'a> {
                 }
                 std::slice::from_raw_parts(data_ptr, height.saturating_sub(1) * stride + width * 4)
             };
+
+            // A sanity check for all the unsafe indexing that follows.
+            let src_off_max = (height - 1) * stride;
+            let dst_off_max = (height - 1) * width * 4;
+            let x_max = width - 1;
+            buf.get(dst_off_max + x_max * 4 + 3).unwrap();
+            data.get(src_off_max + x_max * 4 + 3).unwrap();
+
             for y in 0..height {
                 let src_off = y * stride;
                 let dst_off = y * width * 4;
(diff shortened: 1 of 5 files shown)


piet-common/src/cairo_back.rs
```diff
@@ -101,33 +101,60 @@ impl<'a> BitmapTarget<'a> {
...
-                raw_data[dst_off + x * 4 + 2] = buf[src_off + x * 4 + 0];
-                raw_data[dst_off + x * 4 + 3] = buf[src_off + x * 4 + 3];
-            }
+        let size = width * height * 4;
+        if buf.len() < size {
+            return Err(piet::Error::InvalidInput);
         }
-        Ok(raw_data)
+        unsafe {
+            // Cairo's rust wrapper has extra safety checks that we want to avoid: it won't let us
```
cmyr, line 125: I would open an issue upstream and see if cairo-rs is open to adding API for this.

For what it's worth, I believe it was the lack of this API that is the reason this functionality was not originally implemented.

jneem:
> do we want the user to pass in a buffer that we copy to, or do we want to just return data? I think this comes down to how we expect this API to be used; if there's a reasonable case that the user should be reusing a buffer, then I can imagine it being passed in, but if not I think we should be just returning data, which is simpler, API-wise.

I have a use-case for passing in a buffer, although I'm not sure how common it is: to interface with `gstreamer`, I need to use its own buffer types. Those buffers allow me to get a `&mut [u8]`, but as far as I can tell there is no way to convert a `Vec<u8>` into one of gstreamer's buffers.

> My basic belief around public Rust API is that both inout params and tuple return values should be avoided

The `std::io::Read` API has precedent for the pattern of having a low-level interface that takes a mutable buffer parameter, together with high-level interfaces that provide owned return values.

piet-common/src/cairo_back.rs
jneem: I'll go do that now.

cmyr:
yea, so `std::io::Read` is exactly the situation where you generally want to reuse a buffer. It sounds like you have a similar situation.

One option would be to have `read_raw_pixels` (basically what you've implemented) and then also have `to_raw_pixels`, that just creates a vec (maybe precomputing size and using `with_capacity`) and passes it to `read_raw_pixels`.

piet-common/src/cairo_back.rs
cmyr: Also: I *am* okay with having the unsafe here now, but I would include a link to the docs that support the assertion that immutable access is okay.

piet-common/src/cairo_back.rs
```diff
@@ -101,33 +101,60 @@ impl<'a> BitmapTarget<'a> {
...
+            // get the data from an ImageSurface that's still referenced by a context. The C docs
+            // don't seem to think that's a problem, as long as we call flush (which we already
+            // did), and promise not to mutate anything.
+            let data = {
+                let data_ptr = cairo_sys::cairo_image_surface_get_data(self.surface.to_raw_none());
+                if data_ptr.is_null() {
+                    let err = cairo::BorrowError::from(cairo::Status::SurfaceFinished);
+                    return Err((Box::new(err) as Box<dyn std::error::Error>).into());
+                }
+                std::slice::from_raw_parts(data_ptr, size)
```
cmyr, line 135: I'm not sure that using `size` here is correct; what happens if `stride` is different from `width`?

cmyr: Also (and please forgive my spam) one other little thing: I would try to figure out what the smallest `unsafe` function we can write is, and write that (the byte copying stuff below doesn't need to be unsafe, for instance.)

jneem: If `stride` is different from `width`, we potentially read a different region of the source data. But no matter what, the part of `buf` that gets written is `[0..size]`.

Regarding the unsafe block, rustc will accept
````rust
    let data = unsafe { ... };
```

I made it bigger, so that it encompasses the entire region where we need to maintain safety-related invariants. (Specifically, we need to avoid any mutations while `data` is live.) Do you still prefer the smaller version?

cmyr: but we're creating a slice with a size that doesn't take the `stride` into account, so that we can potentially index out of bounds? This isn't a soundness concern (we won't ever be creating a slice that is larger than our source buffer) but we might at least crash on a bounds check?

Also thinking about that it's possible that this is bounds checking every index, which would be annoying, and might be something to check if we're worried about performance.

jneem: Ah, sorry, I thought you were talking about the returned size value :facepalm: I'll fix this...

jneem: Regarding the bounds checks, unfortunately you're right. Here's the inner loop, with 8 bounds checks per iteration.
```
.LBB12_7:
 leaq    (%r10, %rcx), %rsi
 addq    $2, %rsi
 cmpq    %rdx, %rsi
 jae     .LBB12_24
 leaq    (%rcx, %rbp), %rsi
 cmpq    %r15, %rsi
 jae     .LBB12_25
 movzbl  -1(%rax, %rcx), %ebx
 movb    %bl, -3(%r14, %rcx)
 leaq    (%r10, %rcx), %rsi
 addq    $1, %rsi
 cmpq    %rdx, %rsi
 jae     .LBB12_27
 leaq    (%rcx, %rbp), %rsi
 addq    $1, %rsi
 cmpq    %r15, %rsi
 jae     .LBB12_28
 leaq    (%r10, %rcx), %rsi
 movzbl  -2(%rax, %rcx), %ebx
 movb    %bl, -2(%r14, %rcx)
 cmpq    %rdx, %rsi
 jae     .LBB12_29
 leaq    (%rcx, %rbp), %rsi
 addq    $2, %rsi
 cmpq    %r15, %rsi
 jae     .LBB12_30
 movzbl  -3(%rax, %rcx), %ebx
 movb    %bl, -1(%r14, %rcx)
 leaq    (%r10, %rcx), %rsi
 addq    $3, %rsi
 cmpq    %rdx, %rsi
 jae     .LBB12_31
 leaq    (%rcx, %rbp), %rsi
 addq    $3, %rsi
 cmpq    %r15, %rsi
 jae     .LBB12_32
 addq    $1, %rdi
 movzbl  (%rax, %rcx), %ebx
 movb    %bl, (%r14, %rcx)
 addq    $4, %rcx
 cmpq    %r12, %rdi
 jb      .LBB12_7
```

After adding get_unchecked:
```
 .LBB12_20:
 movdqu  xmm3, xmmword, ptr, [rcx, +, 4*rbx]
 movdqu  xmm2, xmmword, ptr, [rcx, +, 4*rbx, +, 16]
 movdqu  xmm1, xmmword, ptr, [rcx, +, 4*rbx, +, 32]
 movdqu  xmm4, xmmword, ptr, [rcx, +, 4*rbx, +, 48]
 movdqa  xmm5, xmm2
 punpckhbw xmm5, xmm0
<snip>
 packuswb xmm1, xmm5
 movdqu  xmmword, ptr, [rsi, +, 4*rbx, +, 32], xmm1
 movdqu  xmmword, ptr, [rsi, +, 4*rbx, +, 48], xmm4
 movdqu  xmmword, ptr, [rsi, +, 4*rbx], xmm3
 movdqu  xmmword, ptr, [rsi, +, 4*rbx, +, 16], xmm2
 add     rbx, 16
 cmp     r11, rbx
 jne     .LBB12_20
```
Unrolling and autovectorization ftw!

jneem:
Clippy doesn't like `to_raw_pixels`, because it takes `&mut self`, so I just went with `raw_pixels`.

piet-common/src/cairo_back.rs
```diff
@@ -101,41 +101,84 @@ impl<'a> BitmapTarget<'a> {
...
+            for y in 0..height {
+                let src_off = y * stride;
+                let dst_off = y * width * 4;
+                for x in 0..width {
+                    // These unchecked indexes allow the autovectorizer to shine.
+                    // Note that dst_off maxes out at (height - 1) * width * 4, and so
+                    // dst_off + x * 4 + 3 maxes out at height * width * 4 - 1, which is size - 1.
+                    // Also, src_off maxes out at (height - 1) * stride, and so
+                    // src_off + x * 4 + 3 maxes out at (height - 1) * stride + width * 4 - 1.
+                    *buf.get_unchecked_mut(dst_off + x * 4 + 0) =
```
cmyr, line 147: if we're doing this I would like us to do one checked get at the end of our range before the loop, as a way of asserting that this is not UB.

piet-common/src/cairo_back.rs
```diff
@@ -101,41 +101,84 @@ impl<'a> BitmapTarget<'a> {
         CairoRenderContext::new(&mut self.cr)
     }
 
-    /// Get raw RGBA pixels from the bitmap.
-    pub fn into_raw_pixels(mut self, fmt: ImageFormat) -> Result<Vec<u8>, piet::Error> {
+    /// Get raw RGBA pixels from the bitmap by copying them into `buf`. If all the pixels were
+    /// copied, returns the number of bytes written. If `buf` wasn't big enough, returns an error
+    /// and doesn't write anything.
+    pub fn copy_raw_pixels(
```
cmyr, line 107: do these actually need `&mut self`? I guess if we need to flush the buffer or something?

jneem: That's the current behavior, yes. Technically, cairo allows a flush without `&mut self`, but the d2d backend does require `&mut self`.

piet-common/src/cairo_back.rs
cmyr: what I was thinking here is something like,

```rust
let data_len = height.saturating_sub(1) * stride + width * 4;
let data = { .. };
// all get_unchecked are in this range must succeed
assert!(data.get(data_len - 1).is_some());
```

In any case I don't think this is strictly necessary, but I find things like this helpful when reading the code? :shrug:

jneem: Yup, that does look clearer. Thanks!

cmyr:
awesome, thanks for this!

PR #215, opened 2020-06-19
Flush the surface on `finish`.

jneem:
Should the cairo surface be flushed on `finish`? It should definitely be flushed somewhere...

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -243,6 +243,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
     }
 
     fn finish(&mut self) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


psychon:
Testing tipp: Set `CAIRO_DEBUG=xrender-version=-1.-1` as an env var (yes, there is an equal sign inside the value of the env var). This makes cairo pretend the X11 server does not support the Render extension. This causes it to hit the code path that usually only ends up not being used. Since the X11 server cannot do most of the drawing operations, you are almost guaranteed to hit a fallback. This means that cairo downloads the contents of your drawable to a cairo image surface and then does all the drawing locally. Uploading that image after every single drawing operation would be quite slow, so it is only uploaded when the surface is flushed.

PR #217, opened 2020-06-25
Make Device Send and !Sync on all backends.

### LINKED ISSUE #211: `Send` for `Device`?
Can `Device` be `Send`? On Windows, it currently isn't, although the [API docs](https://docs.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-render-multi-thread-intro) suggest that `D3D11Device` is Sync and `D3D11DeviceContext` is Send, and [these docs](https://docs.microsoft.com/en-us/windows/win32/direct2d/multi-threaded-direct2d-apps) suggest that everything `d2d` is Sync. I couldn't find anything about DirectWrite, but probably if something needs to be called from a particular thread, then it would be documented?

Edit: if `Device` can't be made Send, maybe it should be !Send on all platforms? Otherwise it's a portability footgun.

jneem:
Fixes #211, and also makes Device !Sync, jut to reduce portability friction.

Note that there are currently no checks for this. If you're willing to add `static-assertions` as a dev-dep, I could add tests...

### DIFF
--- a/piet-common/Cargo.toml
+++ b/piet-common/Cargo.toml
@@ -45,3 +45,6 @@ wasm-bindgen = "0.2.59"
 version = "0.3.36"
 features = ["console", "Window", "CanvasGradient", "CanvasRenderingContext2d", "CanvasWindingRule",
             "Document", "Element", "HtmlCanvasElement", "ImageBitmap", "ImageData", "TextMetrics"]
+
+[dev-dependencies]
+static_assertions = "1.1.0"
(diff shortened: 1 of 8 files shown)


cmyr:
totally happy to add the dev dependency if you think it will save someone some annoyance down the road.

jneem:
Ok, there it is. I think it's pretty easy for a change to fall through the cracks otherwise.

PR #219, opened 2020-06-26
Use new upstream methods for various core-graphics/text things

cmyr:
These weren't available when this was first written, but they
let us clean things up quite a bit.

### DIFF
--- a/piet-coregraphics/src/ct_helpers.rs
+++ b/piet-coregraphics/src/ct_helpers.rs
@@ -9,7 +9,6 @@ use core_foundation::{
     attributed_string::CFMutableAttributedString,
     base::TCFType,
     boolean::CFBoolean,
-    dictionary::CFDictionaryRef,
     string::{CFString, CFStringRef},
 };
 use core_foundation_sys::base::CFRange;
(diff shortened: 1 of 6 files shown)


piet-coregraphics/src/ct_helpers.rs
```diff
@@ -131,38 +109,25 @@ impl<'a> Line<'a> {
...
+        self.0.get_typographic_bounds()
     }
 
     pub(crate) fn get_string_index_for_position(&self, position: CGPoint) -> CFIndex {
-        unsafe { CTLineGetStringIndexForPosition(self.0.as_concrete_TypeRef(), position) }
+        self.0.get_string_index_for_position(position)
     }
 
-    /// return the 'primary' and 'secondary' offsets on the given line that the boundary of the
+    /// Return the 'primary' and offsets on the given line that the boundary of the
```
raphlinus, line 123: Minor editing nit, "and" should be deleted.

piet-coregraphics/src/ct_helpers.rs
```diff
@@ -131,38 +109,25 @@ impl<'a> Line<'a> {
...
     /// character at the provided index.
     ///
+    /// There is a 'secondary' offset that is not returned by the core-text crate.
```
raphlinus, line 126: Again, the secondary offset is for BiDi. I suggest language along the lines of "the underlying Core Text function also returns a secondary offset, which is useful for BiDi, but since we are not implementing that yet, we ignore it for now."

PR #220, opened 2020-06-29
Work around core-foundation changes.

jneem:
Looks like core-foundation broke semver [here](https://github.com/servo/core-foundation-rs/commit/b68e9c782af87534d50596d7c2a7b965e067eaef), leading to CI failures in druid.

### DIFF
--- a/piet-coregraphics/Cargo.toml
+++ b/piet-coregraphics/Cargo.toml
@@ -13,7 +13,7 @@ piet = { version = "0.2.0", path = "../piet" }
 
 core-graphics = "0.19.1"
 core-text = "16.0"
-core-foundation = "0.7.1"
+core-foundation = "=0.7.1"
 core-foundation-sys = "0.7"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


cmyr:
ugh, that's annoying.

Will report upstream, and see what they do; they may yank.

piet-coregraphics/Cargo.toml
```diff
@@ -13,7 +13,7 @@ piet = { version = "0.2.0", path = "../piet" }
 
 core-graphics = "0.19.1"
 core-text = "16.0"
-core-foundation = "0.7"
+core-foundation = "0.7.1"
```
cmyr, line 16:
```suggestion
core-foundation = "=0.7.1"
```

If they end up yanking it's possible that 0.7.2 would be just a revert and we would break again, so let's pin exactly to this release until we know what's up.

jneem:
Looks like they were quick with the yank button. Maybe we don't need this after all...

PR #227, opened 2020-07-03
Text alignment

cmyr:
This adds a `TextAlignment` enum, which can be used with `TextLayoutBuilder`.

It also uses this on mac and windows (windows impl was written blind, will fix it up shortly).

This is based off of #225, because it includes a test drawing that exercises these layouts, and which on macOS looks like,

coregraphics-test-7

Also looking at that sample is making me think that we're not doing hidpi correctly when writing out the PNGs or something? I'll have to look into that.

### DIFF
--- a/piet-coregraphics/src/ct_helpers.rs
+++ b/piet-coregraphics/src/ct_helpers.rs
@@ -2,13 +2,15 @@
 
 use std::borrow::Cow;
 use std::convert::TryInto;
+use std::ffi::c_void;
 use std::ops::Deref;
 
 use core_foundation::{
     array::{CFArray, CFArrayRef, CFIndex},
     attributed_string::CFMutableAttributedString,
-    base::TCFType,
+    base::{CFTypeID, TCFType},
     boolean::CFBoolean,
+    declare_TCFType, impl_TCFType,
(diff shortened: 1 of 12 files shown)


piet-coregraphics/src/ct_helpers.rs
```diff
@@ -34,22 +40,98 @@ pub(crate) struct Frame(pub(crate) CTFrame);
...
+    //FirstLineHeadIndent = 1,
+    //HeadIndent = 2,
+    //TailIndent = 3,
+    //TabStops = 4,
+    //TabInterval = 5,
+    //LineBreakMode = 6,
+    // there are many more of these
+}
+
+#[repr(u8)]
```
raphlinus, line 65: This all seems a bit strange, but I'll trust that you researched and tested this.

cmyr: indeed. Had a few choices here, this feels the most true to the semantics of CoreText. The `CTParagraphStyleSetting` business is pretty hairy API; this is the bit I'd got hung up on, because the field ordering in the docs is different from the actual layout of the struct, which I eventually found in a header.

PR #229, opened 2020-07-10
Initial rich text support

cmyr:
This adds support for rich text, with implementations for mac and windows.

A few things are in a funny state as a result of this work:

***font construction***: currently to construct a font you pass a name and a size, and get back a `FontBuilder`; the idea at the time was likely that you would set properties like weight/style on this builder, and then build your font object. We're moving closer to a world like CSS, where you don't generally use font objects directly, but instead specify familes/sizes/weights individually on the layout. This has some API implications:
- `FontBuilder` can go, and methods that construct fonts can return `Option<Font>`.
- The `Text::font` method is semantically less about *creating* a font than it is about validating the existence of a font family.
- `TextLayoutBuilder` should get methods for setting the default attributes for the layout.
- `Text::new_text_layout` might also take an explicit size, if size is no longer included in the `Font` object.

***brushes and text drawing***: currently we pass a brush when drawing text. With this work we've moved to allowing text color to be set when constructing the layout. I think we should remove the brush argument when drawing text, but this _is_ a regression in the case of wanting to draw text with a gradient or similar. Making these things work well together is tricky, especially in a platform independent way; I haven't looked into it too seriously, but one option might be to make the `brush` argument optional, in which case it would override attributes on the layout when it was passed. Another option would be to let a 'default brush' be set on the layout, which would have the same behaviour. At least on macOS it is going to be difficult to support mixing of colored regions with things like gradient drawing, so I think it should be an either/or thing.

### DIFF
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -1,17 +1,24 @@
 //! Text related stuff for the coregraphics backend
 
+use std::ops::{Range, RangeBounds};
+
+use core_foundation::base::TCFType;
+use core_foundation::dictionary::CFMutableDictionary;
+use core_foundation::number::CFNumber;
+use core_foundation::string::CFString;
 use core_foundation_sys::base::CFRange;
 use core_graphics::base::CGFloat;
+use core_graphics::color::CGColor;
 use core_graphics::context::CGContextRef;
 use core_graphics::geometry::{CGPoint, CGRect, CGSize};
 use core_graphics::path::CGPath;
-use core_text::font::{self, CTFont};
+use core_text::{font, font::CTFont, font_descriptor, string_attributes};
 
 use piet::kurbo::{Point, Size};
 use piet::util;
 use piet::{
-    Error, Font, FontBuilder, HitTestMetrics, HitTestPoint, HitTestTextPosition, LineMetric, Text,
-    TextAlignment, TextLayout, TextLayoutBuilder,
+    Error, Font, FontBuilder, FontWeight, HitTestMetrics, HitTestPoint, HitTestTextPosition,
+    LineMetric, Text, TextAlignment, TextAttribute, TextLayout, TextLayoutBuilder,
 };
 
 use crate::ct_helpers::{AttributedString, Frame, Framesetter, Line};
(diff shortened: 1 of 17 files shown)


hugo1707:
Great work,

piet-coregraphics/src/text.rs
```diff
@@ -35,11 +43,228 @@ pub struct CoreGraphicsTextLayout {
...
+            self.italic = Span::new(false, remaining_range);
+        }
+    }
+}
+
+/// coretext uses a float in the range -1.0..=1.0, which has a non-linear mapping
+/// to css-style weights. This is a fudge, adapted from QT:
+///
+/// https://git.sailfishos.org/mer-core/qtbase/commit/9ba296cc4cefaeb9d6c5abc2e0c0b272f2288733#1b84d1913347bd20dd0a134247f8cd012a646261_44_55
+fn convert_to_coretext(weight: FontWeight) -> CFNumber {
```
raphlinus, line 255: I would have a TODO to implement piecewise linear interpolation, which I think is the strategy to minimize the error. But this is Good Enough.

piet-direct2d/src/dwrite.rs
```diff
@@ -3,24 +3,36 @@
...
 use wio::com::ComPtr;
-use wio::wide::ToWide;
+use wio::wide::{FromWide, ToWide};
 
-use piet::TextAlignment;
+use piet::{FontWeight, TextAlignment};
+
+use crate::Brush;
+
+/// "en-US" as null-terminated utf16.
```
raphlinus, line 34: Note: https://www.reddit.com/r/rust/comments/hse568/announce_utf16_lit100_for_all_your_utf16_literal/

piet-direct2d/src/dwrite.rs
```diff
@@ -38,25 +50,23 @@ pub struct DwriteFactory(ComPtr<IDWriteFactory>);
...
+/// A cheap to clone reference to an existing family name.
 ///
-/// TODO: provide lots more capability.
-pub struct TextFormatBuilder {
-    factory: DwriteFactory,
-    size: Option<f32>,
-    family: Option<String>,
+/// We hold onto both the wide string and the rust string.
+///
+/// This type AsRefs to both `str` and `[u16]`.
```
raphlinus, line 62: A bit off-topic for review, but this would actually make a cool motivating example for `AsRef` as opposed to `Borrow`, as I know that confuses a lot of people.

piet-direct2d/src/dwrite.rs
```diff
@@ -133,42 +151,114 @@ impl DwriteFactory {
...
-    pub fn size(mut self, size: f32) -> TextFormatBuilder {
-        self.size = Some(size);
-        self
+        family.family_name().ok()
     }
+}
+
+impl FontFamily {
+    // this monster is taken right out of the docs :/
+    /// Returns the localized name of this family.
```
raphlinus, line 181: Getting localized names feels really wrong to me, though I'm willing to accept that it is the best of not-so-good options for identifying a font. I'm curious whether there is other precedent for this (browsers, font-kit).

I did look at the documentation for `GetFamilyNames` and accept that this is a good implementation :)

piet-direct2d/src/dwrite.rs
```diff
@@ -133,42 +151,114 @@ impl DwriteFactory {
...
-        self.family = Some(family);
-        self
+            if !SUCCEEDED(hr) {
+                return Err(hr.into());
+            }
+
+            let mut wide_name: Vec<u16> = Vec::with_capacity(length as usize + 1);
+            let hr = names.GetString(index, wide_name.as_mut_ptr(), length + 1);
+            if SUCCEEDED(hr) {
+                wide_name.set_len(length as usize + 1);
```
raphlinus, line 228: I think maybe we want `set_len(length)` here so we don't include the trailing null.

piet-direct2d/src/dwrite.rs
```diff
@@ -133,42 +151,114 @@ impl DwriteFactory {
...
-            let hr = self.factory.0.CreateTextFormat(
+            let hr = factory.0.CreateTextFormat(
                 family.as_ptr(),
                 null_mut(), // collection
                 DWRITE_FONT_WEIGHT_NORMAL,
                 DWRITE_FONT_STYLE_NORMAL,
                 DWRITE_FONT_STRETCH_NORMAL,
                 size,
-                locale.as_ptr(),
+                //TODO: this should be the user's locale? It will influence font fallback behaviour?
```
raphlinus, line 260: Yes, this should be the user's locale. This is a fairly complex topic, and being able to override the locale on a per-span basis should be a feature at some point, but this is ok for now. The main thing that will be impacted is CJK.

piet-direct2d/src/dwrite.rs
```diff
@@ -178,81 +268,113 @@ impl TextFormatBuilder {
...
+#[allow(overflowing_literals)]
+#[allow(clippy::unreadable_literal)]
+const E_NOT_SUFFICIENT_BUFFER: HRESULT = 0x8007007A;
 
-    pub fn alignment(mut self, alignment: TextAlignment) -> TextLayoutBuilder {
-        self.alignment = alignment;
-        self
+fn make_text_range(start: usize, len: usize) -> DWRITE_TEXT_RANGE {
+    DWRITE_TEXT_RANGE {
+        startPosition: start as u32,
```
raphlinus, line 289: The `as` conversions here really should be `try_into` with panicking unwrap. That said, unlikely to be an issue in practice.

piet-direct2d/src/dwrite.rs
```diff
@@ -178,81 +268,113 @@ impl TextFormatBuilder {
...
+impl TextLayout {
+    pub(crate) fn new(
+        dwrite: &DwriteFactory,
+        format: TextFormat,
+        width: f32,
+        text: &[u16],
+    ) -> Result<Self, Error> {
+        const QUITE_TALL_HEIGHT: f32 = 1e6;
         let len = text.len();
         assert!(len <= 0xffff_ffff);
```
raphlinus, line 303: I probably wrote that assert, but more idiomatic style would be `try_into` instead of `as`.

piet-direct2d/src/text.rs
```diff
@@ -58,32 +81,25 @@ impl Text for D2DText {
...
-                .or_else(|_| {
-                    TextFormatBuilder::new(self.dwrite.clone())
-                        .size(size as f32)
-                        .family("Arial".into())
-                        .build()
-                })
-                .unwrap(),
-        )
+        let collection = self.dwrite.system_font_collection().unwrap();
+        //TODO: this is maybe not the best thing? I _think_ if we pass an empty string
```
raphlinus, line 95: In practice I think we can expect "Segoe UI" to always succeed. At some point, we should test this on Windows 7.

piet-svg/src/text.rs
```diff
@@ -60,11 +63,20 @@ pub struct TextLayoutBuilder;
...
     fn alignment(self, _alignment: piet::TextAlignment) -> Self {
         self
     }
 
+    fn add_attribute(
+        self,
+        _range: impl RangeBounds<usize>,
+        _attribute: impl Into<TextAttribute<Self::Font>>,
+    ) -> Self {
+        self
```
raphlinus, line 77: Should there be at least a TODO here indicating that it's unimplemented functionality?

piet-direct2d/src/dwrite.rs
cmyr: worth it?

piet-direct2d/src/dwrite.rs
cmyr: I've debated taking this out, and instead just using whatever name we've been given if it returns a family; we can trust that name will return a family in the future, and there's no advantage to storing the canonical name (that i know of) and getting it requires us to jump through all of these hoops. Ultimately I decided to just leave this here since it's "more correct" and it already works, but I could go either way.

piet-direct2d/src/dwrite.rs
cmyr: I think the null is part of the 'string', and should be included?

piet-direct2d/src/dwrite.rs
cmyr: hm, how about a saturating cast? `start::try_into::<u32>().unwrap_or(u32::max_value())`?

piet-direct2d/src/dwrite.rs
cmyr: A lot of plumbing there to figure out, since we'll want the locale to be part of the general piet API in this case, likely as an attribute on the layout; going to punt to a future PR.

piet-direct2d/src/dwrite.rs
cmyr: my inclination is to truncate instead of asserting, but agree it's a mostly theoretical concern

piet-direct2d/src/text.rs
cmyr: yea, i'm trying to think a bit more about what we should expect to exist on a given platform. If a user has deleted system fonts, I'm not sure we owe them much special treatment.

piet-svg/src/text.rs
cmyr: the whole svg text API is unimplemented; I don't think you can even get a `TextLayoutBuilder` without panicking.

piet-direct2d/src/dwrite.rs
raphlinus: On second thought, I think you're right. It's probably within the contract for the vec not to preserve the null.

piet-direct2d/src/dwrite.rs
raphlinus: That can, among other things, cut a code point. I prefer panicking. Obviously we will never see a 4GB string for reals, it's invariably either an attack or a logic bug.

piet-direct2d/src/dwrite.rs
raphlinus: If there are a bunch of these, otherwise not.

cmyr: too late I did it 😞

PR #232, opened 2020-07-16
Add TextLayout::size method

cmyr:
This replaces TextLayout::width; it is reasonable that the caller
would want to know the size of the entire layout, for things like
drawing a background or similar.

### DIFF
--- a/piet-direct2d/src/text.rs
+++ b/piet-direct2d/src/text.rs
@@ -5,19 +5,17 @@ mod lines;
 use std::convert::TryInto;
 use std::ops::RangeBounds;
 
-use wio::wide::ToWide;
-
 pub use d2d::{D2DDevice, D2DFactory, DeviceContext as D2DDeviceContext};
 pub use dwrite::DwriteFactory;
+use wio::wide::ToWide;
 
-use piet::kurbo::Point;
+use piet::kurbo::{Point, Size};
 use piet::util;
 use piet::{
     Error, Font, FontBuilder, HitTestMetrics, HitTestPoint, HitTestTextPosition, LineMetric, Text,
     TextAlignment, TextAttribute, TextLayout, TextLayoutBuilder,
 };
 
-use self::lines::fetch_line_metrics;
 use crate::conv;
 use crate::d2d;
 use crate::dwrite::{self, FamilyName, TextFormat};
@@ -44,6 +42,7 @@ pub struct D2DTextLayout {
     pub text: String,
     // currently calculated on build
     line_metrics: Vec<LineMetric>,
+    size: Size,
     pub layout: dwrite::TextLayout,
 }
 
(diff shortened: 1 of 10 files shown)


piet-direct2d/src/text.rs
```diff
@@ -127,18 +128,27 @@ impl TextLayoutBuilder for D2DTextLayoutBuilder {
         let layout = self.builder.build()?;
 
         let line_metrics = fetch_line_metrics(&layout);
+        let text_metrics = layout.get_metrics();
+        //FIXME: why is this including trailing whitespace?
```
raphlinus, line 132: It probably should not. Basically the distinction is that for drawing and layout purposes, don't include trailing whitespace. The one case where including trailing whitespace is likely valid is cursor positioning, but conventions vary because positioning the cursor outside the layout rect can yield poor user experience.

Because both use cases are valid, a case can be made that we should have a metrics structure rather than a simple size. This metrics structure can contain other information such as the baseline of the first line. But I'm happy to consider that future work.

piet-web/src/text.rs
```diff
@@ -138,45 +138,45 @@ impl TextLayoutBuilder for WebTextLayoutBuilder {
...
-        self.width
+        self.size.width
+    }
+
+    fn size(&self) -> Size {
+        self.size
     }
 
-    // TODO refactor this to use same code as build
     fn update_width(&mut self, new_width: impl Into<Option<f64>>) -> Result<(), Error> {
```
raphlinus, line 164: Should this be renamed "update_size"?

piet/src/text.rs
```diff
@@ -102,8 +102,18 @@ impl Default for TextAlignment {
 ///
 pub trait TextLayout: Clone {
     /// Measure the advance width of the text.
+    #[deprecated(since = "0.2.0", note = "Use size().width insead")]
     fn width(&self) -> f64;
 
+    //TODO: ideally this would just be the bounding box of the glyphs? Otherwise
```
raphlinus, line 108: Ultimately we will need *both* a layout size and a bounding box of the glyphs, for different reasons. Using one for the other is a persistent source of paper-cut bugs. I don't think this comment is very helpful as-is, as it seems to conflate the two concerns.

Regarding discrepancies between Core Text and DirectWrite, I think at some point we're going to have to do a careful review of vertical measurement, line space, and padding, something along the lines of [the Figma blog post on line height](https://www.figma.com/blog/line-height-changes/), and systematically document what we expect the behavior to be, as well as implement that behavior. It also needs to take into account scripts which have excessive line height (such as Myanmar). It's not a simple problem, as evidenced by continual wrestling over it.

piet-web/src/text.rs
cmyr: the 'width' here refers to our width *constraint*, not to the width of our final layout (although it will impact that width).

The larger question, perhaps, is whether or not we should allow for both width **and** height constraints.

piet/src/text.rs
cmyr: yea, left this comment in so we could have this discussion, and broadly I agree. I think the end game should be to get close to where figma is; in particular I think agreeing on a common definition of line-height would be really valuable, but that will require us to go to a runs-based API (or a line-based api that is backed by runs on most platforms).

piet-web/src/text.rs
raphlinus: Ah, got it.

PR #234, opened 2020-07-17
Use top-left as text origin

cmyr:
This changes behaviour and will likely break text in druid, but that's future work.

in addition, LineMetric::cumulative_height is replaced with LineMetric::y_offset, which (I think) better communicates the intent of this value, which is to determine the position of the line relative to the origin of the layout.

(draft so I can run CI)

### DIFF
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -420,11 +420,15 @@ impl TextLayout for CoreGraphicsTextLayout {
                 _ => false,
             })
             .count();
-        let height = typo_bounds.ascent + typo_bounds.descent + typo_bounds.leading;
         // this may not be exactly right, but i'm also not sure we ever use this?
         //  see https://stackoverflow.com/questions/5511830/how-does-line-spacing-work-in-core-text-and-why-is-it-different-from-nslayoutm
-        let cumulative_height =
-            (self.line_y_positions[line_number] + typo_bounds.descent + typo_bounds.leading).ceil();
+        let ascent = (typo_bounds.ascent + 0.5).floor();
+        let descent = (typo_bounds.descent + 0.5).floor();
+        let leading = (typo_bounds.leading + 0.5).floor();
+        let height = ascent + descent + leading;
+        let y_offset = self.line_y_positions[line_number] - ascent;
+        let cumulative_height = self.line_y_positions[line_number] + descent + leading;
+        #[allow(deprecated)]
         Some(LineMetric {
             start_offset,
             end_offset,
@@ -432,6 +436,7 @@ impl TextLayout for CoreGraphicsTextLayout {
             baseline: typo_bounds.ascent,
             height,
             cumulative_height,
+            y_offset,
         })
     }
 
@@ -646,7 +651,7 @@ mod tests {
         assert_eq!(line4.trailing_whitespace, 0);
 
         let total_height = layout.frame_size.height;
-        assert_eq!(line4.cumulative_height, total_height);
+        assert_eq!(line4.y_offset + line4.height, total_height);
 
         assert!(layout.line_metric(4).is_none());
     }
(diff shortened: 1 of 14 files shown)


piet/src/text.rs
```diff
@@ -306,17 +306,24 @@ pub struct LineMetric {
...
-    /// Distance of the baseline from the top of the line
+    /// The distance from the top of the line (`y_offset`) to the baseline.
     pub baseline: f64,
 
-    /// Line height
+    /// The height of the line.
     pub height: f64,
 
     /// Cumulative line height (includes previous line heights)
+    #[deprecated(since = "0.2.0", note = "use y_offset instead")]
```
raphlinus, line 319: The main reason we went with cumulative height is that all values provide useful information. By measuring the top, the first value is always 0 (I think; there could be handling of padding that makes this not so), and the height can be recovered by subtracting the value for the previous line (unwrap_or 0). But I'm open to the argument that being able to recover height is not important, we don't care too much about the 8 bytes.

piet/src/text.rs
```diff
@@ -306,17 +306,24 @@ pub struct LineMetric {
...
     pub cumulative_height: f64,
+
+    /// The y position of the top of this line, relative to the top of the layout.
```
raphlinus, line 322: Is it an invariant that y_offset of the next line is equal to y_offset of this line + height? If so, it should be stated. If not, the reason why not should be clarified. (Obviously this depends on the way line gap is handled)

piet-coregraphics/src/text.rs
```diff
@@ -416,18 +416,23 @@ impl TextLayout for CoreGraphicsTextLayout {
...
         // this may not be exactly right, but i'm also not sure we ever use this?
         //  see https://stackoverflow.com/questions/5511830/how-does-line-spacing-work-in-core-text-and-why-is-it-different-from-nslayoutm
-        let cumulative_height =
-            (self.line_y_positions[line_number] + typo_bounds.descent + typo_bounds.leading).ceil();
+        let ascent = (typo_bounds.ascent + 0.5).floor();
+        let descent = (typo_bounds.descent + 0.5).floor();
+        let leading = (typo_bounds.leading + 0.5).floor();
+        let height = ascent + descent + leading;
+        let y_offset = self.line_y_positions[line_number] - ascent;
+        let cumulative_height = self.line_y_positions[line_number] + ascent + leading;
```
raphlinus, line 426: Did you mean descent here? Actually I believe the invariant is that y_offset + height = cumulative_height, but I'm not 100% sure of this.

piet/src/text.rs
cmyr: on macOS, the y_offset of the first line is not 0; I believe that using `y_offset` lets us still recover all values (since y_offest + height == cumulative height) but is more ergonomic in more cases?

piet-coregraphics/src/text.rs
cmyr: yes, good catch.

piet/src/text.rs
cmyr: It's not, although it is true in many cases. There are fonts that this does not hold for, though (again, Courier on mac) and I haven't investigated.

I think making these firm invariants will only be possible when we have finer grained control over layout and line spacing.

piet/src/text.rs
```diff
@@ -306,17 +306,24 @@ pub struct LineMetric {
...
-    /// Length in (in UTF-8 code units) of current line's trailing whitespace.
+    /// The length of the trailing whitespace at the end of this line, in utf-8 code units.
     pub trailing_whitespace: usize,
 
-    /// Distance of the baseline from the top of the line
+    /// The distance from the top of the line (`y_offset`) to the baseline.
     pub baseline: f64,
 
-    /// Line height
+    /// The height of the line.
```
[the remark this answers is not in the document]
cmyr: Agreed, but I don't think we will be able to make this a firm claim until we have total control over one spacing.

The only claim I can make here is that "this height is suitable for things such as drawing cursors and selection regions."

piet/src/text.rs
cmyr: also: the main reason I like this formulation is that it simplifies finding the baseline, which is now just `y_offset + baseline`, instead of `cumulative_height - height + baseline`.

PR #238, opened 2020-07-21
Add TextLayout::image_bounds

cmyr:
This method returns the bounding rectangle of the glyphs in the
layout, relative to the layout's origin.

It also adds a new sample picture (`picture_10.rs`) that shows the
difference between this region and the layout's bounds in an
exaggerated case (zalgo).

coregraphics-test-10

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -7,8 +7,7 @@ use std::ops::RangeBounds;
 
 use cairo::{FontFace, FontOptions, FontSlant, FontWeight, Matrix, ScaledFont};
 
-use piet::kurbo::{Point, Size};
-
+use piet::kurbo::{Point, Rect, Size};
 use piet::{
     Error, Font, FontBuilder, HitTestMetrics, HitTestPoint, HitTestTextPosition, LineMetric,
     RoundInto, Text, TextAttribute, TextLayout, TextLayoutBuilder,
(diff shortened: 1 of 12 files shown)


piet-cairo/src/text.rs
```diff
@@ -158,6 +157,10 @@ impl TextLayout for CairoTextLayout {
         self.size
     }
 
+    fn image_bounds(&self) -> Rect {
```
raphlinus, line 160: Do we want a TODO here that it's approximate? But the entire cairo stack is currently a "toy" so I don't think it matters much.

piet-web/src/text.rs
```diff
@@ -172,6 +172,11 @@ impl TextLayout for WebTextLayout {
         self.size
     }
 
+    fn image_bounds(&self) -> Rect {
+        //FIXME: figure out actual image bounds on web?
```
raphlinus, line 176: Yes, there are methods for this, see https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics (which it might be valid to link). I think the image bounds methods are implemented on all browsers which can reasonably run Rust, but for good measure it's probably best to sense at runtime.

piet-cairo/src/text.rs
cmyr: yea, that was my position.

piet-web/src/text.rs
cmyr: going to leave this in and I'll open an issue.

PR #239, opened 2020-07-21
[d2d] Don't pass infinite width constraint to dwrite

cmyr:
DirectWrite uses the provided width and height in some calculations
(such as in the output of DWRITE_OVERHANG_METRICS) and passing
f32::INFINITY produces nonsense values.

This was motivated by #238, where currently  `TextLayout::image_bounds` returns garbage on windows if no width constraint is provided.

### DIFF
--- a/piet-direct2d/src/dwrite.rs
+++ b/piet-direct2d/src/dwrite.rs
@@ -35,6 +35,13 @@ use crate::Brush;
 /// "en-US" as null-terminated utf16.
 const DEFAULT_LOCALE: &[u16] = utf16_lit::utf16_null!("en-US");
 
+/// The max layout constraint we use with dwrite.
+///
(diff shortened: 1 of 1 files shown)


piet-direct2d/src/dwrite.rs
```diff
@@ -413,6 +425,13 @@ impl TextLayout {
     }
 
     pub fn set_max_width(&mut self, max_width: f64) -> Result<(), Error> {
+        // infinity produces nonsense values for the inking rect on d2d
+        let max_width = if !max_width.is_finite() {
+            MAX_LAYOUT_CONSTRAINT
```
raphlinus, line 430: What happens if we just don't set a max width? In any case, I think this should be fine.

cmyr: The absence of a max width is represented by `f64::INFINITY` at this point.

PR #243, opened 2020-07-22
Make italic/underline text attributes take a boolean

cmyr:
This will be necessary when attributes for specific ranges can
override default attributes.


piet-coregraphics/src/text.rs
```diff
@@ -134,12 +134,19 @@ impl CoreGraphicsTextLayoutBuilder {
                         color.as_CFType(),
                     )
                 }
-                TextAttribute::Underline => {
-                    #[allow(non_upper_case_globals)]
+                #[allow(non_upper_case_globals)]
```
raphlinus, line 137: Do you have to move this to avoid the warnings?

cmyr: previously with a single `const` here I could put this annotation on the individual item, but adding a second `const` requires me to put it on the nearest containing block.

PR #250, opened 2020-07-29
Snapshots

cmyr:
This is a first simple attempt to add snapshot testing to piet.

In this PR this will run as part of normal CI; it's possible we would want a separate CI step for this, maybe one that we run manually? I don't want to get too fancy right now.

This uses a submodule, which feels like the lesser of the various possible evils.


piet-coregraphics/src/lib.rs
```diff
@@ -77,6 +77,10 @@ impl<'a> CoreGraphicsContext<'a> {
             let xform = Affine::FLIP_Y * Affine::translate((0.0, -height));
             ctx.concat_ctm(to_cgaffine(xform));
         }
+        // these two options are the default on 10.15; we set them explicitly to ensure
```
raphlinus, line 80: These don't seem like useful defaults, I think font smoothing should be on.

cmyr: okay, I'll look into this a bit more.

cmyr: okay, I removed this, does seem like it might be a gotcha.

PR #251, opened 2020-07-31
Font API changes

cmyr:
This PR is now a big rollup of font-related changes, including:

- use of a `FontFamily` type as the token you receive when querying a font
- non-failable 'generic font families' (`FontFamily::SERIF`, e.g.)
- removal of `Font` and `FontBuilder` traits
- a bunch of doc improvements around text

This is hopefully the last major API change; there are still a few API additions to make, but hopefully nothing earth-shattering.

### DIFF
--- a/piet/src/text.rs
+++ b/piet/src/text.rs
@@ -10,34 +10,14 @@ pub trait Text: Clone {
     type TextLayoutBuilder: TextLayoutBuilder<Out = Self::TextLayout>;
     type TextLayout: TextLayout;
 
-    /// Returns a a [`Font`] object corresponding to the given family name,
-    /// if one can be found.
+    /// Query the platform for a font with a given name, and return a `FontFamily`
+    /// object corresponding to that font, if it is found.
     ///
-    /// The family name should be either the explicit name of a font family
-    /// (such as "Times New Roman", or "Helvetica") or it can be a "generic
-    /// family name"; see below.
-    ///
-    /// ## Platform behaviour:
-    ///
-    /// Different platforms may search for fonts in different ways. On macOS,
-    /// for instance, passing "courier" will (assuming "Courier" is not installed)
-    /// will return "Courier New". on Windows, in the same situation, it will return
-    /// `None`. You should assume that matching is literal, and you should try
-    /// to use known literal names.
-    ///
-    /// ## Generic family names
-    ///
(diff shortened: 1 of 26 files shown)


piet/src/text.rs
```diff
@@ -3,25 +3,33 @@
...
+    /// ```
+    /// # use piet::*;
+    /// # let mut ctx = NullRenderContext::new();
+    /// # let text = ctx.text();
+    /// let text_font = text.font("Charter")
+    ///     .or_else(|| text.font("Garamond"))
+    ///     .unwrap_or(FontFamily::SERIF);
+    /// ```
+    ///
+    /// [`Font`]: trait.Font.html
```
cmyr, line 31: this doesn't exist any more

piet/src/text.rs
```diff
@@ -30,15 +38,67 @@ pub trait Text: Clone {
...
+///
+/// [`Text::font`]: trait.Text.html#tymethod.font
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FontFamily {
+    inner: ArcOrStaticStr,
+}
+
+impl FontFamily {
+    /// A san-serif font, such as Arial or Helvetica.
+    pub const SANS_SERIF: FontFamily = FontFamily::new_const("sans-serif");
```
raphlinus, line 62: I think it would be better for these to be enums.

piet/src/text.rs
```diff
@@ -30,15 +36,76 @@ pub trait Text: Clone {
...
+/// [`Text::font`]: trait.Text.html#tymethod.font
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FontFamily(FontFamilyInner);
+
+/// The inner representation of a font family.
+///
+/// This is not public API for users of piet; it is exposed for backends only.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[doc(hidden)]
+#[non_exhaustive]
```
raphlinus, line 61: I'm not sure non_exhaustive is appropriate here; the back-end should probably handle all cases. But since a default fallback font is reasonable, I'm ok with it.

cmyr: I think my reasoning was that since this is technically public api, we wanted to indicate that more variants might be added. 🤷

PR #254, opened 2020-08-04
Add TextLayout::text and TextLayout::rects_for_range

cmyr:
The TextLayout::text was added because we need access to the
length of the text in order to provide a nice default implementation
for TextLayout::rects_for_range.

This also includes a new sample picture, picture_12, which
illustrates drawing selections and highlights.

coregraphics-test-12

### DIFF
--- a/piet/src/text.rs
+++ b/piet/src/text.rs
@@ -358,6 +358,9 @@ pub trait TextLayout: Clone {
     /// This is sometimes called the bounding box or the inking rect.
     fn image_bounds(&self) -> Rect;
 
+    /// The text used to create this layout.
+    fn text(&self) -> &str;
+
     /// Change the width of this `TextLayout`.
     ///
     /// This may be an `f64`, or `None` if this layout is not constrained;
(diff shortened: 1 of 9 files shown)


piet/src/text.rs
```diff
@@ -427,6 +430,47 @@ pub trait TextLayout: Clone {
...
     fn hit_test_text_position(&self, idx: usize) -> Option<HitTestPosition>;
+
+    /// Returns a vector of `Rect`s that cover the region of the text indicated
+    /// by `range`.
+    ///
+    /// The returned rectangles are suitable for things like drawing selection
+    /// regions or highlights.
+    ///
+    /// `range` will be clamped to the length of the text if necessary.
+    fn rects_for_range(&self, range: impl RangeBounds<usize>) -> Vec<Rect> {
```
raphlinus, line 441: Might be worth adding a comment that this default implementation is not BiDi capable, and that it will be removed when BiDi becomes a requirement.

PR #255, opened 2020-08-05
Font loading

cmyr:
This implements the minimum possible version of custom fonts.

- fonts can be loaded, but not unloaded
- loaded fonts are referred to by their family name

### DIFF
--- a/piet-direct2d/Cargo.toml
+++ b/piet-direct2d/Cargo.toml
@@ -14,12 +14,9 @@ piet = { version = "0.2.0", path = "../piet" }
 utf16_lit = "1.0"
 
 wio = "0.2.2"
-
-[dependencies.winapi]
-version = "0.3.8"
-features = ["d2d1", "d2d1_1", "d2d1effects", "d3d11", "dxgi", "winnls"]
+winapi = { version = "0.3.8", features = ["d2d1", "d2d1_1", "d2d1effects", "d3d11", "dxgi", "winnls"] }
+dwrote = { version = "0.11.0", default_features = false }
 
 [dev-dependencies]
 piet = { version = "0.2.0", path = "../piet", features = ["samples"] }
-
 image = "0.23.2"
--- a/piet-direct2d/src/text.rs
+++ b/piet-direct2d/src/text.rs
@@ -2,11 +2,15 @@
 
 mod lines;
 
+use std::cell::RefCell;
 use std::convert::TryInto;
 use std::ops::{Range, RangeBounds};
+use std::rc::Rc;
+use std::sync::Arc;
 
 pub use d2d::{D2DDevice, D2DFactory, DeviceContext as D2DDeviceContext};
 pub use dwrite::DwriteFactory;
+use dwrote::{CustomFontCollectionLoaderImpl, FontCollection, FontFile};
 use wio::wide::ToWide;
 
 use piet::kurbo::{Insets, Point, Rect, Size};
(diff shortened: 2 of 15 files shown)


piet-direct2d/Cargo.toml
```diff
@@ -14,12 +14,9 @@ piet = { version = "0.2.0", path = "../piet" }
 utf16_lit = "1.0"
 
 wio = "0.2.2"
-
-[dependencies.winapi]
-version = "0.3.8"
-features = ["d2d1", "d2d1_1", "d2d1effects", "d3d11", "dxgi", "winnls"]
+winapi = { version = "0.3.8", features = ["d2d1", "d2d1_1", "d2d1effects", "d3d11", "dxgi", "winnls"] }
```
raphlinus, line 17: Time to upgrade to 0.3.9 while we're doing this?

piet-web/src/text.rs
```diff
@@ -63,6 +63,10 @@ impl Text for WebText {
         Some(FontFamily::new_unchecked(family_name))
     }
 
+    fn load_font(&mut self, _data: &[u8]) -> Result<FontFamily, Error> {
```
raphlinus, line 66: I think this could be added, though, at considerable difficulty? Probably deserves some thought, maybe an issue.

piet-direct2d/src/text.rs
```diff
@@ -24,6 +28,17 @@ use crate::dwrite::{self, TextFormat};
 pub struct D2DText {
     dwrite: DwriteFactory,
     device: d2d::DeviceContext,
+    loaded_fonts: Rc<RefCell<LoadedFonts>>,
```
raphlinus, line 31: I have a feeling we might want to change this to Arc at some point, to allow resource loading across threads, but there's probably a bunch of stuff that's going to change to go fully multithreaded, and the compiler will tell us what to do.

piet/src/text.rs
```diff
@@ -29,6 +28,59 @@ pub trait Text: Clone {
...
+    /// attempts to load it, making it subsequently available for use.
+    ///
+    /// If loading is successful, this method will return a [`FontFamily`] handle
+    /// that can be used to select this font when constructing a [`TextLayout`].
+    ///
+    /// # Notes
+    ///
+    /// ## font familes and styles:
+    ///
+    /// If you wish to use multiple fonts in a given family, you will need to
```
raphlinus, line 43: So this makes me a bit uncomfortable. The *right* thing to do is load multiple styles in a single family, but I understand that's a lot harder. A note that we plan to change the API would be sufficient for this cycle.

piet-direct2d/Cargo.toml
```diff
@@ -14,12 +14,9 @@ piet = { version = "0.2.0", path = "../piet" }
...
+dwrote = "0.11.0"
```
raphlinus, line 18: I'm not happy with a dependency on dwrote, at least at the moment, as it has safety problems. Also we have a patchwork where we use different crates for different things. I appreciate the fact that using it is a lot less code, so am not going to block approval, but wonder what's the best way forward. One is to have an issue to remove the dwrote dep (by duplicating the functionality). Another possibility that occurs to me is to file an issue on dwrote, and see what happens.

piet-direct2d/Cargo.toml
cmyr: will do that as a separate patch.

piet-direct2d/src/text.rs
cmyr: I'm specifically unsure about the thread safety of the contained types, and wanted to be defensive for the time being.

piet-web/src/text.rs
cmyr: yes certainly *likely* possible; I'm not treating web as a 'tier 1' target for the purposes of this current work though.

piet/src/text.rs
cmyr: maybe my wording is bad. A single file that contains multiple styles *will* work; you just can't address the individually loaded styles using the handle you get back, you need to resolve the style by setting italic/weight as appropriate.

One thing that we don't make easy, here, is accessing a condensed style, but I'm okay with that for now.

piet-direct2d/Cargo.toml
cmyr: Agree this isn't the ideal solution; it's totally a matter of expediency. I'm happy to open an issue to remove this dep once com-rs has matured a little bit.

raphlinus: One other thing about this: I think dwrote pulls in serde by default, so better to turn off default features.

PR #258, opened 2020-08-11
Support wght axis when present on a variable font

cmyr:
We don't expose fine grained control for variable fonts, but if
a font has a weight axis, we will use it.

### DIFF
--- a/piet-coregraphics/src/ct_helpers.rs
+++ b/piet-coregraphics/src/ct_helpers.rs
@@ -388,6 +388,12 @@ impl FontCollection {
 extern "C" {
     static kCTFontFamilyNameKey: CFStringRef;
 
+    pub static kCTFontVariationAxisIdentifierKey: CFStringRef;
+    //static kCTFontVariationAxisMinimumValueKey: CFStringRef;
+    //static kCTFontVariationAxisMaximumValueKey: CFStringRef;
+    //static kCTFontVariationAxisDefaultValueKey: CFStringRef;
+    //pub static kCTFontVariationAxisNameKey: CFStringRef;
+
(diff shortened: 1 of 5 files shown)


piet-coregraphics/src/ct_helpers.rs
```diff
@@ -388,6 +388,12 @@ impl FontCollection {
 extern "C" {
     static kCTFontFamilyNameKey: CFStringRef;
 
+    pub static kCTFontVariationAxisIdentifierKey: CFStringRef;
+    //static kCTFontVariationAxisMinimumValueKey: CFStringRef;
```
raphlinus, line 392: Generally commented out code is something we would want to avoid, but in this particular case I see the value - these are things we may want to add later.

piet-coregraphics/src/text.rs
```diff
@@ -220,6 +220,10 @@ impl CoreGraphicsTextLayoutBuilder {
     fn current_font(&self) -> CTFont {
         //TODO: this is where caching would happen, if we were implementing caching;
         //store a tuple of attributes resolved to a generated CTFont.
+
+        // 'wght' as an int
+        const WEIGHT_AXIS_ID: i32 = 2003265652;
```
raphlinus, line 225: This would be easier to read in hex (0x77676874), but even easier to read as `tag("wght")`, which is now viable due to const fn improvements:

```rust
const fn tag(s: &str) -> u32 {
    let b = s.as_bytes();
    ((b[0] as u32) << 24) | ((b[1] as u32) << 16) | ((b[2] as u32) << 8) | (b[3] as u32)
}
```

PR #259, opened 2020-08-12
[cg] Fallback to synthetic italic when italic face is missing

### LINKED ISSUE #257: support synthetic oblique
This is implemented by the platform on windows, but not on macOS.

If there is no italic font available for the current family, and italic has been requested, we should use an affine to slant it.

cmyr:
This will better match user expectations, and matches the behaviour
on windows.

closes #257

### DIFF
--- a/piet-coregraphics/src/ct_helpers.rs
+++ b/piet-coregraphics/src/ct_helpers.rs
@@ -21,7 +21,7 @@ use core_graphics::{
     color::CGColor,
     data_provider::CGDataProvider,
     font::CGFont,
-    geometry::{CGPoint, CGRect, CGSize},
+    geometry::{CGAffineTransform, CGPoint, CGRect, CGSize},
     path::CGPathRef,
 };
 use core_text::{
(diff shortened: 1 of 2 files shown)


piet-coregraphics/src/ct_helpers.rs
```diff
@@ -384,6 +384,23 @@ impl FontCollection {
...
+pub(crate) fn make_font(desc: &CTFontDescriptor, pt_size: f64, affine: Affine) -> CTFont {
+    let [a, b, c, d, e, f] = affine.as_coeffs();
+    let affine = CGAffineTransform::new(a, b, c, d, e, f);
+    unsafe {
+        let font_ref = CTFontCreateWithFontDescriptor(
+            desc.as_concrete_TypeRef(),
+            pt_size as CGFloat,
+            // ownership here isn't documented, which I presume to mean it is copied,
+            // and we can pass a stack pointer.
+            &affine as *const _,
```
raphlinus, line 398: You don't need the `as` here, it should automatically convert the reference into the appropriate pointer.

piet-coregraphics/src/text.rs
```diff
@@ -223,6 +228,9 @@ impl CoreGraphicsTextLayoutBuilder {
 
         // 'wght' as an int
         const WEIGHT_AXIS_ID: i32 = 2003265652;
+        // taken from android:
+        // https://api.skia.org/classSkFont.html#aa85258b584e9c693d54a8624e0fe1a15
+        const SLANT_RADIANS: f64 = 0.25;
```
raphlinus, line 233: I would name this "SLANT_TANGENT", as it's the tan of the actual angle (~0.245 radians = ~14°). It's a *very* small difference, but I think being precise can avoid confusion.

PR #262, opened 2020-08-13
Standardize behaviour of TextLayout::hit_test_text_position

cmyr:
We will now always clamp an out of bounds position to the end of
the text, and we will always panic if the text position is not
a character boundary.

### DIFF
--- a/piet-direct2d/src/text.rs
+++ b/piet-direct2d/src/text.rs
@@ -305,22 +305,19 @@ impl TextLayout for D2DTextLayout {
     }
 
     // Can panic if text position is not at a code point boundary, or if it's out of bounds.
-    fn hit_test_text_position(&self, text_position: usize) -> Option<HitTestPosition> {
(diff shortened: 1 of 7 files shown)


piet-direct2d/src/text.rs
```diff
@@ -305,25 +305,23 @@ impl TextLayout for D2DTextLayout {
...
         // Now convert the utf8 index to utf16.
         // This can panic;
-        let idx_16 = util::count_utf16(&self.text[0..text_position]);
-        let line = util::line_number_for_position(&self.line_metrics, text_position);
-        // panic or Result are also fine options for dealing with overflow. Using Option here
-        // because it's already present and convenient.
-        // TODO this should probably go before convertin to utf16, since that's relatively slow
-        let idx_16 = idx_16.try_into().ok()?;
-
-        let trailing = false;
```
raphlinus, line 323: Very minor style nit, but I feel it's useful for this to continue to be a var; a bool by itself (as opposed to a two-state enum value) feels like it should have some documentation in the code.

PR #269, opened 2020-08-19
Compute default line height for empty layout

### LINKED ISSUE #268: decide on behaviour when hit testing empty string
Ran into this when attempting to update druid: there is currently no good way to determine the cursor position and size when there is no text; and more than that, several platforms currently crash if you call `hit_test_text_position` with position `0` when there is no text.

I can imagine a few approaches:

- we could guarantee that there always exists at least one "line" in a text layout, even if that line is empty; and if required we could fabricate the `LineMetric` object for that line, either by manually resolving the font and using the font metrics, or by having a dummy layout object containing a single space (or something) and using that for the metrics of the empty line.
- we could change `HitTestTextPosition` to include information required to draw the cursor, such as the line height; in the general (non-empty string) case we would just take this from the appropriate `LineMetric`, but in the empty string case we would take it from the font.

I think my preference is for the first of these? I'm trying to think through whether this will cause any obvious problems, and i don't _think_ it does..

cmyr:
This attempts to address the issue of how to determine the correct
cursor size in an empty layout.

Essentially: when constructing the layout, we compute the line height
and baseline values that would be used for the layout's default
font and font size, and we stash these; if the layout is empty, we
use these values to synthesize the result of calls to hit_test_position
and line_metric.

- closes #268

### DIFF
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -60,6 +60,9 @@ pub struct CoreGraphicsTextLayout {
     pub(crate) frame_size: Size,
     image_bounds: Rect,
     width_constraint: f64,
+    // these two are stored values we use to determine cursor extents when the layout is empty.
+    default_baseline: f64,
+    default_line_height: f64,
 }
 
 /// Building text layouts for `CoreGraphics`.
@@ -75,6 +78,8 @@ pub struct CoreGraphicsTextLayoutBuilder {
     /// this happens either when the first range attribute is added, or when
     /// we build the string.
     has_set_default_attrs: bool,
+    default_baseline: f64,
+    default_line_height: f64,
     attrs: Attributes,
 }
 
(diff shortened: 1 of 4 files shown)


piet-coregraphics/src/text.rs
```diff
@@ -149,6 +154,11 @@ impl CoreGraphicsTextLayoutBuilder {
         self.has_set_default_attrs = true;
         let whole_range = self.attr_string.range();
         let font = self.current_font();
+        let ascent = (font.ascent() + 0.5).floor();
```
raphlinus, line 157: Any particular reason these aren't just `.round()`?

Also... I'm more than a little worried about this. Are they rounding to px values ("points" in Apple-speak) or device pixels? If the former, that's too coarse when dpi scaling is more than 1.

piet/src/text.rs
```diff
@@ -438,7 +438,14 @@ pub trait TextLayout: Clone {
     /// Given a line number, return a reference to that line's underlying string.
     fn line_text(&self, line_number: usize) -> Option<&str>;
 
-    /// Given a line number, return a reference to that line's metrics.
+    /// Given a line number, return a reference to that line's metrics, if the line exists.
+    ///
+    /// If this layout's text is the empty string, calling this method with `0`
+    /// return some [`LineMetric`]; this will use the layout's default font to
```
raphlinus, line 444: "returns", I think.

piet-coregraphics/src/text.rs
cmyr: I am following a convention for how apple calculates line height, based on this:
https://stackoverflow.com/questions/5511830/how-does-line-spacing-work-in-core-text-and-why-is-it-different-from-nslayoutm

raphlinus: What a rabbithole indeed. I think that convention is wrong, but we're right to follow it :)

PR #280, opened 2020-08-27
Additional cleanup of the text docs

cmyr:


piet/src/text.rs
```diff
@@ -565,7 +555,14 @@ pub struct LineMetric {
...
 
-    /// The length of the trailing whitespace at the end of this line, in utf-8 code units.
+    /// The length of the trailing whitespace at the end of this line, in utf-8
+    /// code units.
+    ///
+    /// Lines are broken at the grapheme boundary after whitespace, but that
+    /// whitespace is not necessarily rendered since it's just the trailing
+    /// whitespace at the end of a line. Keeping the trailing whitespace data
+    /// available allows API consumers to determine their own trailing
+    /// whitespace strategy.
```
raphlinus, line 565:
```suggestion
    /// Lines are broken at the grapheme boundary after whitespace, but that
    /// whitespace is not necessarily rendered since it's just the trailing
    /// whitespace at the end of a line. Keeping the trailing whitespace data
    /// available allows API consumers to determine their own trailing
    /// whitespace strategy.
```

When lines are broken on whitespace (as is common), the whitespace is assigned to the end of the preceding line. Reporting the size of the trailing whitespace section lets an API consumer measure and render only the trimmed line up to the whitespace.

PR #284, opened 2020-08-28
Move FontFamily and FontWeight to new fonts module

cmyr:
I'd like to try adding FontStyle instead of a bool for italics,
which motivated a bit of cleanup.


cmyr:
yes, no functional change.

PR #285, opened 2020-08-28
Add FontStyle enum

cmyr:
based off of #284.

This was motivated by wondering how I would go about storing something like font style in the druid environment, or stash it in other places, and convincing myself that this does make more sense as an independent type.

### DIFF
--- a/piet/src/text.rs
+++ b/piet/src/text.rs
@@ -3,7 +3,7 @@
 use std::ops::{Range, RangeBounds};
 
 use crate::kurbo::{Point, Rect, Size};
-use crate::{Color, Error, FontFamily, FontWeight};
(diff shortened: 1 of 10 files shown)


piet/src/text.rs
```diff
@@ -111,8 +111,8 @@ pub enum TextAttribute {
     Weight(FontWeight),
     /// The foreground color of the text.
     ForegroundColor(crate::Color),
-    /// Italics.
-    Italic(bool),
+    /// The [`FontStyle`](enum.FontStyle.html).
```
raphlinus, line 114: A classic boilerplate comment. I'd probably add "; regular or italic." This will help people who are specifically searching for how to apply italics imho.

PR #286, opened 2020-08-28
Add FontDescriptor

cmyr:
This is motivated by druid; it would be nice if we could store
a single object that represented not only the font family but
also size, weight, and style.

This is sort of speculative; I was looking at adding font stuff to the druid `Env` and realized it would be annoying to store weight/italic/family all separately.

It might make sense for this to live in druid for now, and then get moved here later? I could honestly go either way.

### DIFF
--- a/piet/src/font.rs
+++ b/piet/src/font.rs
@@ -55,6 +55,22 @@ pub enum FontStyle {
     Italic,
 }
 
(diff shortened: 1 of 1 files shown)


cmyr:
I think I'll experiment with this in druid for a bit before making any big decisions.

PR #291, opened 2020-09-02
Impl PartialEq for Color, Hash for FontFamily

### LINKED ISSUE #290: Implement PartialEq/Hash for Color?
Hash is debatable, since we might want to to allow future representations of colors to contain floats, which aren't hashable? PartialEq just feels like an oversight though.

Also: how about `Eq`?

cmyr:
closes  #290


cmyr:
@raphlinus I see someone is in a good mood 🤨

PR #294, opened 2020-09-04
TextLayout::size should use line height with empty string

### LINKED ISSUE #292: TextLayout::size should use default line height when layout is empty
Currently an empty layout always has size `(0, 0)`, but this makes it difficult to do things like size text boxes appropriately. We already store the line height of the default font for cursor drawing, and we should use this as the height of the layout when it is empty.

cmyr:
- closes #292

This also fixes a bug in the d2d backend where we were not correctly recomputing various metrics when the width of the layout changed.


raphlinus (approved):
Yes, these look like good improvements, thanks!

PR #297, opened 2020-09-04
Make new_text_layout take impl Into<Arc<str>>

### LINKED ISSUE #295: TextLayout objects should store text as Arc<str> instead of String
This makes layout objects cheaper to clone, and makes it possible to avoid duplicating the text data between piet and its users.

cmyr:
This will still allow the simple case of passing a string slice, but
will also let druid share string objects with piet, which should enable
better APIs that aren't giving up efficiency.

Another benefit of this work is making cloning of layout objects
cheaper.

closes #295

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -4,6 +4,7 @@ mod grapheme;
 mod lines;
 
 use std::ops::RangeBounds;
+use std::sync::Arc;
 
 use cairo::{FontFace, FontOptions, FontSlant, FontWeight, Matrix, ScaledFont};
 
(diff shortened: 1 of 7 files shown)


cmyr:
To explain the rationale here a bit better; besides just cheaper cloning, this fixes an issue that I ran into in https://github.com/linebender/druid/pull/1182. Because we create the layout object lazily in druid, if we don't want to hold onto an additional copy of the text we need to have the text passed in whenever the layout object is rebuilt, which doesn't feel exactly right.

raphlinus (approved):
Ok, this change is fine, but as discussed in chat this is bringing up some complex issues around cloning and mutability in the existing design - doing a mutation after a clone will affect the original copy. Some deeper thought is needed whether it's even worthwhile having a mutable width change, and if so how to keep it both safe and efficient. Colin will file an issue, and we can discuss there.

cmyr:
Okay, going to try this, I think it lets me write a much better API in druid.

PR #299, opened 2020-09-05
Add support for strikethrough formatting on text.

smmalis37:
Made by basically just copying underline. Just like underline this does not have any support for formatting the strikethrough itself, although the underlying apis support it. ~~I do not have access to a mac to test the coregraphics implementation (nor have I ever written mac code before)~~, however the d2d implementation appears to work on my Win 10 box. This will need a snapshots update.

### DIFF
--- a/piet/src/text.rs
+++ b/piet/src/text.rs
@@ -123,6 +123,8 @@ pub enum TextAttribute {
     Style(FontStyle),
     /// Underline.
     Underline(bool),
+    /// Strikethrough
+    Strikethrough(bool),
 }
 
 /// A trait for laying out text.
--- a/piet/src/text.rs
+++ b/piet/src/text.rs
@@ -123,7 +123,7 @@ pub enum TextAttribute {
     Style(FontStyle),
     /// Underline.
     Underline(bool),
-    /// Strikethrough
+    /// Strikethrough.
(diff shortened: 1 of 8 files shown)


smmalis37:
Whoops, I thought one of the files I was editing was piets, its actually core-graphics (vscode just lets you edit source of downloaded crates, fun). So that implementation won't work at the moment. Investigating what's needed.

smmalis37:
Blocked on https://github.com/servo/core-foundation-rs/pull/409

smmalis37:
From Zulip:
> Raph Levien: The best path forward might be to stub out the Apple implementation and open an issue to add it.

I'm fine with that if the maintainers are.

piet/src/text.rs
```diff
@@ -123,6 +123,8 @@ pub enum TextAttribute {
     Style(FontStyle),
     /// Underline.
     Underline(bool),
+    /// Strikethrough
```
cmyr, line 126:
```suggestion
    /// Strikethrough.
```

cmyr (approved):
Okay, I'm happy to have this go in as-is; at some point I'd like to think more about a better API for "text decorations" but this is totally reasonable for now.

We'll need an issue that this needs implementation on macOS and cairo.

PR #300, opened 2020-09-06
Add linear interpolation for colors.

derekdreery:
This is something that I would find useful, so I'm contributing it as a suggestion.


derekdreery:
I'm gonna close this for now. Interpolation is a larger topic.

PR #305, opened 2020-09-12
Cache solid color brushes in d2d.

### LINKED ISSUE #303: Repeated create_solid_color calls in D2DTextLayout::draw are expensive and unnescessary
https://github.com/linebender/piet/blob/c1d182c31bedaa71694fc8a189cdb7e1bf357e22/piet-direct2d/src/text.rs#L403

The call to create_solid_color on that line gets expensive when you have 700 different labels all painting and recreating the brush every time. It could probably be cached.

### LINKED ISSUE #304: d2d::create_solid_color is expensive to call, but results could likely be cached
Some profiling turned up repeated calls to d2d::create_solid_color (or rather the call to CreateSolidColorBrush inside it) as consuming lots of CPU time. If we assume the user is likely to be using the same colors over and over again (which seems likely to me) then some sort of internal cache for the brushes would likely improve things noticeably.

smmalis37:
Fixes #303, fixes #304.

Doing this required a few weird changes: removing Clone on DeviceContext and adding it to Error, but it all works out. Apparently nothing was using the Clone on DeviceContext? Testing with https://github.com/smmalis37/sudoku_rust shows a massive responsiveness improvement.

Design questions:
Currently this is caching the full result of the CreateSolidBrush call. This means if the call fails we won't retry it until that brush leaves the cache. Only caching successful brushes would require doing something to work around wrap, so I went with this approach instead. Good enough or should I try to only cache successes?

Cache capacity. It supports powers of 2 up to 8192, I picked 64 arbitrarily.

Replacement strategy. Round robin is the simplest so I went with it. Lru is possible but requires some additional code to insert timestamps.

Cache storage. It has to live somewhere, so I just added it to the DeviceContext. If DeviceContext must be Clone then it would have to move somewhere else (the cache doesn't impl Clone for some reason).

### DIFF
--- a/piet-direct2d/src/d2d.rs
+++ b/piet-direct2d/src/d2d.rs
@@ -12,7 +12,7 @@ use std::marker::PhantomData;
 use std::ops::Deref;
 use std::ptr::{null, null_mut};
 
-use associative_cache::{AssociativeCache, Capacity64, HashDirectMapped, RoundRobinReplacement};
+use associative_cache::{AssociativeCache, Capacity1024, HashDirectMapped, RoundRobinReplacement};
 
 use wio::com::ComPtr;
 
@@ -83,7 +83,7 @@ pub struct DeviceContext(
     AssociativeCache<
         ColorWrapper,
         Result<Brush, Error>,
-        Capacity64,
+        Capacity1024,
         HashDirectMapped,
         RoundRobinReplacement,
     >,
--- a/piet-direct2d/src/d2d.rs
+++ b/piet-direct2d/src/d2d.rs
@@ -12,8 +12,6 @@ use std::marker::PhantomData;
 use std::ops::Deref;
 use std::ptr::{null, null_mut};
 
-use associative_cache::{AssociativeCache, Capacity1024, HashDirectMapped, RoundRobinReplacement};
-
 use wio::com::ComPtr;
 
 use winapi::shared::dxgi::{IDXGIDevice, IDXGISurface};
@@ -51,7 +49,6 @@ pub enum FillRule {
     NonZero,
 }
 
-#[derive(Clone)]
 pub enum Error {
     WinapiError(HRESULT),
 }
@@ -78,16 +75,8 @@ unsafe impl Send for D2DDevice {}
 ///
 /// This struct is public only to use for system integration in piet_common and druid-shell. It is not intended
 /// that end-users directly use this struct.
-pub struct DeviceContext(
-    ComPtr<ID2D1DeviceContext>,
-    AssociativeCache<
-        ColorWrapper,
-        Result<Brush, Error>,
-        Capacity1024,
-        HashDirectMapped,
-        RoundRobinReplacement,
-    >,
-);
+#[derive(Clone)]
+pub struct DeviceContext(ComPtr<ID2D1DeviceContext>);
 
 pub struct PathGeometry(ComPtr<ID2D1PathGeometry>);
 
@@ -119,10 +108,6 @@ pub struct Effect(ComPtr<ID2D1Effect>);
 // Note: there may be an opportunity here to combine with the version in
 // piet-common direct2d_back, but the use cases are somewhat different.
 pub struct BitmapRenderTarget(ComPtr<ID2D1BitmapRenderTarget>);
-
-// A wrapper that exists solely to implement Hash and PartialEq over
-// color values so we can cache brushes.
-struct ColorWrapper(D2D1_COLOR_F);
 
 impl From<HRESULT> for Error {
     fn from(hr: HRESULT) -> Error {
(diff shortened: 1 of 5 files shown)


smmalis37:
Also the case in #303 could be made to cache locally in addition to this, but I don't think it really helps much.

piet-direct2d/src/d2d.rs
```diff
@@ -75,8 +78,16 @@ unsafe impl Send for D2DDevice {}
 ///
 /// This struct is public only to use for system integration in piet_common and druid-shell. It is not intended
 /// that end-users directly use this struct.
-#[derive(Clone)]
-pub struct DeviceContext(ComPtr<ID2D1DeviceContext>);
+pub struct DeviceContext(
+    ComPtr<ID2D1DeviceContext>,
+    AssociativeCache<
```
raphlinus, line 83: This doesn't feel like the right place to store the cache. `DeviceContext` is intended to be a thin wrapper over the raw winapi object, and exists for integration. I had in mind the field would be added to D2DRenderTarget. Is there a strong reason why that doesn't work?

The question of mutability and cloning is a hard one. I'm open to the idea of removing Clone here, but it exists for a reason: to at some point allow background threads to create expensive resources such as images.

If we do end up with this being extended, then it should be a struct with named fields rather than a tuple.

piet-direct2d/src/d2d.rs
```diff
@@ -109,6 +120,10 @@ pub struct Effect(ComPtr<ID2D1Effect>);
 // piet-common direct2d_back, but the use cases are somewhat different.
 pub struct BitmapRenderTarget(ComPtr<ID2D1BitmapRenderTarget>);
 
+// A wrapper that exists solely to implement Hash and PartialEq over
```
raphlinus, line 123: This might go easier if you add the method to D2DRenderContext (instead of DeviceContext) and use 32 bit RGBA values (instead of floats) as the basis for the cache key.

raphlinus (commented):
Thanks for this! I like that the cache logic itself reads smoothly and of course that it results in such a significant performance improvement.

As the comments suggest, I think this will be cleaner in a number of ways if the logic is moved into D2DRenderContext and away from DeviceContext.

I wouldn't worry too much about handling errors. In practice, I think brush creation failure won't happen, and the process will very likely crash soon anyway if and when it does. Just doing panic is probably acceptable, but a straightforward approach as you've done is probably more principled.

piet-direct2d/src/d2d.rs
smmalis37: The reason I put it here was because there are 3 different places where `create_solid_color` is called. If the cache were moved up a level into `D2DRenderContext` then the 2 calls in `D2DTextLayout` would not benefit from the cache. I could also look into refactoring those to call through the render context instead, but I have no idea how involved of a change that would be. I agree putting it here doesn't feel right either. Thoughts?

raphlinus: Urgh. This is indeed a sticky question. Colors are already a place where the otherwise fairly clean abstraction boundary between text and drawing breaks down. I've taken another look at that code. Probably the cleanest thing to do is to change `D2DTextLayout::draw` to take a reference to D2DRenderTarget instead of DeviceContext, but that might create other problems. If so, a targeted approach would be to give that method a (mutable) reference to the color cache as well.

If this is too deep water, then I'd happily accept a PR that just makes the change for drawing, and consider the text stuff a followup, as it's definitely messier.

smmalis37:
Think I've got a refactoring that all works out nicely. One Comcast decides
to bring my internet connection back up I'll push it.

piet/src/color.rs
```diff
@@ -7,7 +7,7 @@ use std::fmt::{Debug, Formatter};
 /// Currently this is only a 32 bit RGBA value, but it will likely
 /// extend to some form of wide-gamut colorspace, and in the meantime
 /// is useful for giving programs proper type.
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, Hash)]
```
raphlinus, line 10: This is sticky. The reason we don't implement Hash here is that future extensions of Color might include floating point values. I'm wrestling with the question because of course there's no compatibility guarantee, and we can just take away Hash if and when we get there, or alternatively decide that we can hash the bit patterns.

But it would ease my mind slightly to use the as_rgba_u32 value instead as your hash key.

piet-direct2d/src/lib.rs
```diff
@@ -14,6 +14,8 @@ mod text;
 use std::borrow::Cow;
 use std::ops::Deref;
 
+use associative_cache::{AssociativeCache, Capacity1024, HashDirectMapped, RoundRobinReplacement};
```
raphlinus, line 17: Oh, and something like HashFourWay is going to be a better choice here, otherwise if two colors hash to the same value it's going to bounce.

raphlinus:
Other than this, it looks nice, thanks!

smmalis37:
Yeah, the refactoring to pipe everything through D2DRenderContext ended up simpler than I was worried about, I'm much happier with how it looks now too.

raphlinus (approved):
Sweet, this looks fine to me now, thanks for taking my feedback. I'll merge this but also give you commit access for any future work.

PR #306, opened 2020-09-12
Remove TextLayout::update_width

### LINKED ISSUE #298: TextLayout::update_width buggy by design
The way `TextLayout::update_width` works means that it can result in invalid layout objects when a layout has been cloned.

When this method is called, it calls a mutating method on the platform layout object, and updates state (such as line metrics) on the piet object; however the platform layout object is shared between all clones of a layout object, whereas other state is not.

We need to think more about how we want to address this; it may make sense to just remove `update_width` for the time being.

cmyr:
See https://github.com/linebender/piet/issues/298

closes #298

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -170,27 +170,6 @@ impl TextLayout for CairoTextLayout {
         &self.text
     }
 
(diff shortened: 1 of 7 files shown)


raphlinus (approved):
It feels like we should pour one out for this feature. But I believe it will return, when we're ready.

PR #307, opened 2020-09-13
Add shape specializations to d2d.

### LINKED ISSUE #60: d2d: Use specialized functions for fill/draw
The direct2d backend converts every shape to a path and build a geometry from this.
The API provides specialized functions for most of the shapes which should be prefered.

(TODOs in the code, this issue is mostly for tracking these)

smmalis37:
Fixes #60.

I plan to tackle the rest of the TODO comments on this, but I wanted to ensure I'm doing things correctly first. I haven't done much testing, are the snapshots good enough or should I do more? Also, do the FillRules affect simple shapes? A quick read of the docs on msdn leads me to think no, but this is all new to me. If not then I'll probably refactor the two methods into just calling a combined `fill_impl`, but if there needs to be something more involved then they may stay separate.  And does calling fill on a line actually need to do any work (like I've currently got) or will it all be handled by `stroke` (which will then get specialized to call DrawLine)?

### DIFF
--- a/piet-direct2d/src/d2d.rs
+++ b/piet-direct2d/src/d2d.rs
@@ -170,6 +170,10 @@ fn optional<T>(val: &Option<T>) -> *const T {
     val.as_ref().map(|x| x as *const T).unwrap_or(null())
 }
 
+fn stroke_style_to_d2d(style: Option<&StrokeStyle>) -> *mut ID2D1StrokeStyle {
+    style.map(|ss| ss.0.as_raw()).unwrap_or(null_mut())
+}
+
 impl D2DFactory {
     /// Create a new Direct2D factory.
     ///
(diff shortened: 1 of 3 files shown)


raphlinus:
Quick answers to the questions raised:

FillRule does not affect simple shapes. It only changes rendering when a shape has multiple subpaths or crosses over itself, which certainly won't happen for circle, rounded rect, etc.

Doing fill on a line is not meaningful, and I'm not sure what we expect to happen in that case. What I'd personally do is not specialize it, just let it fall through to a geometry builder. Obviously stroke is a different story :)

The snapshots are probably good enough for now, but more testing is always a good thing.

piet-direct2d/src/lib.rs
```diff
@@ -418,6 +408,24 @@ impl<'a> RenderContext for D2DRenderContext<'a> {
 }
 
 impl<'a> D2DRenderContext<'a> {
+    fn fill_impl(&mut self, shape: impl Shape, brush: &impl IntoBrush<Self>, fill_rule: FillRule) {
+        let brush = brush.make_brush(self, || shape.bounding_box());
+
+        // TODO: do something special (or nothing at all) for line?
+        if let Some(rect) = shape.as_rect() {
```
smmalis37, line 415: I wish there was an enum we could exhaustive match on so that anytime a new shape gets added you can't forget to handle it here. How likely do you think it is that new shape types will be added in the future?

smmalis37:
Alright I think this is ready. This is probably best reviewed commit-by-commit. The last commit is completely optional, but it reduces the number of `as` casts in the codebase by 1 and gains a bit of float precision, so I consider it a win.

piet-direct2d/src/d2d.rs
```diff
@@ -167,6 +170,10 @@ fn optional<T>(val: &Option<T>) -> *const T {
     val.as_ref().map(|x| x as *const T).unwrap_or(null())
 }
 
+fn stroke_style_to_d2d(style: Option<&StrokeStyle>) -> *mut ID2D1StrokeStyle {
```
smmalis37, line 173: This lives here instead of in conv because the field is private. Could move this to conv if we're ok with making the field pub.

smmalis37:
Those are some tiny diffs, but they are diffs. Are they acceptable?

raphlinus:
The diffs are acceptable, and not surprising. There isn't an expectation of pixel-for-pixel accuracy, and it's taking fairly different paths through Direct2D. The thing to do, as usual, is look at them by eye to verify that they convey the intent. I see you've already uploaded snapshots, so hopefully that'll be an easy process.

(I haven't got a chance to sit down and review the code carefully yet, but soon)

smmalis37:
Haven't made a snapshot PR yet, that other one is further similar work. I'll make one now.

smmalis37:
Looking at the diffs again, 6 has an actually noticeable diff on the bottom left circle, where the segments all seem to be changing size slightly. Not sure why, but I'm ok with it if you are.

piet-direct2d/src/lib.rs
raphlinus: The theory is that failing to handle a shape is totally harmless, and not all back-ends will implement all shapes. Realistically I don't think the number will grow much.

piet-direct2d/src/lib.rs
```diff
@@ -418,6 +388,55 @@ impl<'a> RenderContext for D2DRenderContext<'a> {
 }
 
 impl<'a> D2DRenderContext<'a> {
+    fn fill_impl(&mut self, shape: impl Shape, brush: &impl IntoBrush<Self>, fill_rule: FillRule) {
+        let brush = brush.make_brush(self, || shape.bounding_box());
+
+        // TODO: do something special (or nothing at all) for line?
```
raphlinus, line 394: As mentioned, this comment could be changed from a TODO to an explanation why we're not handling lines.

raphlinus (approved):
This looks great, thanks!

piet-direct2d/src/lib.rs
smmalis37: I figured theres a chance that it could be possible to turn lines into a noop, or something more efficient than the fallback, and if someone does something like me and draws 1000 lines that may end up being worthwhile.

Issue #309, opened 2020-09-13
(D2D) clip<RoundedRect> is expensive


smmalis37:
Admittedly my test case (https://github.com/smmalis37/sudoku_rust) with over 700 labels visible at once may be a bit extreme, but this shows up in profiling as one of the most expensive functions throughout the whole app. This is due to a combination of things:
- path_from_shape itself is expensive, spending lots of time in CreatePathGeometry, PathGeometry::Open and GeometrySink::close
- both CreateLayer and PushLayer take up a lot of time
- dropping the PathGeometry and the Layer at the end of the function are also surprisingly expensive

smmalis37:
MSDN has this to say: "In general, you should create geometries once and retain them for the life of the application, or until they need to be modified."

https://docs.microsoft.com/en-us/windows/win32/api/d2d1/nn-d2d1-id2d1geometry

PR #310, opened 2020-09-14
Add shape specialization to d2d geometry generation.

smmalis37:
Working on #309. With this change my time spent in path_from_shape (now also geometry_from_shape) is cut roughly in half.

Note that this currently builds on top of #307. Only the last 2 commits are new. That PR should be merged first, and then this rebased on top of that merge (and then given a fresh CI run), before merging.

### DIFF
--- a/piet-direct2d/src/d2d.rs
+++ b/piet-direct2d/src/d2d.rs
@@ -21,9 +21,10 @@ use winapi::shared::dxgiformat::DXGI_FORMAT_R8G8B8A8_UNORM;
 use winapi::shared::minwindef::TRUE;
 use winapi::shared::winerror::{HRESULT, SUCCEEDED};
 use winapi::um::d2d1::{
-    D2D1CreateFactory, ID2D1Bitmap, ID2D1BitmapRenderTarget, ID2D1Brush, ID2D1Geometry,
-    ID2D1GeometrySink, ID2D1GradientStopCollection, ID2D1Image, ID2D1Layer, ID2D1PathGeometry,
-    ID2D1SolidColorBrush, ID2D1StrokeStyle, D2D1_ANTIALIAS_MODE_PER_PRIMITIVE, D2D1_BEZIER_SEGMENT,
+    D2D1CreateFactory, ID2D1Bitmap, ID2D1BitmapRenderTarget, ID2D1Brush, ID2D1EllipseGeometry,
+    ID2D1Geometry, ID2D1GeometrySink, ID2D1GradientStopCollection, ID2D1Image, ID2D1Layer,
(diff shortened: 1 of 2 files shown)


piet-direct2d/src/lib.rs
```diff
@@ -109,45 +109,54 @@ impl<'b, 'a: 'b> D2DRenderContext<'a> {
...
-                if need_close {
-                    sink.end_figure(false);
+) -> Result<Geometry, Error> {
+    // TODO: Do something special for line?
+    if let Some(rect) = shape.as_rect() {
+        Ok(d2d.create_rect_geometry(rect)?.into())
+    } else if let Some(round_rect) = shape.as_rounded_rect() {
+        Ok(d2d.create_round_rect_geometry(round_rect)?.into())
+    } else if let Some(circle) = shape.as_circle() {
+        Ok(d2d.create_circle_geometry(circle)?.into())
```
smmalis37, line 124: I'm open to alternative ideas on how to achieve the upcast here.

smmalis37:
This is gonna fail CI for now because #307 is failing CI.

smmalis37:
Now ready for review.

piet-direct2d/src/d2d.rs
```diff
@@ -146,6 +155,30 @@ impl From<Error> for piet::Error {
     }
 }
 
+impl From<PathGeometry> for Geometry {
```
smmalis37, line 158: I'm open to other ideas on cleaner/shorter ways to achieve this upcast.

raphlinus: I think this is ok. From a performance perspective, upcasts in COM are basically free, but downcasts require QueryInterface.

raphlinus (approved):
This patch looks good to me. My understanding is that it will reduce the geometry creation overhead for clips, but no expected gain on fills or strokes, as they already have specialized methods. And for clips, I'm not expecting huge gains, because I expect more overhead in the layer creation and compositing. There's a bigger win for moving clips of rects to the axis-aligned specialization, but as discussed on the other issue, that won't help if the clip itself is a rounded rect.

Regarding the TODO, clipping using Line makes just as little sense as filling, so there's no need to do anything special in that case.

piet-direct2d/src/d2d.rs
smmalis37: Yeah, I just meant in terms of lines of code.

smmalis37:
Your estimation of where the gains are is accurate, but in my case its still a decently sized win.

piet-direct2d/src/d2d.rs
raphlinus: By COM standards involving any kind of subclassing, I've seen a lot worse :)

PR #314, opened 2020-09-22
Fix incorrect type alias in piet-common

cmyr:
This apparently isn't actually used anywhere?


raphlinus (approved):
Not surprising it isn't used, the type is usually inferred. But it's important to have anyway, as it certainly is easy to imagine helper functions or whatnot that would type it explicitly.

PR #315, opened 2020-09-24
Add TextStorage trait

cmyr:
This is a further refinement of how we want to store the
actual text backing a layout object. The motivation here
is to allow the client maximum flexibility in terms of
how they store their text.

This moves from using `Arc<str>` everywhere to using
`Rc<dyn TextStorage>`; this introduces another layer
of pointer indirection in various cases, but saves us
needing to add a type parameter.

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -4,14 +4,14 @@ mod grapheme;
 mod lines;
 
 use std::ops::RangeBounds;
-use std::sync::Arc;
+use std::rc::Rc;
 
 use cairo::{FontFace, FontOptions, FontSlant, FontWeight, Matrix, ScaledFont};
 
 use piet::kurbo::{Point, Rect, Size};
 use piet::{
     util, Color, Error, FontFamily, FontStyle, HitTestPoint, HitTestPosition, LineMetric, Text,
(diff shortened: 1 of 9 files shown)


raphlinus (approved):
I see the value of this.

Some questions for discussion:

I'm a bit surprised not to see a clone mechanism here, as all of the impls are Clone, and in particular the rc'ed smart pointers support cheap clone. I'm not sure what type magic is involved; I know from recent experience it can get tricky.

Am I reading this right that this may evolve into an (optional) rich text trait? So that there's a get_spans method that returns None on raw strings?

raphlinus:
Obviously I don't expect to merge while the Windows CI is red. It looks a little concerning, there's a lifetime issue, but hopefully won't be too bad to sort out.

cmyr:
There's no `Clone` bound because internally we use `Rc<dyn TextStorage>` and so we have clone that way; we could add the bound but I think it would mostly be noise?

I'm not sure if the rich text stuff will actually make it down to that level; but something like you describe will be in druid.

raphlinus:
Re the clone bound, got it. I think I got confused seeing reference-counted containers as the inners.

We'll come back to rich text later :) For now this seems fine.

Issue #311, opened 2020-09-14
Web example text not showing up


Ben-PH:
Details shown in this zulip chat: [link](https://xi.zulipchat.com/#narrow/stream/255910-druid-help/topic/text.20not.20showing.20up/near/209961417)

Nothing more to add. Will follow up with a check using 0.6

Ben-PH:
checked out to version 0.6: 

![screengrab](https://user-images.githubusercontent.com/16680090/93039886-fb27f700-f648-11ea-9fe3-174b89ea09dd.png)

sebastiano-barrera:
Hi... I have recently had the same issue as you with druid and piet on WebAssembly, and I'm also interested in helping get this fixed. Is there anything I could do to help?

Also, could you please clarify what you mean by "checked out version 0.6"? I believe piet goes up to 0.2, and I'm having this exact problem with druid 0.6.

Thank you!

cmyr:
Yes, this is broken as of the recent text work. Help would definitely be welcome; it would involve digging into the text stuff in `piet-web/src/text`. Web text doesn't need to have parity with text on other platforms, but text should definitely show up!

sebastiano-barrera:
Yes, I read something about it on Zulip, but I feel like I'm missing on some context.

In particular, that remark about 0.6 working sounds like a clue, but I'm not sure I quite understand... druid 0.6 does not work for me, and there is no piet >= 0.6, right? 

If you could help figure that out, I could try to try doing the bisection too. In alternative, I'm available to team-up with @Ben-PH, or continue on from his work if he's no longer interested in it. Whatever comes easier to you!

sebastiano-barrera:
Nice! Thank you for the pointer. I'll start to look into this as soon as I get some time 👍

PR #316, opened 2020-09-24
Update version to 0.2.0-pre4

cmyr:


raphlinus (approved):
Woot! To my mind, final is blocking on #311, nothing else that I know of, but might make sense to review.

PR #320, opened 2020-09-29
Deduplicate unpremul.

jneem:


raphlinus (approved):
Nice cleanup, thanks, and the coregraphics one was missing edge cases. I looked at the [code in Godbolt](https://godbolt.org/z/1Gxh71) and am not super pleased, but I don't think there's much to be done for it besides going to handwritten SIMD intrinsics.

jneem:
Huh, TIL that SSE pre-4.1 didn't have a floor instruction...

By the way, I don't have write access here so I can't merge it myself.

cmyr:
@jneem you should have access?

jneem:
Pretty sure I didn't before, but it looks like I do now?

PR #321, opened 2020-09-29
Adds a new ImageBuf type for representing in-memory pixel buffers.

jneem:
This is basically the same code that I recently added to druid-shell.


jneem:
The second patch changes the more "user-friendly" of the two pixel-retrieving functions to return an `ImageBuf` instead of just a `Vec` of pixels. I'm assuming it's ok to delete `raw_pixels` without a deprecation warning, because it was never in a release.

cmyr:
@jneem yep that's fine!

piet-common/src/cairo_back.rs
```diff
@@ -163,34 +163,38 @@ impl<'a> BitmapTarget<'a> {
         Ok(size)
     }
 
-    /// Get raw RGBA pixels from the bitmap.
-    pub fn raw_pixels(&mut self, fmt: ImageFormat) -> Result<Vec<u8>, piet::Error> {
+    /// Get an in-memory pixel buffer from the bitmap.
+    pub fn make_image_buf(&mut self, fmt: ImageFormat) -> Result<ImageBuf, piet::Error> {
```
cmyr, line 167: I think by ]rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) this can be `to_image_buf`?

cmyr (approved):
One naming nit but this looks great, should make working with images much easier!

piet-common/src/cairo_back.rs
jneem: That was my first attempt, but clippy complained about a `to_xxx` function taking `&mut self`...

cmyr: ah gotcha. hmhm...

cmyr: options:

- this name
- `#[ignore(clippy::XXX)]`
- interior mutability

I'm personally okay with option 2. I think the confusion of having this method requrie `&mut self` is lower than the confusion of a new prefix.

I think option 3 is also something to discuss, because these signatures do "feel" wrong, but that's a more invasive change.

Option 1 is also reasonable.

I'm not feeling very decisive; you should feel free to go with whatever you think makes sense.

jneem: I think I like option 2 best.

PR #322, opened 2020-09-29
Standardize handling of newline at EOF

cmyr:
- this should be included in the layout's reported size
- this should be reported as an additional line (count_lines())
- this should be a valid cursor position during hit testing


raphlinus (approved):
Looks good (though I didn't have time for a careful review). As I mentioned on Zulip, I'm really happy to see common tests rather than have them fragmented by back-end.

PR #328, opened 2020-10-01
Standardize behaviour of empty TextLayout

cmyr:
This includes some of the conformance testing stuff from #326, which I will close. It is an attempt to unify the behaviour of a layout object when the layout text is the empty string.

This changes piet-coregraphics and piet-cairo to behave
like piet-d2d in this regard:

- An empty layout is considered to have a single line
- the text for this line (empty string) can be retrieved
- the LineMetric for this line can be retrieved; it will
be synthesized as needed (this was already in place)

This also adds conformance tests to piet-common that
check the consistency of this across implementations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
   test-stable:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


raphlinus (approved):
This makes lots of sense. The implementations should be consistent, and treating the empty line as roughly equivalent to a zero width space is the most sensible behavior I think.

Issue #323, opened 2020-09-30
cursor position, leading/trailing and affinity


cmyr:
I ran into an issue in piet-coregraphics where `hit_test_position` wasn't respecting trailing newlines, and this has led me down a bit of a rabbit hole. In particular, I'm curious that there doesn't seem to be any concept of leading/trailing or upstream/downstream here? Is this an oversight, or am I missing something?

As an illustration, imagine the text "AAA BBB" with a soft break, so lines are "AAA " and "BBB".  If I hit test position 4, both "AAA |" and "|BBB" are valid. Is there some trick I'm missing? This doesn't seem to be as much of an issue on piet-d2d, although I haven't been able to play around there quite as much.

I'm pretty sure I could implement this in piet as some kind of `cursor_pos_for_point` method, because then I could ensure I take the line of the point into account when choosing my position instead of resolving to a string index and resolving from that back to a point, which is lossy?

raphlinus:
Suuuuper good question.

Long story short, the best way to do this is probably to plumb affinity into the hit test method. I think "position + affinity" is the cleanest high-level representation of the concept. Also note that affinity is closely related to the primary/secondary distinction in BiDi.

I found it very interesting to see how DWrite deals with this question. There, it's not really a question of "position" and "point," but rather the model represents text as grapheme clusters with finite width. Each grapheme cluster then has a leading and a trailing position, and the ordering of that reflects BiDi direction. In your example above, "AAA |" is represented as the trailing position of the space, and "|BBB" is represented as the leading position of the first "B". One of the significant advantages to this model is that in a BiDi context, each grapheme cluster has a definite direction, while a "point", being in general a boundary between two grapheme clusters, can be ambiguous in this way at direction boundaries, hence the primary/secondary distinction.

The "grapheme cluster has finite width" approach is also an extremely clean mental model for the selection region issue: the selection region is simply the union of the rects defined by the grapheme clusters within the selection.

One approach is to adopt the D2D approach, and shoehorn other APIs into it. But, as I mentioned, I think point + affinity + maybe pri/sec better captures the higher level use case. If we design an interface for this, we should probably at least be thinking about BiDi implications. I actually think I could work this out in about a half hour of thinking/experimenting/writing.

raphlinus:
Ok, we discussed this thoroughly on Zulip, and here's my synthesis:

[...]

PR #329, opened 2020-10-01
Make rects_for_range return nothing for empty range

cmyr:
Previously it would return a rect with zero size.

This also removes a duplicate implementation of this
method in piet-coregraphics.

This is based off of #328


raphlinus (approved):
Agree. As I mentioned in #323, I think the right way to think about rects_for_range is the union of rectangles covered by runs within the range. Thus, an empty range should be an empty result.

PR #330, opened 2020-10-01
[cg] Compute LineMetrics greedily

cmyr:
Unlike the other backends, with coretext we were not storing
a Vec<LineMetric> in the layout object, but instead generating
them on demand.

As I work on standardizing the behaviour around newlines at EOF
and treating an empty layout as having a single line, it is easier
to reason about a single source of 'line truth'; the logic started
to get hard to follow in the previous case.

This also cleans up the API in ct_helpers somewhat; much of it
was written before some improvements were upstreamed to the
core-text crate, and we can be less fancy there.

This includes one small change to behaviour: we were previously
storing the floored baseline in the line_y_positions vector,
which I _think_ was hiding a small bug. This may cause subtle
changes to rendering behaviour, which I will correct in snapshots
if required.

### DIFF
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -54,16 +54,13 @@ pub struct CoreGraphicsTextLayout {
     attr_string: AttributedString,
     framesetter: Framesetter,
     pub(crate) frame: Option<Frame>,
-    // distance from the top of the frame to the baseline of each line
-    pub(crate) line_y_positions: Rc<[f64]>,
-    /// offsets in utf8 of lines
-    line_offsets: Rc<[usize]>,
     pub(crate) frame_size: Size,
     image_bounds: Rect,
     width_constraint: f64,
     // these two are stored values we use to determine cursor extents when the layout is empty.
     default_baseline: f64,
     default_line_height: f64,
+    line_metrics: Rc<[LineMetric]>,
 }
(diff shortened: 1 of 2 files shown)


piet-coregraphics/src/text.rs
```diff
@@ -783,82 +741,114 @@ impl CoreGraphicsTextLayout {
...
-                    }
-                }
-                panic!("error calculating utf8 offsets");
-            })
-            .collect();
-    }
-
     fn line_range(&self, line: usize) -> Option<(usize, usize)> {
         // no lines: we return the empty string
+        //FIXME: just never have no lines
```
raphlinus, line 755: Wouldn't this function be slightly cleaner if it returned a Range rather than a pair? Otoh, I wouldn't be surprised if Range created some other problem.

raphlinus (approved):
This makes sense. I didn't dig in terrible detail, but didn't see anything that caused me concern either. Thanks!

piet-coregraphics/src/text.rs
cmyr: I forget exactly the motivation for this originally, and it feels like an "if it ain't broke" situation. This is actually pretty redundant now, since `LineMetric` stores the line range, and we can just use that if we want. I tried to keep this patch minimal.

PR #331, opened 2020-10-02
Standardize hit_test_point on newline at EOL

cmyr:
When there's a newline at the end of a visual line, a hit-test
at the end of that line should return the offset of the
character preceding the newline, not the newline itself.

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -220,6 +220,9 @@ impl TextLayout for CairoTextLayout {
 
         let mut htp = hit_test_line_point(&self.font, line, point);
         htp.idx += lm.start_offset;
+        if htp.idx == lm.end_offset {
+            htp.idx -= util::trailing_nlf(line).unwrap_or(0);
+        }
         htp.is_inside &= y_inside;
         htp
     }
@@ -259,7 +262,7 @@ impl CairoTextLayout {
                 height: self.font.extents().height,
                 ..Default::default()
             })
-        } else if self.text.as_bytes().last() == Some(&b'\n') {
+        } else if util::trailing_nlf(&self.text).is_some() {
             let newline_eof = self
                 .line_metrics
                 .last()
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -577,10 +577,11 @@ impl TextLayout for CoreGraphicsTextLayout {
             None => {
                 // if we can't find a line we're either an empty string or we're
                 // at the newline at eof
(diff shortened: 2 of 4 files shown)


piet/src/util.rs
```diff
@@ -49,6 +49,11 @@ pub fn count_until_utf16(s: &str, utf16_text_position: usize) -> Option<usize> {
     None
 }
 
+/// Very naive, don't use on windows, etc etc
```
raphlinus, line 52: I'm not in love with the "eof" name. Unicode uses "EOT" for this, which is less familiar but more precise.

Also I think the comment can be improved: I suggest:

```rust
/// Determine whether the string is terminated by a newline.
///
/// Note: this implementation assumes that paragraphs are separated
/// by the newline character. This will be true almost all of the time, but
/// Unicode in its infinite wisdom provides for a number of codepoints
/// that can act as paragraph separators. At some point, we should
/// probably make sure we respect those as well.
```

piet-coregraphics/src/text.rs
```diff
@@ -583,16 +584,22 @@ impl TextLayout for CoreGraphicsTextLayout {
...
+                    + util::count_until_utf16(line_text, rel_offset)
+                        .unwrap_or_else(|| line_text.len())
             }
             // some other value; should never happen
             _ => panic!("gross violation of api contract"),
         };
 
+        // if the offset is EOL && EOL is a newline, return the preceding offset
+        let offset = if offset == metric.end_offset && util::has_newline_eof(line_text) {
+            offset - 1
```
raphlinus, line 598: Add: `// Note: assumes line terminator is a single byte`

piet-cairo/src/text.rs
```diff
@@ -220,6 +220,9 @@ impl TextLayout for CairoTextLayout {
 
         let mut htp = hit_test_line_point(&self.font, line, point);
         htp.idx += lm.start_offset;
+        if htp.idx == lm.end_offset && util::has_newline_eof(line) {
+            htp.idx -= 1;
```
raphlinus, line 224: Same "Note:" comment as below. I think the code is fine, just want a marker in place if and when we do the other Unicode separators.

raphlinus (approved):
Good to see the attention to detail on these, especially making the back-ends work consistently. Thanks!

piet-cairo/src/text.rs
cmyr: I could make `has_newline_eof` be something like `offset_to_cursor_at_eol` or similar, and it could return a `usize`...

cmyr: This is now... `fn trailing_nlf(s: &str) -> Option<usize>` 🤢

PR #335, opened 2020-10-12
Negative width constraints

### LINKED ISSUE #334: Guard against negative width in TextLayout objects
I believe that windows currently may crash when passed a negative width constraint; it should defend against this.

see https://github.com/linebender/druid/issues/1297#issuecomment-707221243

cmyr:
- fixes #334
- references linebender/druid#1297


raphlinus (approved):
Agree, some sanity-checking of inputs makes sense here.

PR #336, opened 2020-10-12
Fix new clippy lints in rust 1.47

cmyr:


raphlinus (approved):
Approve. That vec one is a good lint, a sophisticated analysis. I occasionally get annoyed by clippy lints (float compare is one I particularly dislike and am considering disabling it in Runebender), but overall the ratio of useful to annoying is solid.

PR #340, opened 2020-10-14
[d2d] Use default line height as height when text is empty

### LINKED ISSUE #339: D2D TextLayout does not use default attributes when empty
This is due to the fact that we create our `IDWriteTextFormat` when we first create the builder, and we use placeholder font & size values there; we need to be smarter about this, stashing default attributes as they're set, and only creating the base format on the first call to `range_attribute` or when we build the final layout.

cmyr:
This is a less principled approach to fixing #339 than is
outlined in that issue, but is less invasive and works just as well.

This also fixes a test in piet-common that was intended to catch
this, but that was using a layout with a default font size, which
doesn't manifest this problem.

- fixes #339
- fixes https://github.com/linebender/druid/issues/1309

### DIFF
--- a/piet-common/tests/text.rs
+++ b/piet-common/tests/text.rs
@@ -63,8 +63,16 @@ fn rects_for_empty_range() {
 #[test]
 fn empty_layout_size() {
     let mut factory = make_factory();
-    let empty_layout = factory.new_text_layout("").build().unwrap();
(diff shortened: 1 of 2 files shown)


raphlinus (approved):
Looks good, thanks!

PR #342, opened 2020-10-16
make image formats optional via features

JAicewizard:
related to [druid #1305](https://github.com/linebender/druid/issues/1305), this PR makes piet default to no supported image formats, and allow adding support via features.
If only `png` and `jpeg` are required this cuts build times by 55% on my machine, and even with rayon build times are 35% less (for just piet).
 
| added feature  | total deps | total --realses build time (s) |
| ------------- | ------------- |------------- |
| 0  | 3  | 1.3  |
| image  | 21 | 4.9  |
| image_png  | 32  | 5.5  |
| jpeg  | 33  | 6.5  |
| jpeg_rayon  | 60  | 9.7  |
| gif  | 63  | 10.2  |
| bmp  | 63  | 10.7  |
| ico  | 63  | 10.8  |
| tiff  | 65  | 12.2  |
| webp  | 65  | 12.5  |
| pnm  | 65  | 13  |
| dds  | 65  | 13.7  |
| tga  | 65  | 13.8  |
| farbfeld  | 65  | 13.8  |
| dxt  | 65  | 13.9  |
| hdr  | 66  | 14.6  |

These build times are stacking, so the hdr build time also include png and jpeg...
But you can still clearly see that it is worth not having some of these, especially for the more obscure ones like everything after webp. 
I have not benchmarked the impact on binary size, but I could make a small example program to see the impact on that.


cmyr (approved):
Okay, lets give this a shot. I was a little worried it would over-complicate things but we can always revisit if that ends up being the case.

JAicewizard:
I don't like it either, but as it is the default features of `image` add a lot of bloat in total.
Please if you have another way of doing things https://github.com/linebender/druid/issues/1305.
Should put in a PR for druid once this is in a pre-release?

cmyr:
@JAicewizard sounds good!

PR #343, opened 2020-10-20
Bump version and update dependencies

cmyr:
Fingers crossed, this will be the last pre-release.

I'd like to at least get #340 and probably #342 in before doing a release, but this gets the paperwork in order.

### DIFF
--- a/piet-cairo/Cargo.toml
+++ b/piet-cairo/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "piet-cairo"
-version = "0.2.0-pre5"
+version = "0.2.0-pre6"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 description = "Cairo backend for piet 2D graphics abstraction."
 license = "MIT/Apache-2.0"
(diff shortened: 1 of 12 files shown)


cmyr:
Thinking about it, we're just about ready to do the actual 0.2.0 release; I don't anticipate any more breaking changes.

piet-cairo/Cargo.toml
```diff
@@ -10,12 +10,12 @@ keywords = ["graphics", "2d"]
 categories = ["rendering::graphics-api"]
 
 [dependencies]
-piet = { version = "0.2.0-pre5", path = "../piet" }
+piet = { version = "0.2.0-pre6", path = "../piet" }
```
raphlinus, line 13: Did you mean "=0.2.0-pre6" here? Same is true for piet dep in the other back-ends.

raphlinus:
I also have a not-great feeling about web text (#311), as this is a regression from the previous release. But I'm not sure that should block an 0.2, I think it can be fixed in a point release.

cmyr:
Okay, decided to just fix the web text, ended up not being much to it at all. Let's get that merged, and then I'll pull the trigger on this.

PR #344, opened 2020-10-21
Fix basic drawing of text on web

### LINKED ISSUE #311: Web example text not showing up

cmyr:
This doesn't support alignment, soft breaks, or style spans, but
it does support basic default styles and drawing of multiline,
when those lines are hard breaks.

- closes #311

before:

Screen Shot 2020-10-21 at 11 53 24 AM
after:

Screen Shot 2020-10-21 at 11 52 39 AM

### DIFF
--- a/piet-web/examples/basic/src/lib.rs
+++ b/piet-web/examples/basic/src/lib.rs
@@ -7,6 +7,9 @@ use web_sys::{window, HtmlCanvasElement};
 use piet::{samples, RenderContext};
 use piet_web::WebRenderContext;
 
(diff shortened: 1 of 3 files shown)


raphlinus (approved):
Excellent! This removes one of the big concerns I had about getting piet to 0.2.

PR #346, opened 2020-10-21
Link README in Cargo.toml, add piet-common readme

cmyr:
This will improve the docs available on crates.io


README.md
```diff
@@ -17,24 +17,27 @@ A companion for Bézier path representation and geometry is [kurbo].
 
 ## Backends
 
+*For cross-platform use, the [`piet-common`][] crate reexports the most
```
raphlinus, line 20: I don't think you need the following `[]`, but maybe I'm missing something.

raphlinus (approved):

README.md
cmyr: I've started doing this in some places; it is the way you're supposed to do implicit links per the original markdown 'spec'. 🤷

PR #348, opened 2020-10-29
Adds creating a color from a hex string

wess:
Create a color from a hex string value.
```
use piet::Color;

let hex_color = "#f0f0f0";
let color = Color::from_hex(hex_color);
```

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -40,7 +40,6 @@ impl Color {
       match u16::from_str_radix(hex, 16) {
         Ok(v) => v as f64,
         Err(_) => {
-          println!("Unable to convert hex pairs: {}", hex);
           std::process::exit(128);
         }
       }
@@ -63,11 +62,11 @@ impl Color {
         3 => {
           components.iter().map(|&c| format!("{}{}", c, c)).collect()
         },
-        6 => components.iter().map(|&c| format!("{}", c)).collect()
+        _ => components.iter().map(|&c| format!("{}", c)).collect()
       };
 
 
-      let hex_rgb:&str = hex.join("").as_str();
+      let hex_rgb:&str = &hex.join("").as_str();
 
       Color::rgb(
         Color::hex_to_decimal(&hex_rgb[0..2]),
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -36,6 +36,46 @@ impl Color {
         Color::Rgba32(rgba)
     }
 
+    pub const fn hex_to_decimal(hex: &str) -> f64 {
+      match u16::from_str_radix(hex, 16) {
+        Ok(v) => v as f64,
+        Err(_) => {
+          println!("Unable to convert hex pairs: {}", hex);
+          std::process::exit(128);
+        }
+      }
+    }
+
+    /// Create a color from a hex string value.
+    ///
+    /// ```
+    /// use piet::Color;
+    ///
+    /// let hex_color = "#f0f0f0";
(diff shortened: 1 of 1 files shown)


cmyr (changes requested):
Is there a specific need for hex strings? I'm not opposed, but you can create a color from hex values already, like:

```rust
let red = Color::rgb8(0xff, 0, 0);
let semi_transparent_blue = Color::rgba8(0, 0, 0xff, 0x80);
```

which might cover your use-case.

That said, I'm not strictly opposed to creating colors from hex strings, it could make sense for things like loading colors from a json file or similar. In that case, I would like to support the various formats allowed in modern CSS; so both 6 and 3 char codes (which this does), as well as alpha variants, so that we would parse all of "`#FF00FF`, `#F0F`, `#DDFFDD88`, and `#DFD8`. (as well as without the trailing `#`.)


edit:

okay, so I started writing some comments about avoiding allocation and then I started playing around and I ended up largely implementing this.

basically: we want to avoid all the `String` and `Vec` stuff. This is pretty easy to do in modern rust, and so I think a rough sketch of what this implementation should actually look like would be something like the following, with  the main missing thing being an actual error type.


```rust
fn from_hex(hex_str: &str) -> Result<Color, ()> {
    let hex_str = hex_str.trim_start_matches('#');
    if !hex_str.is_ascii() {
        return Err(())
    }
    
    let mut four_bit_channels = match hex_str.as_bytes() {
        &[r, g, b] => [r, r, g, g, b, b, b'f', b'f'],
        &[r, g, b, a] => [r, r, g, g, b, b, a, a],
        &[r0, r1, g0, g1, b0, b1] => [r0, r1, g0, g1, b0, b1, b'f', b'f'],
        &[r0, r1, g0, g1, b0, b1, a0, a1] => [r0, r1, g0, g1, b0, b1, a0, a1],
        _ => return Err(()),
    };
    
    // convert to hex in-place
    for chan in &mut four_bit_channels {
        match hex_from_ascii_byte(*chan) {
            Ok(hex) => *chan = hex,
            Err(_bad) => {
                panic!("err {}: '{}'", hex_str, _bad);
                return Err(())
            }
        }
    }
    Ok(color_from_4bit_hex(four_bit_channels))
}

fn color_from_4bit_hex(components: [u8; 8]) -> Color {
    debug_assert!(components.iter().all(|x| *x < 16));
    let [r0, r1, g0, g1, b0, b1, a0, a1] = components;
    Color::rgba8(r0 << 4| r1, g0 << 4 | g1, b0 << 4 | b1, a0 << 4 | a1)
}

fn hex_from_ascii_byte(b: u8) -> Result<u8, u8> {
    match b {
        // 0-9
        0x30..=0x39 => Ok(b - 0x30),
        // uppercase A-F
        0x41..=0x46 => Ok(b - 0x41 + 10),
        // lowercase a-f
        0x61..=0x66 => Ok(b - 0x61 + 10),
        _ => Err(b),
    }   
}
```

So... if you're interested in updating this PR to do this or something roughly equivalent, I'd be happy to merge. The important details are:

- not const
- has an error type
- does not allocate

If this isn't interesting to you, let me know and I'll open an issue to add it eventually. :)

derekdreery:
There's also `Color::from_rgba32_u32`, meaning you can do:

```rust
let color = Color::from_rgba32_u32(0xff0000ff);
```

The only problem with that is that if you miss off the alpha channel you won't get what you expect (you'll actually be missing the red channel).

PR #350, opened 2020-11-03
Support grayscale images

ruffson:
This branch adds a type to the `ImageFormat` enum for grayscale support.
As well as adding support for it in the two backends: `coregraphics` and `cairo` (check below for limitations).

## Open Todos
- Other backends
- Since cairo does not support grayscale image types, I used the alpha type `A8`, as recommended by them on the mailing list. This is why I invert the pixel value (line 256 in ..cairo.../lib.rs). However, besides this it is important that the image has a white background. This may needs to be done automatically on a lower level instead of doing it when loading the image with `.background(Color::WHITE)`.

I confirmed that this works on Linux Fedora/GNOME.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -272,7 +272,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                     }
                     ImageFormat::GrayScale => {
                         for x in 0..width {
-                            data[dst_off + x] = buf[src_off + x];
+                            data[dst_off + x] = 255 - buf[src_off + x];
                         }
                     }
                     _ => return Err(Error::NotSupported),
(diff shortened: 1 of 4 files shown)


cmyr:
Yep this works fine on mac!

JAicewizard:
@ruffson could you implement it for the web backend as well? Right now a greyscale image needs to be copied(clone) to RGBA and then that needs to be cloned again (due to wasm-bindgen needing mutable data). It would be nice to have have this be only one clone (inside piet-web).

I can implement it if you dont have time.

ruffson:
> @ruffson could you implement it for the web backend as well? Right now a greyscale image needs to be copied(clone) to RGBA and then that needs to be cloned again (due to wasm-bindgen needing mutable data). It would be nice to have have this be only one clone (inside piet-web).

I am pretty busy right now so I cannot promise anything, but will look into it. If I can fix it blindly like for macOS, I will give it a try.

PR #352, opened 2020-11-03
Correct rtl alignment

cmyr:
I had failed to notice that text alignment was not behaving as expected for RTL scripts on windows; it turns out that you *do* need to explicitly set the `DWRITE_READING_DIRECTION` in order for `DWRITE_TEXT_ALIGNMENT_LEADING` to correspond to the right edge in a RTL script.

In the examples below, the first two columns of arabic text should be aligned to the opposite edge than the corresponding columns of english text.

### Before:
d2d-test-7
### After:
d2d-test-7


raphlinus (approved):
I didn't analyze the logic carefully, but if it makes the results more consistent, sounds good.

PR #353, opened 2020-11-09
Update versions for v0.2.0

cmyr:


piet-cairo/Cargo.toml
```diff
@@ -11,12 +11,12 @@ keywords = ["graphics", "2d"]
 categories = ["rendering::graphics-api"]
 
 [dependencies]
-piet = { version = "=0.2.0-pre6", path = "../piet" }
+piet = { version = "=0.2.0", path = "../piet" }
```
raphlinus, line 14: I think these *can* be `=` dependencies, but they don't have to be, and we haven't done that before. Point releases are supposed to be semver-compatible, and I think we can use those for performance improvements etc.

What I think would be a reasonable policy is `=` for `-pre` releases and a caret (default) dep otherwise.

raphlinus (approved):

PR #359, opened 2020-12-02
Add TextLayout::trailing_whitespace_width

### LINKED ISSUE #357: No easy way to get the whitespace-inclusive width of a text layout
This has come up in druid, and is a thing that we really should provide. There are various ways to go about it, with the easiest being some method on `TextLayout` like `width_including_whitespace()`. It's also arguable that we should provide a `whitespace_width` field on `LineMetric`, although this may be slightly annoying to implement on windows.

cmyr:
Another little oversight. This will probably require us to go
to 0.3.0, but that will also let us get the greyscale image stuff
landed.

closes #357

I've only tested on mac, so i'm sure this will fail somewhere in CI. Also it looks like the cairo backend *was* including ws width, maybe? We'll find out shortly... 😬

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -37,6 +37,7 @@ pub struct CairoTextLayout {
     // color here and then just grab it when we draw ourselves.
     pub(crate) fg_color: Color,
     size: Size,
+    trailing_ws_width: f64,
     pub(crate) font: ScaledFont,
     pub(crate) text: Rc<dyn TextStorage>,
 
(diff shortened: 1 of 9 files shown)


derekdreery:
https://github.com/linebender/piet-snapshots/blob/e5aa031c34f70752bd274d2ead5543b6bf2fd27a/cairo/cairo-test-10.png looks like the text is overflowing its box. Is this expected/unrelated to this PR?

cmyr:
The problem there is that the toy text API does not know how to handle newlines, so it's missing glyph symbols in their place. It is correct that the newline is not included in the width calculation, though, and so they should appear outside of the box, which does not include trailing whitespace in its width. (Previously cairo *was* including trailing whitespace in width calculations, but that was a bug.)

derekdreery (approved):
I'm not 100% sure of this code, but it looks reasonable.

raphlinus (approved):
Yup, looks good. One of my metrics it that the DW implementation is super clean and simple. The fact that CG and Cairo are a bit messier is not surprising to me :)

PR #360, opened 2020-12-03
Finish up grayscale work

### LINKED ISSUE #313: Add support for gray-scale images
## Problem description
Currently the only supported color formats offered by `piet::render_context::ImageFormat` are `Rgb`, `RgbaSeparate` and `RgbaPremul`. It would be nice to have a format representing gray-scale images with 8bit values per pixel.

## Obstacles
Unfortunately it looks like it is not just a matter of adding a new item to `ImageFormat` and the backends, as some backends do not support gray-scale images.
### Linux/GTK backend
By looking at `cairo-rs` it was discovered that `cairo` itself does not support gray-scale image formats. As its [documentation shows](https://www.cairographics.org/manual/cairo-Image-Surfaces.html#cairo-format-t) the only 8bit format `A8` is supposed to be used for alpha values.
### macOS/core graphics
Similary the macOS backend is also not able to deal with grayscale and raises an `unimplemented` error [here](https://github.com/ruffson/piet/blob/bd86db92b18224750eff34a8f19d78e522c3d041/piet-coregraphics/src/lib.rs#L283).

## Workarounds?
If anyone has any suggestion on a way to display gray-scale images with a linux supported backend efficiently, then please share your knowledge here.
I am dealing with quite large images and the only way I have found to display them right now is to save each original pixel value three times in a new vector (I know).

// _Updated with macOS backend findings._

cmyr:
This is based off of #350; I'd like to get this landed before the next release, but we should at least get d2d support working.

- closes #350, closes #313

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -222,6 +222,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         let cairo_fmt = match format {
             ImageFormat::Rgb => Format::Rgb24,
             ImageFormat::RgbaSeparate | ImageFormat::RgbaPremul => Format::ARgb32,
+            ImageFormat::Grayscale => Format::A8,
             _ => return Err(Error::NotSupported),
         };
         let mut image = ImageSurface::create(cairo_fmt, width as i32, height as i32)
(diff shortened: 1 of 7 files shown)


piet-cairo/src/lib.rs
```diff
@@ -269,6 +270,11 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                             data[dst_off + x * 4 + 3] = a;
                         }
                     }
+                    ImageFormat::Grayscale => {
```
raphlinus, line 273: It seems to me it should be possible to avoid this. I see a TODO in #350 regarding this. Given that it's reported to work, I'm inclined to accept this now, and consider a more efficient render path to be potential followup work.

raphlinus (approved):
Overall looks pretty good, though I have concerns the workaround in Cairo is being done the most efficient way. Thanks!

piet-cairo/src/lib.rs
cmyr: The more I think about this the more I think it's just incorrect, and pushes too much expectation down to the consumer; the behaviour coming out of piet should be the same for all backends, which means we should just use rgb24. :shrug:

ruffson: Wait so instead you are quadrupling the data? How is this more efficient? I am dealing with 500MB+ images, this looks very bad to me.

PR #361, opened 2020-12-04
Web greyscale

JAicewizard:
This ads grayscale support to the web backend.


JAicewizard:
this is probably blocked on https://github.com/linebender/piet-snapshots/pull/16 for the snapshot updates

cmyr:
This looks good; I think the problem is you have committed an old version of the snapshots? I would do the following:

```sh
git fetch origin
git rebase origin/master
git add snapshots
git commit -m 'Fixup snapshots'
git push -f
```

or something similar; basically you need to update the `snapshots` submodule to be the same as masters, and then commit the submodule. :)

JAicewizard:
yes I was just waiting on the snapshot PR, now I should be able to just update the sub-module and be done with it.

JAicewizard:
This put the sub-modules back at master instead of a specific commit.

cmyr:
submodules are always at a specific commit; they don't update automatically (which is a good thing). the correct commit right now is definitely the current `master`, though!

cmyr:
(This is why it works without needing to merge the PR; that commit still exists before the merge happens, and the merge doesn't change the commit. If it needed to be merged then the PR that originally motivated that commit (#360?) also would not have passed.)

cmyr (approved):
Thanks!

cmyr:
In any case very nice to keep the web stuff up to date, it's often easy to let it fall behind because we don't test it as thoroughly.

JAicewizard:
This probably helps me out a bunch. Right now I create the RGBA image in a working thread -> clone to js -> clone to main thread -> clone to wasm -> clone because of https://github.com/rustwasm/wasm-bindgen/issues/2364. This might be exaggerated because the browser probably optimizes this a bit, but wouldn't bet on that.
This could reduce the amount of data having to be cloned by ~4x?

cmyr:
It's a good question, and depends on how the image data is stored.

This will definitely reduce the size of whatever you are passing around *before* you call `RenderContext::make_image`, but because we're then drawing into a canvas element we have to convert to rgba at that point, and so the returned object is still as big as it was before.

There's another thing you could consider, too: If you're storing your image data initially as a `Vec<u8>`, you could switch this to a `Arc<[u8]>` (which is easy, you can just call `into()` from a `Vec<u8>`) and then cloning *that* data will become as cheap as cloning any `Arc` is.

Additionally, if we're seeing that people are cloning the image returned from `piet::make_image`, we could make sure that this is also behind an `Arc` wherever possible.

JAicewizard:
It was more of a rhetorical question :)

The problem isn't rust, its JS. JS is 100% single threaded and just like with languages like dart dont share memory with worker threads. The same is true for JS/wasm, the only way to share memory is to copy all of it, there are no fancy tricks around that form what I know. 

There are shared array buffers, and once they become a thing again are definitely also useful. I might look into that one as well.
Basically that is just a pointer to data shared between threads/wasm and it is even possible to write to it on one end and have it propagate to the other threads. That would be huge for my use-case.

PR #362, opened 2020-12-05
Remove soft break tofu in cairo

derekdreery:
I still need to figure out how to regenerate the snapshots.

### DIFF
--- a/piet-cairo/src/text/lines.rs
+++ b/piet-cairo/src/text/lines.rs
@@ -132,10 +132,17 @@ pub(crate) fn calculate_line_metrics(text: &str, font: &ScaledFont, width: f64) 
             }
 
             // now do the hard break
+            let mut line_end = line_break;
+            // Remove any trailing newline characters from the line.
+            while matches!(text.get(line_end - 1..line_end), Some("\n") | Some("\r"))
+                && line_end > line_start
+            {
+                line_end -= 1;
+            }
             add_line_metric(
                 text,
                 line_start,
-                line_break,
+                line_end,
                 baseline,
                 height,
                 &mut y_offset,


derekdreery:
It might be that this has to wait until proper shaping, but I will try to land something sensible sooner.

piet-cairo/src/text/lines.rs
```diff
@@ -132,10 +132,17 @@ pub(crate) fn calculate_line_metrics(text: &str, font: &ScaledFont, width: f64)
             }
 
             // now do the hard break
+            let mut line_end = line_break;
+            // Remove any trailing newline characters from the line.
+            while matches!(text.get(line_end - 1..line_end), Some("\n") | Some("\r"))
```
cmyr, line 137: Was slightly confused here for a second; I couldn't tell that this wouldn't screw up in cases where there were multiple newlines in a row, which should be treated as actual newlines.

Since this is linux I'm also not sure we care about `\r`?

just for fun I'd have probably written something like,

```rust
let line_text = &text[line_start..line_end];
let newline_len = line_text.as_bytes().rev().take_while(|b| b == b'\n'').count();
let line_end = line_end - newline_len;
```

cmyr (approved):
Cool, definitely worth getting this cleaned up.

There are instructions for adding snapshots somewhere, I think in `CONTRIBUTING.md`. Ping me on zulip if you need help!

piet-cairo/src/text/lines.rs
derekdreery: Yeah `\r` is probably overkill.

derekdreery: I also think your solution is better.

cmyr:
going to close this since we'll resolve in #389.

PR #363, opened 2020-12-08
Add the basic colors from css3.

derekdreery:
This patch adds the basic CSS3 colors as constants to `Color`. They are present as a convenience if you want to use a standard color. What do you think?

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -193,9 +193,53 @@ impl Color {
         )
     }
 
-    /// Opaque white.
-    pub const WHITE: Color = Color::rgb8(0xff, 0xff, 0xff);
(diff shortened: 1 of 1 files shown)


cmyr:
I personally think having a few more color constants make sense, but raph and I had a minor disagreement on this point; his argument was that including more constants would encourage people to use them, instead of actually picking *good* colors.

I think this argument is fair, but I also think that just for things like debugging it can be useful to have some more constants available, so I'm in favour.

raphlinus:
I'll agree with Colin here. Colors shipped in the UI should be coming from a theme anyway.

cmyr (approved):

PR #365, opened 2020-12-13
Fix errors when drawing empty image.

derekdreery:
# Implement empty images

Whilst making an image viewer, I wanted a default value for the image to display. What I found is that if I used the default constructor `ImageBuf::empty` I was hitting a panic.

Even worse than this, when I then made tests for empty images and tried out the CI, I found that on macos it led to a segmentation fault. This would always be caught, since I think that pointers that don't point anywhere (to zero-sized types) are set to point so some rediculous memory address that tells the operating system/processor to issue a memory page fault, but still it's UB which in the Rust world is *bad*. I raised an [issue with core-foundations-rs](https://github.com/servo/core-foundation-rs/issues/431) where they said the would do an `assert` to check for empty `[u8]`s, but still that would be a panic in our code, like Windows and Cairo.

This patch adds code to handle the case where an empty image is both created and drawn, and essentially does nothing in both cases. I argue that this is expected behaviour, and that the empty image is a useful placeholder where you have an image that is optionally present (you avoid needing to use things like the `Either` widget). An alternative would be to check the image dimensions and panic when they are incorrect, but I think this is a less pleasant user experience. (If you disagree please say so on the comments below)

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -224,8 +224,16 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
             ImageFormat::RgbaSeparate | ImageFormat::RgbaPremul => Format::ARgb32,
             _ => return Err(Error::NotSupported),
         };
-        let mut image = ImageSurface::create(cairo_fmt, width as i32, height as i32)
+        let width_int = width as i32;
+        let height_int = height as i32;
+        let mut image = ImageSurface::create(cairo_fmt, width_int, height_int)
             .map_err(|e| Error::BackendError(Box::new(e)))?;
+
+        // early-return if the image has no data in it
+        if width_int == 0 || height_int == 0 {
+            return Ok(image);
+        }
+
         // Confident no borrow errors because we just created it.
         let bytes_per_pixel = format.bytes_per_pixel();
         let bytes_per_row = width * bytes_per_pixel;
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -333,11 +333,7 @@ fn draw_image<'a>(
     };
     // Cairo returns an error if we try to paint an empty image, causing us to panic. We check if
     // either the source or destination is empty, and early-return if so.
-    if src_rect.width() == 0.
-        || src_rect.height() == 0.
-        || dst_rect.width() == 0.
-        || dst_rect.height() == 0.
-    {
+    if src_rect.is_empty() || dst_rect.is_empty() {
         return;
     }
 
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -319,6 +319,20 @@ fn draw_image<'a>(
     dst_rect: Rect,
     interp: InterpolationMode,
 ) {
+    let src_rect = match src_rect {
+        Some(src_rect) => src_rect,
+        None => Size::new(image.get_width() as f64, image.get_height() as f64).to_rect(),
+    };
+    // Cairo returns an error if we try to paint an empty image, causing us to panic. We check if
+    // either the source or destination is empty, and early-return if so.
+    if src_rect.width() == 0.
+        || src_rect.height() == 0.
+        || dst_rect.width() == 0.
+        || dst_rect.height() == 0.
+    {
+        return;
+    }
+
     let _ = ctx.with_save(|rc| {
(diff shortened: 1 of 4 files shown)


derekdreery:
I'm not sure how to test this other than wait to see if it fixes the problems in druid. Suggestions welcome!

cmyr:
@derekdreery a good way to test would be to add a test in `piet-common/tests`; these tests are run against the three major backends, so we can ensure that behaviour is consistent everywhere; your test can use the [`#[should_panic]` attribute](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic).

derekdreery:
I added a test, and it showed that actually I had not fixed the problem (which I now have). That's why tests are good! :)

derekdreery:
@cmyr interesting! It seems that an empty image segfaults on Mac, and also probably on Windows (I'm not so up to speed on windows memory access violations).

derekdreery:
I'm trying to write fixes for mac/win, but I am relying on CI to test my solutions for me.

piet-cairo/src/lib.rs
```diff
@@ -319,17 +327,27 @@ fn draw_image<'a>(
     dst_rect: Rect,
     interp: InterpolationMode,
 ) {
+    let src_rect = match src_rect {
+        Some(src_rect) => src_rect,
+        None => Size::new(image.get_width() as f64, image.get_height() as f64).to_rect(),
+    };
+    // Cairo returns an error if we try to paint an empty image, causing us to panic. We check if
+    // either the source or destination is empty, and early-return if so.
+    if src_rect.width() == 0.
```
cmyr, line 336: Is there no `Rect::is_empty`? Maybe just `src_rect().area() == 0.0`?

derekdreery: Hmm, I could add a `Rect::is_empty`. The `area` solution will work also. I'll have a look at this after I've stopped windows/mac crashing.

derekdreery:
@cmyr would you mind taking over this PR? There is a segfault in the macOS code when drawing an empty image. I can't debug it as I don't have access to a macos computer, and using the CI is too slow.

cmyr:
@derekdreery if you would like to close this PR and then push your branch to `origin` and open a new PR (so I an easily add commits), I would be happy to fix the macOS stuff!

derekdreery:
Please don't merge - I'm just experimenting using the CI.

derekdreery:
Right @cmyr I think this is ready for review! :grinning: 

I realised in the end that I could use `cargo clippy` on mac if I added the darwin target, which meant that I only had to use the CI after I know it would compile.

I'll update the first comment with the final details and some discussion about whether we want this.

piet-direct2d/src/lib.rs
```diff
@@ -532,6 +544,15 @@ fn draw_image<'a>(
     dst_rect: Rect,
     interp: InterpolationMode,
 ) {
+    let src_size = image.get_size();
+    if dst_rect.width() == 0.
```
raphlinus, line 548: I think `dst_rect.is_empty()` might be clearer here.

piet-coregraphics/src/lib.rs
```diff
@@ -336,9 +361,11 @@ impl<'a> RenderContext for CoreGraphicsContext<'a> {
         dst_rect: impl Into<Rect>,
         _interp: InterpolationMode,
     ) {
-        if let Some(cropped) = image.cropped(to_cgrect(src_rect)) {
-            // TODO: apply interpolation mode
-            self.ctx.draw_image(to_cgrect(dst_rect), &cropped);
+        if let CoreGraphicsImage::NonEmpty(image) = image {
+            if let Some(cropped) = image.cropped(to_cgrect(src_rect)) {
+                // TODO: apply interpolation mode
```
raphlinus, line 366: Out of scope for this PR, which is fine, but this TODO bothers me. There should be a common impl with this and `draw_image` that takes care of interpolation mode.

raphlinus (approved):
Looks good, thanks. Improving robustness is important, and any way for user code to cause a segfault should be a priority.

piet-direct2d/src/lib.rs
derekdreery: Sorry I missed this, will fix before merge, and I'll have a quick look to make sure I didn't miss any more.

PR #368, opened 2020-12-16
Always include LineMetric for line 0

### LINKED ISSUE #366: Ensure there is always at least one LineMetric
`TextLayout::line_metric(0)` should always return something; if the text layout is empty we should still synthesize an object that includes the correct default line height (or a reasonable guess).

cmyr:
This matches the behaviour of other backends and fixes a bunch
of broken druid examples.

- closes #366


derekdreery (approved):
Definitely useful - I would expect zero-length text to have 1 empty line_metric.

PR #369, opened 2020-12-16
Fix errors when drawing empty image.

derekdreery:
Continuation of #365 .


cmyr:
Okay after some initial poking there doesn't seem to be an obvious answer here, coregraphics just doesn't want us to pass an empty array here.

backing up, though, I'm newly curious for the rationale behind wanting zero-sized images; My inclination would be to have piet return an error itself if you try and create an image with no size?

derekdreery:
So currently there is an `empty` constructor, so I assumed that you wanted to support empty images. Getting rid of it and making empty images an error/panic is certainly an option. Another option is returning a single pixel transparent image.

derekdreery:
Even if we prevent empty images, there are still empty targets and images cropped to empty that could cause crashes.

derekdreery:
It actually seems that the code segfaults quite early: at the call to `CGDataProvider::from_buffer`. I've [raised an issue with `core-graphics`] because that's a safe method that should handle 0-size types, if only by panicking.

[raised an issue with `core-graphics`]: https://github.com/servo/core-foundation-rs/issues/431

derekdreery:
Note to self: the first thing to do to fix the mac impl is to not pass an empty buffer to the `DataProvider`. A single static buffer like `[0]` could be used instead, with basically no overhead & no allocations (I think). This might be enough to make it work (and do nothing).

PR #370, opened 2020-12-16
Add `Image` trait.

### LINKED ISSUE #367: [Feature request]: Add an `Image` trait as below
This is a proposal/feature request/RFC: Add the following trait to piet

```rust
pub trait Image {
    fn size(&self) -> Size;
}
```

It would be nice to to have to keep an `ImageBuf` around to know the size of the image.

derekdreery:
Closes #367.

The trait allows you to get an image's size in a cross-platform way.

Also renames `piet_common::Image` to `PietImage` to avoid clash. This is a breaking change, but if we want to add the `Image` trait and make the cross-platform type alias names more consistent, better to do it now than later.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -10,8 +10,8 @@ use cairo::{Context, Filter, Format, ImageSurface, Matrix, SurfacePattern};
 
 use piet::kurbo::{Affine, PathEl, Point, QuadBez, Rect, Shape, Size};
 use piet::{
-    Color, Error, FixedGradient, ImageFormat, InterpolationMode, IntoBrush, LineCap, LineJoin,
-    RenderContext, StrokeStyle, TextLayout,
+    Color, Error, FixedGradient, Image, ImageFormat, InterpolationMode, IntoBrush, LineCap,
+    LineJoin, RenderContext, StrokeStyle, TextLayout,
 };
 
 pub use crate::text::{CairoText, CairoTextLayout, CairoTextLayoutBuilder};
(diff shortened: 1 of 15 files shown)


derekdreery:
I'd like to add a test, but it would be a merge conflict with #369, so I'd rather add it after that's merged.

derekdreery:
Ready for review.

piet-cairo/src/lib.rs
```diff
@@ -309,7 +298,13 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         dst_rect: impl Into<Rect>,
         interp: InterpolationMode,
     ) {
-        draw_image(self, image, Some(src_rect.into()), dst_rect.into(), interp);
+        draw_image(
```
cmyr, line 301: it's strange to me that this isn't just a method on `CairoImageContext` itself, but that can be out of scope for this PR.

piet-common/tests/image.rs
```diff
@@ -66,3 +66,19 @@ fn empty_image_area_should_not_panic() {
...
+        if image.size() != Size::new(1., 1.) {
+            return Err(format!(
+                "expected {:?}, found {:?}",
+                Size::new(1., 1.),
+                image.size()
+            ));
+        }
+        Ok(())
+    })
+}
```
cmyr, line 84: do we also want to explicitly test a zero-sized image? I'm not sure what our expectations are there, now...

piet/src/image.rs
```diff
@@ -22,6 +22,12 @@ use crate::kurbo::Size;
 use crate::util::unpremul;
 use crate::{Color, ImageFormat, RenderContext};
 
+/// Represents an image type that knows its size.
```
cmyr, line 25:
```suggestion
/// A trait for a backend's bitmap image type.
```

Don't love my version either, but in general I'd try and make the docs slightly higher level.

cmyr (approved):
I think this makes lots of sense, and generally makes the API more consistent.  Thanks!

piet-cairo/src/lib.rs
derekdreery: I've moved it to be a method on `CairoRenderContext`.

piet-common/tests/image.rs
derekdreery: Yeah, good idea.

derekdreery:
I remember now, I actually solved the empty image issue for D2D by creating a transparent 1x1 pixel, which of course returns a size of `1x1`. I think the best solution is to add a flag to `Bitmap` to denote whether it's an empty image. I can then use this to completely skip rendering in the empty image case as well.

PR #372, opened 2020-12-21
Add a feature "serde" and forward it to kurbo.

derekdreery:
If this lands then a follow-up to https://github.com/linebender/druid/pull/1477 could be added to enable serde on kurbo.

It also has a very small drive-by clean up of doc links.


cmyr (approved):
Makes sense to me.

PR #373, opened 2020-12-22
Color constructors take f64 instead of impl Into<f64>

### LINKED ISSUE #371: Color::rgb usage not very "obvious"
This function actually expects the values to be between 0 and 1, however it accepts anything that can be converted to a f64. So I used it like so rgb(85, 83, 99). Which seems to translate to rgba(1.0, 1.0, 1.0). I know there are the rgb8 functions and others like it which provide other options. I was under the impression that it would scale the number appropriately and was just trying to figure out why the background was always just a white background no matter what value I chose. Could this function be more restrictive?

cmyr:
This will make these methods harder to misuse.

- closes #371

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -54,8 +54,7 @@ impl Color {
     }
 
     /// Create a color with a grey value in the range 0.0..=1.0.
-    pub fn grey(grey: impl Into<f64>) -> Color {
-        let grey = grey.into();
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
Agree, this is an improvement.

derekdreery (approved):
I think this is a valuable improvement. Users can still do the conversions themselves, if they so desire.

PR #374, opened 2020-12-22
Add Color::from_hex_str for parsing CSS-style colors

cmyr:
This code is a bit idiosyncratic because I wanted it to
work in const contexts.

Question: should we have a `FromStr` impl? This might make sense,
but it's not 100% clear since there are any number of ways to
parse a string into a color. Would this cause problems in practice?

- closes #348

reference: https://github.com/linebender/druid/issues/1471

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -12,10 +12,14 @@ pub enum Color {
     Rgba32(u32),
 }
 
-impl Debug for Color {
-    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
-        write!(f, "#{:08x}", self.as_rgba_u32())
-    }
+/// Errors that can occur when parsing a hex color.
+#[derive(Debug, Clone, PartialEq)]
+pub enum ColorParseError {
+    /// The input string has an incorrect length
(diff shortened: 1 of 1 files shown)


piet/src/color.rs
```diff
@@ -243,3 +266,93 @@ impl Color {
...
+
+const fn color_from_4bit_hex(components: [u8; 8]) -> Color {
+    let [r0, r1, g0, g1, b0, b1, a0, a1] = components;
+    Color::rgba8(r0 << 4 | r1, g0 << 4 | g1, b0 << 4 | b1, a0 << 4 | a1)
+}
+
+const fn hex_from_ascii_byte(b: u8) -> Result<u8, u8> {
+    match b {
+        // 0-9
+        0x30..=0x39 => Ok(b - 0x30),
```
raphlinus, line 306: I think these would be slightly easier to read if written `b'0'` etc.

raphlinus (approved):
Nice!

derekdreery:
Love the impl. I'd leave `FromStr` for now, can always be added later if there is a need.

PR #375, opened 2020-12-22
Prepare for 0.3.0 release

cmyr:
Bump crate versions and remove deprecations.


cmyr:
I think we just go ahead and get this out, I'm not particularly worried about churn. 🤷

raphlinus (approved):
Looks good! Curious to see what breaks, but also not very concerned.

PR #378, opened 2020-12-28
Test wasm in ci

cmyr:
Let's see if this works.


cmyr:
... should've done *this* a long time ago 🙄

cmyr:
going to merge this as-is because I want to have a commit in the tree that I can pin to for other testing.

PR #379, opened 2020-12-28
Update cairo/web to xi-unicode 0.3

cmyr:
We had alread bumped web and not noticed the breakage because
we don't actually run any tests; we'd noticed the breakage in
cairo and not bumped.

This standardizes on 0.3 and fixes the breakage by manually
addung the last line if there's no explicit linebreak.

This should address https://github.com/linebender/druid/issues/1488

### DIFF
--- a/piet-cairo/Cargo.toml
+++ b/piet-cairo/Cargo.toml
@@ -15,7 +15,7 @@ piet = { version = "0.3.0", path = "../piet" }
 
 cairo-rs = { version = "0.9.1", default-features = false } # We don't need glib
 unicode-segmentation = "1.3.0"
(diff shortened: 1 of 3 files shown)


raphlinus (approved):
I haven't audited the logic carefully, but this seems ok to me.

Note that xi-unicode 0.3 does change the reporting of the final line break - it's "hard" if the text ends in a paragraph separator. That *might* allow some simplification of this logic (that was the motivation for the change), but I'm not sure.

cmyr:
That's possible, but I'm happy for now; ideally we won't touch this again until the next cycle, when we will throw it all away.

PR #380, opened 2020-12-28
This updates piet-web and piet-common to v0.3.1

cmyr:
This unbreaks druid-web stuff.


raphlinus (approved):
Seems reasonable. I guess traditionally we've had all backends on the same point? But point shouldn't break anything, so I don't see any problems.

cmyr:
I'm not sure if we've ever even bothered with a point release before? This is mostly just me being lazy, but it also seems like less churn.

PR #381, opened 2020-12-31
re-add Send + Sync to dyn Error.

derekdreery:
This got lost when `ImageBuf` got moved from `druid-shell`.


raphlinus (approved):
Thanks for this. A new Rust release is always "fun" due to clippy changes.

PR #385, opened 2021-01-06
This adds a ptr_eq method for ImageBuf

arthmis:
- this method will be used to implement the Data trait in druid for ImageBuf like this:
```rust
impl Data for ImageBuf {
    fn same(&self, other: &Self) -> bool {
        self.ptr_eq(other)
    }
}
```

### DIFF
--- a/piet/src/image.rs
+++ b/piet/src/image.rs
@@ -137,7 +137,7 @@ impl ImageBuf {
             .unwrap()
     }
 
-    /// Checks whether two ImageBufs point to the same image data in memory.
+    /// Returns `true` if the two `ImageBuf`s refer to the same memory location.
     pub fn ptr_eq(&self, other: &ImageBuf) -> bool {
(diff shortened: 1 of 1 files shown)


piet/src/image.rs
```diff
@@ -136,6 +136,11 @@ impl ImageBuf {
         ctx.make_image(self.width(), self.height(), &self.pixels, self.format)
             .unwrap()
     }
+
+    /// Checks whether two ImageBufs point to the same image data in memory.
```
cmyr, line 140:
```suggestion
    /// Returns `true` if the two `ImageBuf`s refer to the same memory location.
```

cmyr (approved):
great, one doc nit because I'm in that kind of mood, thanks for the PR though!

for the time being in druid we can still impl `Data` for this type just as you did in your other code, using `raw_pixels_shared()` directly.

arthmis:
Ok yeah, I was wondering about that, but I figured it would still be useful to add this method anyways.

Issue #222, opened 2020-06-29
Cairo backend extremely slow text hit test 


ForLoveOfCats:
On wimpy system the following amount of text is enough to make dragging selection slow enough to lock up the UI for seconds at a time (due to performing a hit test on every mouse movement event).
![image](https://user-images.githubusercontent.com/30666851/86054117-91278a80-ba27-11ea-8ae9-2b5c8320723d.png)

Here is a `perf` recording showing that the issue resides within `piet_cairo`'s hit test functionality
![image](https://user-images.githubusercontent.com/30666851/86054027-69d0bd80-ba27-11ea-8cc3-d4bcfb97ec79.png)

I'm aware that this entire codepath will soon/eventually die once we no longer use Cairo's toy text API for this backend but even if this is not fixed before that happens it is good to at least have it documented as an issue

stuaxo:
Is it worth using cairos own hit testing with cairo_in_fill or cairo_in_stroke ?

ForLoveOfCats:
Fixed by https://github.com/linebender/piet/pull/389, https://github.com/linebender/piet/pull/425

PR #11, opened 2019-01-08
Use full names for the RenderContext's associated types

rylev:

PR #389, opened 2021-01-22
Replace Cairo Toy Text with Pango

ForLoveOfCats:
Closes https://github.com/linebender/piet/issues/222
These are initial changes to enable almost all of the Piet rich text API on Linux. It is *not* yet ready for merging. Rather than go the route of using skribo or even interfacing with HarfBuzz directly I decided to try to use Pango. For as long as we are pulling in Cairo then pulling in Pango as well seems reasonable.

At the time of submission two unit tests fail and the snapshots need to be updated.
 * eol_hit_testing: This test is currently failing to get the expected value when hit testing a point at a soft line-break. I started to block in a workaround before I realized that I do not know enough about affinity to do so correctly.
 * empty_layout_size: This test is failing on my system as oddly enough the non-empty layout is ~1.2 units shorter than the empty layout.
 * Snapshots: I ran the snapshot comparison locally and of course it failed as this is an entirely different text stack and actually shapes the non-trivial text in test-7 ect. Once the eol_hit_testing test has been fixed someone with more expertise than me will need to look over test-11 before updating the snapshots

As I mentioned this is not the entire Piet rich text API. This does not (yet?) support `Text::load_font()` as Pango does not appear to support loading fonts from memory.

~~Unfortunately Pango makes the distinction between Left/Right alignment unlike our Start/End so that is not going to respect right-to-left scripts :(~~

There is an `unsafe` block in `CairoTextLayout::update_width()` that I'm not very happy about. I'm going to need to write up a PR to expose that upstream which we can use once released but this should suffice in the meantime.

That is all the notes I can think of at the moment. I fully expect to have overlooked some details and as mentioned there are a number of issues to resolve before merging so I don't expect to have this merged in overnight. However I am very excited at the prospect of having reasonable text support on Linux :)


piet-cairo/src/text.rs
```diff
@@ -210,235 +357,149 @@ impl TextLayout for CairoTextLayout {
...
+                        let next_byte = self.text.as_bytes().get(end_offset);
+                        match next_byte {
+                            Some(b'\r') | Some(b'\n') => {
+                                end_offset += 1;
+                            }
+
+                            _ => break end_offset,
+                        }
+                    }
+                };
```
cmyr, line 459: We don't want to take *all* the paragraph separators; two in a row should be two separate lines. What we should do here is check just the end of the line, something like:

```rust
let end_offset = match self.text.as_bytes()[end_offset..] {
    &[b'\r', b'\n', ...] => end_offset + 2,
    &[b'\r', ...] | &[b'\n', ...] => end_offset + 1,
    _ => end_offset,
```

(I'm not sure if my range matching syntax is quite right, but something like this)

piet-cairo/src/text.rs
```diff
@@ -210,235 +357,149 @@ impl TextLayout for CairoTextLayout {
...
+                let ink_rect = line.get_extents().0;
+                let logical_rect = line.get_extents().1;
+
+                if logical_rect.width > widest_logical_width {
+                    widest_logical_width = logical_rect.width;
+                }
+
+                let line_text = &self.text[start_offset..end_offset];
+                let trimmed_len = line_text.trim_end().len();
+                let trailing_whitespace = line_text[trimmed_len..].chars().count();
```
cmyr, line 470: this should be in utf-8 code-units, so just `.len()` instead of `.chars()`.

cmyr: Also if there's not test for this I'd want to really double check that what we get back is inclusive of trailing whitespace and not of trailing newlines.

piet-cairo/src/text.rs
```diff
@@ -210,235 +357,149 @@ impl TextLayout for CairoTextLayout {
...
-        let middle = left + ((right - left) / 2);
-
-        let grapheme_bounds = match get_grapheme_boundaries(font, text, middle) {
-            Some(bounds) => bounds,
-            None => return HitTestPoint::default(),
-        };
+        let mut line_metrics = Vec::new();
+        let mut y_distance = 0.;
+        let mut widest_logical_width = 0;
+        for line_index in 0..self.pango_layout.get_line_count() {
```
cmyr, line 429: so instead of doing this manually, I think we should use `PangoLayoutIter` to iterate through lines; this will also let you access the actual baseline values, that are exposed on that type.

piet-cairo/src/text.rs
```diff
@@ -210,235 +357,149 @@ impl TextLayout for CairoTextLayout {
...
-        // since it's not a hit, check if closer to start or finish
-        // and move the appropriate search boundary
-        if point.x < grapheme_bounds.leading {
-            right = middle;
-        } else if point.x > grapheme_bounds.trailing {
-            left = middle + 1;
-        } else {
-            unreachable!("hit_test_point conditional is exhaustive");
+                let ink_rect = line.get_extents().0;
+                let logical_rect = line.get_extents().1;
```
cmyr, line 462: to clarify these two things: the logical rect is going to be the ascent + the descent (and maybe also + leading) as the y value, and then the sum of the advances of the runs as the x value.  The *ink rect* is going to be the bounding box of the actual drawn glyphs; this can be different because glyphs can extend outside of their logical bounds.

In any case, neither of these are related to the baseline; for that we need to use the `PangoLayoutIter` interface. :)

cmyr (commented):
overall this looks great. it's nice to delete a bunch of old code, and it's nice to see that the implementation is relatively simple. Went over this with raph and we gave it a pretty thorough review. There are a few small logic errors which I've noted below; hopefully fixing these will help resolve some of the test failures and other questions; for instance sample `#11` is drawing hits in the wrong position because we're incorrectly calculating the baseline position.

Other things you mention:

- `eol_hit_testing` test: if a line is wrapped, the offset of the linebreak corresponds to two different positions in the layout; it can either be the end of the first line, or the beginning of the second. Which position you want to use is generally a function of what user action moved the cursor to this offset. We don't keep track of that, and so for now we just always use 'downstream' affinity, and resolve the hit-test to the beginning of the second line. It is totally reasonable to special-case this.
- `empty_layout_size`: this is another special case; in the case that there is no text in the layout, we would still like to know how tall the layout would be if it did have text, so that we can do things like appropriately size an textbox based on the selected font. If necessary you should pick a default value based on either the reported extents of the default font, or else by just taking the default font size and multiplying by 1.2, which is a reasonable guess for the actual height of the font.
- Snapshots: totally natural that this would fail. When you have this patched up I would suggest just visually comparing the produced snapshots with the snapshots from the other backends, and looking for any major discrepancies; I'll do the same when you PR the updated snapshots.
- Unsafe: you might be able to avoid this by using the [`LayoutIter`](http://gtk-rs.org/docs/pango/struct.LayoutIter.html) interface, which exposes a 'get_index` method for returning the current cursor position; I think your current solution is fine, though, and I think an upstream PR also makes sense.
- Font loading: raph thinks this is possible but might be a bit tricky. I think @derekdreery might have been looking into this as well ([see zulip](https://xi.zulipchat.com/#narrow/stream/259397-piet/topic/Plan.20for.20linux.20text/near/213813067)); we'll need to load the fonts ourselves, and then figure out how to get them into a font collection. This can definitely be future work, though.

In any case, I think this is pretty close to being ready to go, and that's great to see!

piet-cairo/src/text.rs
```diff
@@ -89,29 +112,91 @@ impl fmt::Debug for CairoText {
...
-    pub(crate) fn new(family: FontFamily) -> Self {
-        CairoFont { family }
-    }
+impl CairoTextLayoutBuilder {
+    fn build_pango_attribute(attribute: AttributeWithRange) -> PangoAttribute {
+        let mut pango_attribute = match &attribute.attribute {
+            TextAttribute::FontFamily(family) => {
+                let family = family.name();
+                PangoAttribute::new_family(family).unwrap_or_else(|| {
+                    PangoAttribute::new_family(FontFamily::SERIF.name())
```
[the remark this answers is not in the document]
ForLoveOfCats: I figured that `FontFamily::SERIF.name()` is "nicer" in the big picture than `"serif"`. Or are you referring to something else here?

piet-cairo/src/text.rs
```diff
@@ -89,29 +112,91 @@ impl fmt::Debug for CairoText {
     }
 }
 
-impl CairoFont {
-    pub(crate) fn new(family: FontFamily) -> Self {
-        CairoFont { family }
-    }
+impl CairoTextLayoutBuilder {
+    fn build_pango_attribute(attribute: AttributeWithRange) -> PangoAttribute {
```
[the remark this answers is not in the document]
ForLoveOfCats: That seems reasonable!

piet-cairo/src/text.rs
```diff
@@ -123,8 +208,31 @@ impl TextLayoutBuilder for CairoTextLayoutBuilder {
         self
     }
 
-    fn alignment(self, _alignment: piet::TextAlignment) -> Self {
-        // TextAlignment is not supported by cairo toy text.
+    fn alignment(self, alignment: TextAlignment) -> Self {
+        //TODO: This will not respect Start/End of right-to-left scripts
```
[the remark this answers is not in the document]
ForLoveOfCats: Ayyy, that's great news! I don't quite like how it causes Left/Right to be swapped in interpretation instead of having them be called Start/End but I can live with that. I'll add a note here to explain that behavior

piet-cairo/src/text.rs
ForLoveOfCats: Oops, that's a little embarrassing :sweat_smile: 

So trailing whitespace should not include newline characters, that does make sense

raphlinus: My understanding is that the return value *should* return up to one trailing newline sequence. This behavior is derived from [DWrite](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/ns-dwrite-dwrite_line_metrics), and, while we can make arguments about whether it would be better to include or exclude, it is important to be consistent.

piet-cairo/src/text.rs
```diff
@@ -71,14 +78,30 @@ impl Text for CairoText {
     }
 
     fn load_font(&mut self, _data: &[u8]) -> Result<FontFamily, Error> {
+        /*
+         * NOTE(ForLoveOfCats): It does not appear that Pango natively supports loading font
+         * data raw. All online resource I've seen so far point to registering fonts with
+         * fontconfig and then letting Pango grab it from there but they all assume you have
+         * a font file path which we do not have here.
+         * See: https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/12
+         */
```
derekdreery, line 87: It might be possible to load a font with freetype then hand off an `FT_Face` to pango. Not sure but possibly

derekdreery (commented):
Looks great, looking forward to playing with it :)

piet-cairo/src/text.rs
derekdreery: Looking some more, it seems pango really wants to push you down the path of using fontconfig. I [found 1 example](https://stackoverflow.com/questions/10722577/loading-a-font-from-a-file-to-pango) of manually creating a pango font description from a FcFace.

piet-cairo/src/text.rs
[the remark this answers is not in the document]
cmyr: What I mean is that as written I think this only makes sense if Pango recognizes "serif" as a generic font family and finds some serif font. If it doesn't do that, we may have to go ahead and figure out what the concrete fonts should be that correspond to the generic families.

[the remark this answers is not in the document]
ForLoveOfCats: Pango does support that (or at least appears to on my system). This is for `new_unchecked` in the case that the provided family name cannot be found I figured it would be better to fallback to *something* and still render text instead of just panicking.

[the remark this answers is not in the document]
cmyr: Does pango require us to set a font? I would expect it to try and find a fallback if no font was set. In any case I would expect the absolute fallback be a sans-serif, since that's the default for most (all) UI elements. 🤷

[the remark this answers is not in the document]
ForLoveOfCats: That's a very good point :)

piet-cairo/src/text.rs
ForLoveOfCats: I was intending on being a bit more graceful of handling stuff like out of order CRLF ect but blank lines totally slipped my mind.

Sidenote: I am constantly amazed by the expressiveness of pattern matching

piet-cairo/src/text.rs
ForLoveOfCats: I swapped the `.chars().count()` for a `.len()` as the `end_offset` logic above will ensure that the first newline sequence for that line are included in the trailing whitespace

piet-cairo/src/text.rs
ForLoveOfCats: My logic was that because the Pango origin is in the upper left corner that adding the ink height to the top of the line would give the bottom line of the glyphs which is what understood baseline to be. Good to know that Pango has a mechanism for handling this calculation directly :)

cmyr: Fair enough. The way to think about 'baseline' is that it's the line you would be writing on if you were writing by hand in a ruled notebook. The descenders (on g, y, p, etc.) go below the baseline, and everything else is above it.

ForLoveOfCats: I've updated it to use the baseline information provided by `LayoutIter` but I'm not seeing any noticable change in snapshot 11 as a result.

ForLoveOfCats: Looking at it with my untrained eye shows some stuff wrong from the same behavior causing `eol_hit_testing` to fail but the outside hits on the bottom are looking very wrong :thinking:

cmyr: Yes it looks like something is still funky with the baseline calculations, I'll pay around a bit...

cmyr: From the docs, it looks like [`get_baseline`](https://developer.gnome.org/pygtk/unstable/class-pangolayoutiter.html#method-pangolayoutiter--get-baseline) returns an offset relative to the origin of the whole layout, not the specific line; our `LineMetrics` struct expects this value to be relative to the `y_offset` of the specific line, so you'll need to subtract that to get the correct value.

ForLoveOfCats: Ah ha, that explains why these results are visually identical to the old results where I was manually adding the y_offset

ForLoveOfCats:
After fixing `eol_hit_testing` and improving the situation around baselines snapshot 11 looks

ForLoveOfCats:
Apologies for the delays, I know they can make reviews much more difficult. I've been rather busy lately

I've fixed `eol_hit_testing` and improved the baseline situation. I wouldn't be surprised if they were still incorrect but snapshot 11 seems to look sane at this point. The test `empty_layout_size` remains broken as the values don't seem to me to logically connect with the font size value itself. Simply rounding the layout's size "fixes" the test by bringing the value within the `assert_close!` tolerance but that feels like a non-solution.

This has been rebased on current master as the update to Rust 1.50 caused some conflicts.

derekdreery:
> Apologies for the delays

Don't apologise, it's awesome that you're doing this. I've got a few PRs open atm because I'm busy at college.

cmyr:
@ForLoveOfCats fwiw the 'empty layout size' thing *is* always going to be a bit of a guess. I'm not sure exactly what you're doing here, but on macOS what we do is just store some placeholder values that we calculate based on the font, and then we use these values as a fallback when the layout is empty: see https://github.com/linebender/piet/blob/a0b7cd32639fe874a6bb61dff7db828136d2e335/piet-coregraphics/src/text.rs#L169

In any case, the purpose of this is to let us give text boxes a reasonable initial height even if they don't contain any text; we mostly just want some sane value.

ForLoveOfCats:
To be fair the value that it is currently producing is indeed sane (28.79 for empty layout height, 30 for non-empty layout height). So if you are happy with that then we can make the assertion slightly more tolerant.

If you could give snapshot 11 a glance over when you are able to. It looks pretty good to my untrained eye but there are some things at the bottom that make me thonk.

cmyr:
@ForLoveOfCats how are you calculating the empty layout height? Can you point that out to me, just so I can double-check?

ForLoveOfCats:
Pango itself is generating the value and currently that is used regardless of layout contents.

 `piet-cairo/src/text.rs` lines 544-549
```rust
        let ink_extent = self.pango_layout.get_extents().0;
        let logical_extent = self.pango_layout.get_extents().1;
        self.size = Size::new(
            ink_extent.width as f64 / PANGO_SCALE,
            logical_extent.height as f64 / PANGO_SCALE,
        );
```

cmyr:
hmm okay that looks slightly wrong; our size should be our logical size, and then the inking rect should use to calculate what we will return from `image_bounds`. 

Not sure if this was explained well earlier, but there are two distinct ways of measuring the size of text; there are the *typographic bounds* (logical bounds in this case) and then there are the *inking bounds*. The typographic bounds are basically how big the text reports itself as being, but the text is allowed to draw outside of this area. A common example is certain combining accents, which might extent way above or below the reported size of the text they are part of; a more common example would be a fancy script font where, say, the stroke on the 'leg' of the Q might extend way off past the reported end of the Q glyph.

In any case I think this is mostly fine as written since we only care about height, and logical height should be good enough, here. We should also be using logical size as our base size, however.

ForLoveOfCats:
I generally knew most of that from the documentation I was reading but at the time that I wrote those specific lines I was trying to be a bit too clever about trailing whitespace than I should have especially when Pango should handle that for us anyway. However a glyph painting outside logical bounds on the left and right is not something I had thought about but makes perfect sense.

Beyond that what are your thoughts on making the assertion slightly more tolerant? Additionally I would like to know how snapshot 11 looks to you once you are able to take a peek.

ForLoveOfCats:
Ok no, using the logical width breaks `trailing_whitespace_width` test because Pango is including whitespace in the overall layout's logical width. I'm gonna take a closer look tomorrow.

cmyr:
Okay so what I think you're going to have to do is to manually figure out what the non-whitespace width of each line is, probably by using [`pango_layout_line_index_to_x`](https://developer.gnome.org/pango/unstable/pango-Layout-Objects.html#pango-layout-line-index-to-x). So for the whole layout you will keep track of the max (logical) width both with and without whitespace; the width *without* whitespace we'll store in size, and then the width that includes whitespace will be stored separately, and only used when we want to calculate the return value for `trailing_whitespace_width`.

cmyr:
I'm okay with making the assert more tolerant, but I'm curious why the value is different in the first place. If pango is reporting an actual height I would expect it to be correct? 😕 

On snapshot 11:

You're right that something is wrong here. Looking at it visually, it appears that the points where are after the end of the layout are being resolved incorrectly; they should resolve to the end of the last line (the left edge) but they're resolving to the start of the last line.

I'm not sure exactly what's going on here. It *could* be an off-by-one, where we're accidentally resolving to the end of the *second* last line, or it could be some other logic error.

Another interesting thing is that in the bottom-left example, the text is being drawn outside of the layout. Given that the origin of these rects is fixed, this probably implies a miscalculation of the layout width (this would actually be explained reasonably well by the use of the inking-width, maybe?)

ForLoveOfCats:
I have a local patch which does roughly that but I need to keep playing with it. On topic of the empty layout height I too do not like the situation, it's quite frustrating when it is so close to being correct but just isn't.

cmyr:
Indeed. Maybe check the situation with various different fonts? One possibility is that there is a `leading` or `lineGap` value that is not being added to the height of the empty layout, but is being added to the height of other layouts? This value would likely be font-specific, so you might see some variation with different fonts? Something to experiment with at least.

ForLoveOfCats:
I've fixed up the layout width to use the logical width but without the whitespace. I have also changed `image_bounds` to use the ink size of the layout.

Additionally I've tested all const-named fonts and checked their empty heights. Empty `SYSTEM_UI` was 1.2001953125 units smaller whereas the rest are all 1.0625 units smaller. I'll defer to you on how to best interpret those values.

I'm going to poke at snapshot 11 some more

cmyr:
@ForLoveOfCats testing here, in a VM, I have identical values for those heights 🤔

ForLoveOfCats:
As in you are getting the same heights for empty vs non-empty layouts (ie the test is passing) or you are getting the same values I got on my system?

cmyr:
The test is passing; I'm getting identical values for all fonts I've tested, and they're all round numbers.

ForLoveOfCats:
Very interesting, looking at the Ubuntu CI shows that many tests are failing whereas on my system `empty_layout_size` is the only test to fail but in the VM that you are checking in that specific test passes :thinking: Do all the tests pass for you?

cmyr:
all of the `piet-common` tests pass. A number of tests in `piet-cairo` itself fail.

cmyr:
This does feel to me like an issue related to the specific font that is being chosen on the different platforms.

cmyr:
Ah: possibly related, possibly not, but we need to replace this line: https://github.com/linebender/piet/pull/389/files#diff-9dd8c7717a5acd0fefd853598bf7e7cf9484a5fee32069189304fee8abbb2f02L70

Here you should be checking the font collection for the existence of the font provided, and only returning a `FontFamily` if one actually exists.

ForLoveOfCats:
Could you point out the line in question via an inline comment? Github is refusing to resolve that link fully as the large diffs are collapsed by default

cmyr:
I can't, because it didn't change in this diff, but it's the `Text::font_family` trait method, L167 or so.

cmyr:
@ForLoveOfCats how's this coming along? It feels like we're very close and I'd like to get it landed, if you're feeling stuck I'm happy to poke away at some of the remaining work?

ForLoveOfCats:
I'm usually busy with classwork near the beginning of the week. I was actually planning on looking into figuring out what actual fonts pango is resolving to tomorrow but I can't guarantee anything of course.

derekdreery:
Can I just say that I don't think this has to be perfect to merge. It just has to be better than what was there before. If it's easier, we could merge it with a few TODOs, create an issue for them, and then complete them separately.

ForLoveOfCats:
I am still working on this but I have been exceedingly busy the past few weeks. I recognize that this is blocking a new release and I would like to get all remaining issues corrected as soon as possible. I say that because there are still a few remaining important issues that I've not had the time to sit down and tackle which has proven frustrating.

cmyr:
@ForLoveOfCats Looking through the samples it feels like this is pretty close; the only funny behaviour I see is in picture 11, where the out-of-bounds hits on the last arabic line are all resolving to the right edge when they should be resolving to the left edge. I really wouldn't be too worried about this, though, happy to open an issue and track it down later.

ForLoveOfCats:
There are still massive issues with the paint location of rtl and bidi text which I am in the process of tracking down which I would like to put a little more time into

cmyr:
Looking through the examples it certainly doesn't look egregious, is there anything in particular you're concerned about?

ForLoveOfCats:
Here is some bidi text I copied off wikipedia being painted in the entirely wrong location (and the alignment doesn't seem to affect this case)
image

cmyr:
BiDi is tricky and isn't something we officially support on any backend, so I'm not hugely concerned by this (althought it would be nice to get it to paint in the correct position, even if it doesn't hit-test correctly) but in any case you can always change this in a patch and it can be a point release, so I'm happy to do it as follow-up? I think we're definitely at a place now where this patch is a major improvement over the toy-text.

ForLoveOfCats:
If you are happy with this as it stands then I'll be more than glad to sign off on it. I have however tracked down the cause of the wrong hit test results on the last line of arabic text so I'd like to poke at that a bit first.

cmyr:
@ForLoveOfCats sounds good, I'm also going to be doing some work in here this week, but I'd like to do a release soon. :)

ForLoveOfCats:
I have figured out the paint offset issue! Hit testing on BiDi text is still weird but everything seems to be painting in the correct locations now. Still need to poke more at the hit tests on the last line of Arabic text but then I will be much happier with the state of this patch :)

cmyr:
I expect hit testing BiDi to be broken everywhere, so no worries about that. :)

ForLoveOfCats:
I believe I have fixed the incorrect hit test results on the last line of Arabic text! Assuming that the tests pass and you sign off on the new hit test results I believe this is ready for updating the snapshots and then merging :)

ForLoveOfCats:
So I've fixed an oversight with the point hit tests which made hit tests on BiDi text way less weird but for any RTL text the issue of hit testing on the very end still remains. When asking Pango for the coords of the text position at the very end of a text (ie `idx == text.len()`) it gives the appearance of assuming that it is LTR text and gives the far right edge instead of respecting the direction of the character that it is trailing. I don't know if this is what is actually happening but it is the best way to describe what I am seeing.

cmyr:
does the hit-test method report whether the hit is "in bounds" or similar? as in, whether it's inside the text?

cmyr:
In any case I'm happy to try to get this in as a checkpoint, and then we can revisit afterwards? This will at least allow us to do a release, there's a bunch of stuff I'd like to get into druid soon. :)

ForLoveOfCats:
The point hit testing does tell us if the point is inside the text or not but the byte index hit test does not. It does however give us the both the leading and trailing edges of the grapheme (if there is one) in the form of and `x` and a `width` (and similar for y). This does not help us with this specific situation however.

If you are happy then I am more than glad to get this in

ForLoveOfCats:
Snapshot update PR is live https://github.com/linebender/piet-snapshots/pull/21

piet-cairo/src/text.rs
```diff
@@ -1,64 +1,162 @@
...
+                PangoAttribute::new_weight(pango_weights[closest_index].1).unwrap()
+            }
+
+            TextAttribute::TextColor(text_color) => {
+                let (r, g, b, _) = text_color.as_rgba();
+                PangoAttribute::new_foreground(
+                    (r * 65535.) as u16,
+                    (g * 65535.) as u16,
+                    (b * 65535.) as u16,
+                )
```
cmyr, line 117: I'd skip the f64 conversion and just do `text_color.as_rgba8` and then `(r as u16 * 256) + (r as u16)`

piet-cairo/src/text.rs
```diff
@@ -134,37 +253,104 @@ impl TextLayoutBuilder for CairoTextLayoutBuilder {
...
+            .into_pango(),
+        );
+
+        for attribute in self.attributes {
+            add_attribute(attribute.into_pango());
+        }
+
+        self.pango_layout.set_attributes(Some(&pango_attributes));
+
+        //NOTE: We give Pango a width of -1 for no wrapping
```
cmyr, line 341: this is documented? In this case I would change above in where we set our default `wrap_width` to `f64::INFINITY`.

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
-                .find(|l| point.y >= l.y_offset && point.y < l.y_offset + l.height)
-                .unwrap();
-            (true, line)
+            /*
+             * NOTE(ForLoveOfCats): The docs specify that a non-zero value for trailing
+             * indicates that the point aligns to the trailing edge of the grapheme. In
+             * that case the value tells us the number of "characters" in the grapheme.
+             */
+
+            let text = self.pango_layout.get_text().unwrap();
```
cmyr, line 412: we should have our own copy of the text? I'd just use that.

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
+            let index = index.try_into().unwrap();
+            let trailing = trailing.try_into().unwrap();
+
+            text[index..]
+                .char_indices()
+                .nth(trailing)
+                .map(|(offset, _)| index + offset)
+                .unwrap_or_else(|| text.len())
```
cmyr, line 420: The pango docs aren't super clear on what they mean by a 'character', but I assume it's a codepoint, in which case this looks right. 👍

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
-        let y_pos = lm.y_offset + lm.baseline;
+        let (metric_index, metric) = self
+            .line_metrics
+            .iter()
+            .enumerate()
+            .find(|(_, metric)| {
+                metric.start_offset <= index
+                    && index <= metric.end_offset - metric.trailing_whitespace
+            })
+            .unwrap();
```
cmyr, line 431: it looks like this will fail if the text position is in the trailing whitespace, which is probably possible? In the coretext backend we have a helper fn for this:

```rust
fn line_number_for_utf8_offset(&self, offset: usize) -> usize {
        match self
            .line_metrics
            .binary_search_by_key(&offset, |lm| lm.start_offset)
        {
            Ok(line) => line,
            Err(line) => line.saturating_sub(1),
        }
    }

```

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
 
-        // Then for the line, do text position
-        // Trailing whitespace is removed for the line
-        let line = &self.text[lm.range()];
-        let line_position = idx - lm.start_offset;
+        //NOTE: Manually move to start of next line when hit test is at end of line
+        let index = {
+            let text = self.pango_layout.get_text().unwrap();
```
cmyr, line 435: ditto just using our text.

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
+            if index == text.len() || matches!(text.as_bytes()[index], b'\r' | b'\n') {
+                index
+            } else {
```
cmyr, line 438: Does this match the comment on line 435? I'm a bit confused.

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
+                let mut iterator = GraphemeCursor::new(index, text.len(), true);
+                let next = iterator
+                    .next_boundary(text.as_str(), 0)
+                    .unwrap_or(Some(index))
+                    .unwrap_or(index);
+
```
cmyr, line 444: ditto a bit confused here. As I read the docs, the index returned by pango is guaranteed to fall on a grapheme boundary. I'm sure I'm missing something, but I'm not sure what it is.

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
+            let pango_width = new_width * pango::SCALE as f64;
+            self.pango_layout.set_width(pango_width as i32);
         } else {
-            unreachable!("hit_test_point conditional is exhaustive");
+            /*
+             * NOTE: -1 is the default value, however `update_width` *could*
+             * be called any number of times with different values so we need
+             * to make sure to reset back to default whenever we get no width
+             */
+            self.pango_layout.set_width(-1);
```
cmyr, line 501: let's add a `const UNBOUNDED_WRAP_WIDTH: i32 = -1`; or something.

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
 
-    if text_position == 0 {
-        return 0.0;
-    }
+        let ink_extent = self.pango_layout.get_extents().0;
+        let logical_extent = self.pango_layout.get_extents().1;
+        self.size = Size::new(
+            widest_whitespaceless_width as f64 / PANGO_SCALE,
+            logical_extent.height as f64 / PANGO_SCALE,
+        );
```
cmyr, line 574: you can write this as `Size::new(x, y) / PANGO_SCALE`, which might be slightly cleaner?

piet-cairo/src/text.rs
```diff
@@ -210,233 +396,198 @@ impl TextLayout for CairoTextLayout {
...
 
-fn scale_matrix(scale: f64) -> Matrix {
-    Matrix {
-        xx: scale,
-        yx: 0.0,
-        xy: 0.0,
-        yy: scale,
-        x0: 0.0,
-        y0: 0.0,
+impl fmt::Debug for CairoTextLayout {
```
cmyr, line 588: If I'm going to impl this I might put a bit more info in it? line count, original text, I could even imagine printing each line's text and metrics... 🤷

piet-common/tests/text.rs
```diff
@@ -99,7 +99,7 @@ fn empty_layout_size() {
     assert_close!(
         empty_layout.size().height,
         non_empty_layout.size().height,
-        1.0
+        2.0
```
cmyr, line 102: can you add a //FIXME in here to revisit empty layout size in cairo?

cmyr (approved):
Okay a batch of notes but nothing crazy.

I strongly suspect that when we start using this in druid we're going to run into a bunch of problems, have you played around with patching it in, there?

piet-cairo/src/text.rs
ForLoveOfCats: Pango likes using -1 for "no width/height", I'll make this comment slightly better
See: https://gtk-rs.org/docs/pango/struct.Layout.html#method.set_width

piet-cairo/src/text.rs
ForLoveOfCats: I honestly don't remember if I thought it up or if it was derived from some earlier code. Either way that looks more reasonable

piet-cairo/src/text.rs
ForLoveOfCats: Good point

piet-cairo/src/text.rs
ForLoveOfCats: There are several such impls of `fmt::Debug` but they are all inherited from the existing code

piet-cairo/src/text.rs
ForLoveOfCats: With the way the values we get are this shouldn't be a problem by my understanding. Using the text from `eol_hit_testing` "AA AA\nAA" 
This produces three line metrics with start ends [(0, 3), (3, 6), (6, 8)]
The end of the first metric is at
```
012345678
AA AA AA
   ^
```
So once the trailing whitespace is subtracted you get
```
012345678
AA AA AA
  ^
```
The same applies for the end of the second metric. The third metric however has no trailing whitespace so it produces the expected end at
```
012345678
AA AA AA
        ^
```
Let me know if I am massively misinterpreting this

piet-cairo/src/text.rs
ForLoveOfCats: This has been simplified and comments improved

piet-cairo/src/text.rs
ForLoveOfCats: Yeah this was wrong for a multitude of reasons. I've improved the logic which also happens to have fixed a bug I didn't even know existed
Before
image
After
image

piet-cairo/src/text.rs
ForLoveOfCats: I've improved the logic around this and improved the comment to be more descriptive of what the goal of that section is

piet-cairo/src/text.rs
ForLoveOfCats: I personally find this more clear but more importantly written the other way it would still need to be split over multiple lines which makes the division a bit weird (this is how rustfmt writes it)
```rust
Size::new(
    widest_whitespaceless_width as f64,
    logical_extent.height as f64,
) / PANGO_SCALE;
```

ForLoveOfCats:
I've been firing up Druid with this every once and a while ever since I first got it displaying text. However I hadn't done it recently so I just played around with it for a few minutes. The empty layout height issue *is* noticeable where a TextBox will grow by a few pixels when the first character is inserted but beyond that nothing stands out. All that said I will not be surprised when bugs start being filed :sweat_smile:

piet-cairo/src/text.rs
cmyr: @ForLoveOfCats this is my one remaining question I think!

ForLoveOfCats: This is just to move the index forward by one grapheme. The goal is that in the case of hitting on the end of a soft break it should move forward to the beginning of the next line in accordance with the behavior that the tests appear to desire.

cmyr: ooooh.

This is about "affinity", then? The position where the soft break occurs has two graphical positions in the layout; the 'upstream' location is the end of the first line, and the 'downstream' location is the start of the second. Druid doesn't currently represent affinity, which means that we always resolve a text position to the 'downstream' location.

This shouldn't involve needing to change the offset, though, because the same offset represents both positions.

In that graphical test, we're using both `hit_test_point` (to get a text position) and then `hit_test_position` to actually find the position in question, in order to draw the cursor. You shouldn't need to change `hit_test_point` to make that work; it should be just a matter of telling `hit_test_position` that you always want the downstream result.

Looking through pango docs, the relevant call seems to be [`index_to_line_x`](https://gtk-rs.org/docs/pango/struct.Layout.html#method.index_to_line_x); if you pass this the index of a boundary, you should get different results based on what argument you pass to `trailing`. I'm not sure how you're supposed to use the result, but I haven't dug too deeply. Maybe it's through the line object, with [`index_to_x`](https://gtk-rs.org/docs/pango/struct.LayoutLine.html#method.index_to_x)?

let me know if any of this doesn't make sense. I would be slightly worried that advancing the cursor here could cause a bug (that would visually look very similar to the bug above, where the hit-tests were off-by-one?)

ForLoveOfCats: That's not the issue that this is solving. Pango seems to always give us the index of the last character in the line when we hit on or past it which means we get the cursor position just to the left of it. This logic moves that index to the right of that last character. Without it the assertion on line 179 of `piet-common/tests/text.rs` fails with an index of 2 when it expected 3.

I'm not confident about skipping this when the character in question is a hard newline but that is what the tests seem to require such as line 190 of the same file (and same test) where the line in question contains only a lone `\n` and the test expects the index of that newline character.

cmyr:
@ForLoveOfCats to make this more fun for us, it looks like CI has gone and been updated, which could cause unrelated snapshot failures (slightly different default fonts, different version of pango/cairo, etc)

ForLoveOfCats:
Yeah, I am at the end of my rope with the CI. I was attempting to use a local Ubuntu VM and that is a *whole* mess of frustration and issues with that pile of broken things. I've rolled back the linebreaking settings changes and I'm going to trigger it to run again.

cmyr:
@ForLoveOfCats yea, that sounds smart; get us back to the checkpoint that was passing, and update the snapshot with the images that are produced at that point.

piet-cairo/src/text.rs
ForLoveOfCats: Resolving this conversation as it was misinterpreting the actual problem that this snippet was attempting to solve, more discussion has taken place on Zulip, and the attached piece of the diff is now outdated

piet-cairo/src/text.rs
```diff
@@ -210,233 +401,219 @@ impl TextLayout for CairoTextLayout {
...
+            let last_metric = metric_index + 1 >= self.line_metrics.len();
+            let hard_break = matches!(text.as_bytes(), [.., b'\r', b'\n'] | [.., b'\n']);
+            let wrong_index = index == metric.end_offset.saturating_sub(1);
+
+            if contains_text && !last_metric && !hard_break && wrong_index && outside_x {
+                metric.end_offset
+            } else {
+                index
+            }
+        };
```
ForLoveOfCats, line 477: As mentioned on Zulip I've taken another thorough crack at this issue and I would like a visual once over before I start poking at the tests again

ForLoveOfCats: I've just realized that this doesn't account for `self.pango_offset.x`, I will correct that

PR #391, opened 2021-02-03
Support individual corner radii for kurbo::RoundedRect

SecondFlight:
image

This PR adds support for [this Kurbo patch](https://github.com/linebender/kurbo/pull/166). This PR points directly at my Kurbo fork, so it probably shouldn't be merged until the Kurbo PR is merged and a new Kurbo version is published.

`piet-coregraphics` and `piet-cairo` both get this feature for free, because both platforms use paths for rounded rectangles. I have tested both platforms and verified this.

`piet-web` should get this feature for free as well, for the same reason. I have not explicitly tested this, but I feel confident about it since `to_path` is well-tested and is not platform-specific, and RoundedRect is not mentioned in the code.

In `piet-svg`, this has been implemented by simply using a bezier curve if the corner radii aren't all equal, and a rounded rectangle if they are. Rounded rectangles in SVG can only have a single radius, and while it's possible to use some clipping magic, I don't think it's worth the effort.

~~In `piet-direct2d`, this has been implemented by drawing four rounded rectangles, one for each corner. They are clipped using an [axis-aligned clip](https://docs.microsoft.com/en-us/windows/win32/direct2d/how-to-clip-with-axis-aligned-rects), which is faster than masking with a shape. If all corner radii are equal, a single rounded rectangle is drawn.~~

`piet-direct2d` has a similar implementation to `piet-svg`.

### DIFF
--- a/piet-direct2d/src/lib.rs
+++ b/piet-direct2d/src/lib.rs
@@ -123,7 +123,10 @@ fn geometry_from_shape(
         .filter(|r| r.radii().as_single_radius().is_some())
     {
         Ok(d2d
-            .create_round_rect_geometry(round_rect.rect(), round_rect.as_single_radius().unwrap())?
+            .create_round_rect_geometry(
+                round_rect.rect(),
+                round_rect.radii().as_single_radius().unwrap(),
+            )?
             .into())
     } else if let Some(circle) = shape.as_circle() {
         Ok(d2d.create_circle_geometry(circle)?.into())
--- a/piet-direct2d/src/lib.rs
+++ b/piet-direct2d/src/lib.rs
@@ -439,22 +439,24 @@ impl<'a> D2DRenderContext<'a> {
 
         // TODO: do something special (or nothing at all) for line?
         if let Some(rect) = shape.as_rect() {
-            self.rt.fill_rect(rect, &brush);
-            return;
-        } else if let Some(round_rect) = shape.as_rounded_rect() {
-            if let Some(radius) = round_rect.radii().as_single_radius() {
-                self.rt.fill_rounded_rect(round_rect.rect(), radius, &brush);
-                return;
-            }
+            self.rt.fill_rect(rect, &brush)
+        } else if let Some(round_rect) = shape
+            .as_rounded_rect()
+            .filter(|r| r.radii().as_single_radius().is_some())
+        {
+            self.rt.fill_rounded_rect(
+                round_rect.rect(),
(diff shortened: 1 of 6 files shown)


cmyr:
Will give this a look through shortly, and will do a release when we land this + #389 (and maybe #387?)

raphlinus:
> Will give this a look through shortly, and will do a release when we land this + #389 (and maybe #387?)

I'd love to see #389 go in, though I consider it pretty high risk. I don't think #387 is ready, as there are serious unresolved questions in the Direct2D port in particular (and those are not just implementation details, but point to deeper architectural concerns imho).

SecondFlight:
The most recent commit addresses a major oversight when rendering in Direct2D. See the top-middle here:
image

versus here:
d2d-test-6

Sample image 6 has also been updated. I can get screenshots on Ubuntu and Windows, but I don't have easy access to macOS so that may take me a bit.

piet-direct2d/src/lib.rs
```diff
@@ -430,17 +439,22 @@ impl<'a> D2DRenderContext<'a> {
...
+            return;
         } else if let Some(round_rect) = shape.as_rounded_rect() {
-            self.rt.fill_rounded_rect(round_rect, &brush)
-        } else if let Some(circle) = shape.as_circle() {
-            self.rt.fill_circle(circle, &brush)
-        } else {
-            match path_from_shape(self.factory, true, shape, fill_rule) {
-                Ok(geom) => self.rt.fill_geometry(&geom, &brush, None),
-                Err(e) => self.err = Err(e),
+            if let Some(radius) = round_rect.radii().as_single_radius() {
```
cmyr, line 445: I think we can keep this as before, just changing,

```rust
if let Some(round_rect) = shape.as_rounded_rect().filter(|r| r.as_single_radius().is_some()) {
}
```

This isn't a major point, but it cleans up the diff somewhat.

cmyr (approved):
One note that is totally not a blocker, otherwise this looks good to me. I've merged the piet-snapshots PR, so you should be able to update the submodule in this PR and then we'll pass CI (except for this funny wasm failure that I don't think is your fault..)

SecondFlight:
That was just sloppy, sorry. I forgot that building and testing `piet` doesn't build or test `piet-direct2d` or etc.

Now the snapshots are still failing. Not sure what's going on, since they're freshly updated. I'll poke at this more when I get time next, but that may be a couple days.

cmyr (approved):
Looks good!

PR #392, opened 2021-02-08
Avoid clone in web backend

JAicewizard:
This avoids cloning the image when painting.


piet-web/src/lib.rs
```diff
@@ -270,45 +270,44 @@ impl RenderContext for WebRenderContext<'_> {
         let canvas = element.dyn_into::<HtmlCanvasElement>().unwrap();
         canvas.set_width(width as u32);
         canvas.set_height(height as u32);
-        let mut buf = match format {
-            // Discussion topic: if buf were mut here, we could probably avoid this clone.
-            // See https://github.com/rustwasm/wasm-bindgen/issues/2364 for the issue.
-            ImageFormat::RgbaSeparate => buf.to_vec(),
+        let mut new_buf: Vec<u8>;
```
cmyr, line 273: I'm confused that this will compile when this value isn't initialized in all branches?

cmyr (approved):
I'm a bit confused this works, but I guess the compiler is smart enough to figure out that the `new_buf` is only ever used when it is initialized? In any case I agree that saving the allocation makes sense.

To me the more idiomatic way to handle this would be to have the `match` statement evaluate to a `Cow<[u8]>`; the branches that allocate would return `Cow::Owned` and the branch that just passes through the input would return `Cow::Borrowed`.

In any case that's mostly academic, so I'm happy to merge this as written. Thanks!

piet-web/src/lib.rs
JAicewizard: the compiler only checks branches where `new_buf` is used, and all of those are handled properly.  (I think thats how it works)

JAicewizard:
I think the `Cow` would obfuscate the code more then necessary, so ill just leave it like this.

PR #393, opened 2021-02-10
[WIP] Piet skia

pum-purum-pum-pum:
Adds skia backend to piet

Current TODO list
* fix paragraph/simple text font inconsistency (for some reason in skia Paragraph and draw_string uses different defaults for fonts)
~*  "XI" image logo example not working (I found it in my druid fork which works with skia)~
* add snapshots
* clean up code and commits history
* use skia crate (currently it's also fork that is not fully upstream yet)

Also useful links: druid fork with winit for testing this backend https://github.com/pum-purum-pum-pum/druid


cmyr:
@pum-purum-pum-pum I'm happy to treat this as a "second tier" backend, and not include snapshots? This is how (for instance) the svg backend works. You *should* still have an example program that can run the sample images, though; you can probably just adapt one from another backend.

I think we'll also want to add this to `piet-common`, although we'll need to figure out how to make it selectable; a feature flag would work but we tend to have a bad time when we try to use a feature to toggle between two implementations...

pum-purum-pum-pum:
This is most likely postponed until Skia will have a new release. Because skia-safe fork that I'm using currently relies on the master branch Skia.

cmyr:
@pum-purum-pum-pum sounds good, please ping me in a comment when this is ready for a look!

PR #394, opened 2021-02-11
bump core-graphics to 0.22.2

dzhou121:
There is an issue on core-graphics backend that RenderContext.blurred_rect can cause memory leak because of an issue that was fixed by https://github.com/servo/core-foundation-rs/commit/dc588497db727a3ba943aaecc83d2a07dc6d6b0a


raphlinus:
I think you can rebase this against current master to fix the CI problems (caused by the Rust 1.50 release, fixed in #395). Overall it looks fine, but I do obviously want to see green tests before approving.

cmyr (approved):
Thanks!

PR #396, opened 2021-02-12
Fix alpha on cairo

JAicewizard:
Make sure we use the alpha values in fill on linux. This is needed to properly support transparency on gtk


raphlinus (approved):
This looks good to me. I think we might need some clarification in the docs about what clearing with non-opaque values means (as it might not necessarily be the same across all back-ends), btu this seems reasonable. Thanks!

PR #402, opened 2021-02-19
implement new clipping for cairo

JAicewizard:


JAicewizard:
cairo-test-15

JAicewizard:
white = transparent, green corner is semi-transparent, and the clipping works

JAicewizard:
cairo-test-15

piet-cairo/src/lib.rs
```diff
@@ -68,15 +68,33 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
...
+            if let Some(region) = region.into() {
+                rc.clip(region)
+            }
+            //we also need to save/restore the operator, so we can apply the SOURCE operator
+            let op = rc.ctx.get_operator();
+            rc.ctx.set_operator(cairo::Operator::Source);
+            rc.ctx.paint();
+            rc.ctx.set_operator(op);
+            println!("{:?}",matrix);
+            rc.ctx.set_matrix(matrix);
```
cmyr, line 95: we shouldn't need to do this, because it will be reset with the `restore` at the end of this save operation.

piet-cairo/src/lib.rs
```diff
@@ -68,15 +68,33 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
...
+            );
+            // we DO want to clip the specified region
+            if let Some(region) = region.into() {
+                rc.clip(region)
+            }
+            //we also need to save/restore the operator, so we can apply the SOURCE operator
+            let op = rc.ctx.get_operator();
+            rc.ctx.set_operator(cairo::Operator::Source);
+            rc.ctx.paint();
+            rc.ctx.set_operator(op);
```
cmyr, line 93: I'm not *sure* we need to do this, since the old operator might be saved and restored with the `save` operation? I thought I was testing this in the sample picture (by doing a fill after doing a clear) but I should double check that.

cmyr (commented):
A couple little comments but this looks good!

piet-cairo/src/lib.rs
cmyr: (also we should get rid of the println)

piet-cairo/src/lib.rs
JAicewizard: yes i removed these in the mean time, thanks for pointing them out anyways!

cmyr (approved):
Looks good, let's try this out.

PR #409, opened 2021-03-05
[cg] Fix incorrect line ranges with trailing tabs

cmyr:
The CTLineGetStringRange function I was using for string
ranges apparently doesn't include the range of all trailing
whitespace. This seems to maybe only be a problem for tabs?
In any case, it means we were getting subtly wrong reported
line ranges, which I bumped into with IME.

This fixes that, but just using the start offsets of lines;
we assume a line is the range between two line starts.

### DIFF
--- a/piet-coregraphics/src/ct_helpers.rs
+++ b/piet-coregraphics/src/ct_helpers.rs
@@ -35,7 +35,7 @@ use core_text::{
     line::{CTLine, CTLineRef, TypographicBounds},
     string_attributes,
 };
-use foreign_types::ForeignType;
+use foreign_types::{ForeignType, ForeignTypeRef};
 
 use piet::kurbo::{Affine, Rect};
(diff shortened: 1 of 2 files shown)


cmyr:
Okay, this is a slightly tricker problem than I had initially thought, and there are a few different things going on.

The first thing looks suspiciously like a bug in coretext, although I haven't dug too deeply; but basically laying out text with `CTFrame` seems to break when that text contains some small number (~4) contiguous tabs.

The second thing is a bug in the piet spec. We have been counting trailing whitespace incorrectly; it is not really about counting any and all whitespace at the end of a line, but rather the set of whitespace that is not relevant for layout; a good heuristic for this is "whitespace the absence of which would not change the drawing of a selection that spans multiple lines". This last was an oversight on may part.

With that in mind I'm going to hold of on this for right now until I can figure out what the correct solution is.

PR #412, opened 2021-03-08
Fix incorrect selection rects when emergency breaking

cmyr:
Because we lack affinity, we can't explicitly request
that the trailing hit be upstream, and the downstream
width is the same as our starting width, leading to
a zero-width rect.

This should fix the issue with weird missing lines in selections (reported on zulip, and discussed in https://github.com/linebender/piet/issues/323#issuecomment-793049972)

### DIFF
--- a/piet-common/tests/text.rs
+++ b/piet-common/tests/text.rs
@@ -37,6 +37,22 @@ fn make_factory() -> PietText {
     text
 }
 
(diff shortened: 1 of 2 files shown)


cmyr:
I'll wait for #389 for this, when the test should pass on cairo.

PR #425, opened 2021-03-25
Pango take two

cmyr:
I want to open a new branch for this so that I can force push and not screw up the previous history.

This is still a draft, but I think it's getting close.

PR #428, opened 2021-03-26
Fix trailing whitespace calculation for non-ascii space characters

### LINKED ISSUE #427: Panic when trying to render ideographic space character
Note: This was a bug I observed when building something with druid, but since the traceback led to this library I'm filing it here as that is likely where the issue lies.

When rendering the character U+3000 IDEOGRAPHIC SPACE (　) at the end of a line or by itself, the program panics and this a message similar to this one is printed (this specific example is for a string containing the character alone):

    thread 'main' panicked at 'byte index 2 is not a char boundary; it is inside '\u{3000}' (bytes 0..3) of `　`', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/piet-cairo-0.3.0/src/text.rs:313:45

I'm not sure if this is a bug in piet, cairo, or what have you, but I would appreciate it being looked into.

Majora320:
Closes #427


Majora320:
That's really weird, I tested the latest commit of this locally with wasm-pack and everything passed

cmyr (commented):
Thanks for this but the whole cairo backend is going to be replaced very shortly. I believe this should be handled correctly in the new patch, but I think your test case would be good to include, anyway?

So let's wait until #425 lands, and then we can look into this?

cmyr:
@Majora320 okay the new impl is merged, feel free to try this again, and PR the test in any case!

Majora320:
Hm, looks like the threshold might need to be increased? Every test that's failing here works on my machine; I'm really confused about why the wasm-pack one is still failing especially.

Majora320:
Still fails on the wasm target; I thought it was because perhaps the CI machine doesn't have the proper font installed to display kanji, but if that's the case it doesn't have one for fullwidth characters either (?)
Maybe a CI step to install a font for it?

cmyr:
yes it's very possible that we're missing a font. In any case wasm is not a tier-one target and does not have the full text API, so I think it's fine to just disable this test on wasm for now, lots of text tests are disabled there.

cmyr (approved):
Thanks!

PR #429, opened 2021-03-26
Save RenderContext (take two)

cmyr:
This is a bit of cleanup on top of #387 to get it merged.

- closes #387


cmyr:
Note: on merge we will need to open an issue about the missing cairo implementation.

cmyr:
Going to get this in now; if there are any problems they can be fixed without breaking semvar.

PR #444, opened 2021-05-02
Use absolute index to get whitespaceless width & don't sub x coord

ForLoveOfCats:
It seems that the index argument expected here is relative to the start of the text, not the start of the line. Prior to this change long lines w/trailing whitespace would not correctly affect the widest whitespaceless width as their whitespace start index would index past the first line and into part of the second line resulting in a significantly smaller value.

Addresses https://github.com/linebender/druid/issues/1753

### DIFF
--- a/piet-cairo/src/text.rs
+++ b/piet-cairo/src/text.rs
@@ -548,10 +548,9 @@ impl CairoTextLayout {
             //HACK: This check for RTL is to work around https://gitlab.gnome.org/GNOME/pango/-/issues/544
             let non_ws_width = if trailing_whitespace != 0 && !self.is_rtl {
                 //FIXME: this probably isn't correct for RTL
-                let ws_start = line.index_to_x((start_offset + trimmed_len) as i32, false);
-                ws_start.saturating_sub(logical_rect.x)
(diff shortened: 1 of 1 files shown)


ForLoveOfCats:
Annnnnnd it's broken the Arabic text in snapshot 11

ForLoveOfCats:
Tracked down the issue with the Arabic text. The hit test for the whitespace is always returning zero. I think it is https://gitlab.gnome.org/GNOME/pango/-/issues/544 striking again which is unfortunate.

ForLoveOfCats:
Hesitant workaround for the RTL issue by just using the logical width for any RTL text. Additionally snapshot 11 is still broken but in a good way because the results are more correct. I wasn't even aware that snapshot 11's Arabic text had the wrong width to begin with but now it has a much more correct width.

ForLoveOfCats:
The snapshot changes (just snapshot 11) is up at https://github.com/linebender/piet-snapshots/pull/25. This PR has been updated to point to that and CI is passing as of writing this comment

cmyr (approved):
Cool, I think this is a reasonable workaround for the time being.

Issue #224, opened 2020-07-03
Does CairoRenderContext::make_image() work on big endian systems?


psychon:
Hi,

I just passed by and saw this code:
https://github.com/linebender/piet/blob/1568b6d9689520913a9b4d47d16dd25fe27108c7/piet-cairo/src/lib.rs#L270-L299
This code is wrong. I do not understand it completely, but the swapping of the order of channels (stuff like `data[foo + 0] = buf[bar + 2]`) looks very much like a byte order conversion.
From https://www.cairographics.org/manual/cairo-Image-Surfaces.html#cairo-format-t, `CAIRO_FORMAT_RGB24` is described as:
> each pixel is a 32-bit quantity, with the upper 8 bits unused. Red, Green, and Blue are stored in the remaining 24 bits in that order. (Since 1.0)

This means that in C, if you have a pixel that starts at address `ptr`, the blue channel can be read via `uint8_t blue = *((uint32_t*)ptr)  & 0xff`. This reads an `uint32_t` / `u32` in the native endianness and then gets the lowest byte from that. On little endian, the least significant byte is stored first, so here this is equivalent to `uint8_t blue = *ptr`. On big-endian however, this ends up doing `uint8_t blue = *(ptr + 3);`.

I hope this is understandable.

Perhaps you could also clarify in https://docs.rs/piet/0.1.0/piet/enum.ImageFormat.html that the format is meant to be as "with first byte for the red channel, second byte for green, and third byte for blue" or something like that (to tell it apart from cairo's use of a "32 byte quantity" with its endianness).

Slightly unrelated: https://github.com/linebender/piet/blob/1568b6d9689520913a9b4d47d16dd25fe27108c7/piet-cairo/src/lib.rs#L280
I am not shure how changing the order of the channels would help here. Right now you copy from offsets `(2,1,0,3)`. After swapping the order of channels, you end up with `(3,0,2,1)`. That's not really an improvement. Perhaps `ABGR` is meant by the comment? That would match cairo's ARGB order on little endian systems... well, no... now you got me really confused. Anyway, something is up with that `TODO`.

Edit: Random code link to underline what I mean: https://sources.debian.org/src/gtk+2.0/2.24.32-4/gdk/gdkcairo.c/?hl=211#L238
At the time of writing, this has the following code:
```c
#if G_BYTE_ORDER == G_LITTLE_ENDIAN
	      q[0] = p[2];
	      q[1] = p[1];
	      q[2] = p[0];
#else	  
[...]

PR #448, opened 2021-06-13
[cairo] Fix for big-endian systems

psychon:
Cairo defines pixels as "a 32-bit quantity [...] stored in
native-endian". However, the existing code was treating a pixel as four
8-bit quantities. Put differently, it was hardcoding little endian.

This commit changes the code to first calculate the pixel value as an
u32 and then use u32::to_ne_bytes() to get back to "the world of bytes".
For readability, this is done in a new helper function.

I did not actually check if this fixes anything on big endian, but at
least it does not change anything for little endian according to the
test-pictures examples.
    
Fixes: https://github.com/linebender/piet/issues/224
Signed-off-by: Uli Schlachter <psychon@znc.in>

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -241,13 +241,12 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                 .map_err(|e| Error::BackendError(Box::new(e)))?;
             for y in 0..height {
                 let src_off = y * bytes_per_row;
-                let dst_off = y * stride;
+                let data = &mut data[y * stride..];
                 match format {
                     ImageFormat::Rgb => {
                         for x in 0..width {
                             write_rgb(
-                                &mut data,
-                                dst_off,
+                                data,
                                 x,
                                 buf[src_off + x * 3 + 0],
                                 buf[src_off + x * 3 + 1],
@@ -261,8 +260,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                         // TODO: consider adding BgraPremul format.
                         for x in 0..width {
                             write_rgba(
-                                &mut data,
-                                dst_off,
+                                data,
                                 x,
                                 buf[src_off + x * 4 + 0],
                                 buf[src_off + x * 4 + 1],
@@ -279,8 +277,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                         for x in 0..width {
                             let a = buf[src_off + x * 4 + 3];
                             write_rgba(
-                                &mut data,
-                                dst_off,
+                                data,
                                 x,
                                 premul(buf[src_off + x * 4 + 0], a),
                                 premul(buf[src_off + x * 4 + 1], a),
@@ -292,8 +289,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                     ImageFormat::Grayscale => {
                         for x in 0..width {
                             write_rgb(
-                                &mut data,
-                                dst_off,
+                                data,
                                 x,
                                 buf[src_off + x],
                                 buf[src_off + x],
@@ -520,19 +516,18 @@ fn compute_blurred_rect(rect: Rect, radius: f64) -> (ImageSurface, Point) {
     (image, origin)
 }
 
-fn write_rgba(data: &mut [u8], offset: usize, x: usize, r: u8, g: u8, b: u8, a: u8) {
+fn write_rgba(data: &mut [u8], x: usize, r: u8, g: u8, b: u8, a: u8) {
     // From the cairo docs for CAIRO_FORMAT_ARGB32:
     // > each pixel is a 32-bit quantity, with alpha in the upper 8 bits, then red,
     // > then green, then blue. The 32-bit quantities are stored native-endian.
     let (a, r, g, b) = (u32::from(a), u32::from(r), u32::from(g), u32::from(b));
     let pixel = a << 24 | r << 16 | g << 8 | b;
 
-    let pixel_offset = offset + 4 * x;
-    data[pixel_offset..pixel_offset + 4].copy_from_slice(&pixel.to_ne_bytes());
+    data[4 * x..4 * (x + 1)].copy_from_slice(&pixel.to_ne_bytes());
 }
 
-fn write_rgb(data: &mut [u8], offset: usize, x: usize, r: u8, g: u8, b: u8) {
+fn write_rgb(data: &mut [u8], x: usize, r: u8, g: u8, b: u8) {
     // From the cairo docs for CAIRO_FORMAT_RGB24:
     //  each pixel is a 32-bit quantity, with the upper 8 bits unused.
-    write_rgba(data, offset, x, r, g, b, 0);
+    write_rgba(data, x, r, g, b, 0);
 }
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -245,9 +245,14 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                 match format {
                     ImageFormat::Rgb => {
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x * 3 + 2];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x * 3 + 1];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x * 3 + 0];
+                            write_rgb(
+                                &mut data,
+                                dst_off,
+                                x,
+                                buf[src_off + x * 3 + 0],
+                                buf[src_off + x * 3 + 1],
+                                buf[src_off + x * 3 + 2],
+                            );
                         }
                     }
                     ImageFormat::RgbaPremul => {
@@ -255,10 +260,15 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                         // hope that LLVM generates pretty good code for this.
                         // TODO: consider adding BgraPremul format.
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x * 4 + 2];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x * 4 + 1];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x * 4 + 0];
-                            data[dst_off + x * 4 + 3] = buf[src_off + x * 4 + 3];
+                            write_rgba(
+                                &mut data,
+                                dst_off,
+                                x,
+                                buf[src_off + x * 4 + 0],
+                                buf[src_off + x * 4 + 1],
+                                buf[src_off + x * 4 + 2],
+                                buf[src_off + x * 4 + 3],
+                            );
                         }
                     }
                     ImageFormat::RgbaSeparate => {
@@ -268,17 +278,27 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                         }
                         for x in 0..width {
                             let a = buf[src_off + x * 4 + 3];
-                            data[dst_off + x * 4 + 0] = premul(buf[src_off + x * 4 + 2], a);
-                            data[dst_off + x * 4 + 1] = premul(buf[src_off + x * 4 + 1], a);
-                            data[dst_off + x * 4 + 2] = premul(buf[src_off + x * 4 + 0], a);
-                            data[dst_off + x * 4 + 3] = a;
+                            write_rgba(
+                                &mut data,
+                                dst_off,
+                                x,
+                                premul(buf[src_off + x * 4 + 0], a),
+                                premul(buf[src_off + x * 4 + 1], a),
+                                premul(buf[src_off + x * 4 + 2], a),
+                                a,
+                            );
                         }
                     }
                     ImageFormat::Grayscale => {
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x];
+                            write_rgb(
+                                &mut data,
+                                dst_off,
+                                x,
+                                buf[src_off + x],
+                                buf[src_off + x],
+                                buf[src_off + x],
+                            );
                         }
                     }
                     _ => return Err(Error::NotSupported),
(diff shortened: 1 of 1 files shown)


piet-cairo/src/lib.rs
```diff
@@ -245,20 +245,30 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                 match format {
                     ImageFormat::Rgb => {
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x * 3 + 2];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x * 3 + 1];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x * 3 + 0];
+                            write_rgb(
```
cmyr, line 248: Might it be simpler for us to use `#[cfg(target_endian = "big")]` here and have two distinct codepaths? I'm not totally sure this makes sense, it really comes down to what the generated code looks like, but curious if that's an approach you had considered.

psychon: No, I hadn't really considered that since letting `u32::to_ne_bytes()` deal with endianess seemed like the simplest thing to me. Also, using `#[cfg]` means that this is extra code which might need actual test runs after updating. With the `to_ne_bytes()`-approach, I'd feel more confident at changing the code since most of the code is not endianess-specific.

Anyway, do I understand correctly that you'd prefer something like the following? (No testing done at all - I did not even try building this on little endian)
```diff
diff --git a/piet-cairo/src/lib.rs b/piet-cairo/src/lib.rs
index b97ec16..e2985d7 100644
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -242,12 +242,23 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
             for y in 0..height {
                 let src_off = y * bytes_per_row;
                 let dst_off = y * stride;
+                #[all(not(cfg(target_endian = "big"), cfg(target_endian = "little")))]
+                error("unsupported endian");
                 match format {
                     ImageFormat::Rgb => {
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x * 3 + 2];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x * 3 + 1];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x * 3 + 0];
+                            #[cfg(target_endian = "big")]
+                            {
+                                data[dst_off + x * 4 + 3] = buf[src_off + x * 3 + 2];
+                                data[dst_off + x * 4 + 2] = buf[src_off + x * 3 + 1];
+                                data[dst_off + x * 4 + 1] = buf[src_off + x * 3 + 0];
+                            }
+                            #[cfg(target_endian = "little")]
+                            {
+                                data[dst_off + x * 4 + 0] = buf[src_off + x * 3 + 2];
+                                data[dst_off + x * 4 + 1] = buf[src_off + x * 3 + 1];
+                                data[dst_off + x * 4 + 2] = buf[src_off + x * 3 + 0];
+                            }
                         }
                     }
                     ImageFormat::RgbaPremul => {
@@ -255,10 +266,20 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                         // hope that LLVM generates pretty good code for this.
                         // TODO: consider adding BgraPremul format.
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x * 4 + 2];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x * 4 + 1];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x * 4 + 0];
-                            data[dst_off + x * 4 + 3] = buf[src_off + x * 4 + 3];
+                            #[cfg(target_endian = "big")]
+                            {
+                                data[dst_off + x * 4 + 3] = buf[src_off + x * 4 + 2];
+                                data[dst_off + x * 4 + 2] = buf[src_off + x * 4 + 1];
+                                data[dst_off + x * 4 + 1] = buf[src_off + x * 4 + 0];
+                                data[dst_off + x * 4 + 0] = buf[src_off + x * 4 + 3];
+                            }
+                            #[cfg(target_endian = "little")]
+                            {
+                                data[dst_off + x * 4 + 0] = buf[src_off + x * 4 + 2];
+                                data[dst_off + x * 4 + 1] = buf[src_off + x * 4 + 1];
+                                data[dst_off + x * 4 + 2] = buf[src_off + x * 4 + 0];
+                                data[dst_off + x * 4 + 3] = buf[src_off + x * 4 + 3];
+                            }
                         }
                     }
                     ImageFormat::RgbaSeparate => {
@@ -268,17 +289,36 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
                         }
                         for x in 0..width {
                             let a = buf[src_off + x * 4 + 3];
-                            data[dst_off + x * 4 + 0] = premul(buf[src_off + x * 4 + 2], a);
-                            data[dst_off + x * 4 + 1] = premul(buf[src_off + x * 4 + 1], a);
-                            data[dst_off + x * 4 + 2] = premul(buf[src_off + x * 4 + 0], a);
-                            data[dst_off + x * 4 + 3] = a;
+                            #[cfg(target_endian = "big")]
+                            {
+                                data[dst_off + x * 4 + 3] = premul(buf[src_off + x * 4 + 2], a);
+                                data[dst_off + x * 4 + 2] = premul(buf[src_off + x * 4 + 1], a);
+                                data[dst_off + x * 4 + 1] = premul(buf[src_off + x * 4 + 0], a);
+                                data[dst_off + x * 4 + 0] = a;
+                            }
+                            #[cfg(target_endian = "little")]
+                            {
+                                data[dst_off + x * 4 + 0] = premul(buf[src_off + x * 4 + 2], a);
+                                data[dst_off + x * 4 + 1] = premul(buf[src_off + x * 4 + 1], a);
+                                data[dst_off + x * 4 + 2] = premul(buf[src_off + x * 4 + 0], a);
+                                data[dst_off + x * 4 + 3] = a;
+                            }
                         }
                     }
                     ImageFormat::Grayscale => {
                         for x in 0..width {
-                            data[dst_off + x * 4 + 0] = buf[src_off + x];
-                            data[dst_off + x * 4 + 1] = buf[src_off + x];
-                            data[dst_off + x * 4 + 2] = buf[src_off + x];
+                            #[cfg(target_endian = "big")]
+                            {
+                                data[dst_off + x * 4 + 3] = buf[src_off + x];
+                                data[dst_off + x * 4 + 2] = buf[src_off + x];
+                                data[dst_off + x * 4 + 1] = buf[src_off + x];
+                            }
+                            #[cfg(target_endian = "little")]
+                            {
+                                data[dst_off + x * 4 + 0] = buf[src_off + x];
+                                data[dst_off + x * 4 + 1] = buf[src_off + x];
+                                data[dst_off + x * 4 + 2] = buf[src_off + x];
+                            }
                         }
                     }
                     _ => return Err(Error::NotSupported),
```
>  it really comes down to what the generated code looks like, 

I haven't looked at any assembly and even if I did, I am not sure I'd draw the right conclusions. I guess one would need some benchmarks instead to compare approaches.

Do you want to go down this route? If so, I think it would make more sense to add some benchmarks in a separate PR and merge them first.

cmyr: My only real concern is with ensuring we aren't slowing down the little-endian path, so I'm trying to think through what our options are. 

we might even be able to do something like,

```rust
// somewhere
const fn native_endian_offsets() -> (usize, usize, usize, usize) {
    #[cfg(target_endian = "big")]
    {
        return (3, 2, 1, 0)
    }
    return (0, 1, 2, 3)
}

// before the match arm:
let (r, g, b, a) = native_endian_offsets();

// and then our indexing code:
data[dst_off + x * 4 + r] = buf[src_off + x * 3 + 2];
data[dst_off + x * 4 + g] = buf[src_off + x * 3 + 1];
data[dst_off + x * 4 + b] = buf[src_off + x * 3 + 0];
```

or something? Just thinking out loud.

psychon: I tried looking into this a bit more. Here is a GodBolt-experiment:

https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlJvoCqAZ0wAFAB4gA5AAYppAFZdSrZrVDEArkIKkR7ZATx1KmWugDCqVpoC2tZWfQAZPLUwA5OwCNMxEABMMqQADqhChMa0Vjb2ymERRnSu7l62vv5BepgGSbQMBMzEBDF2DpzZuVEFRQQpnj5%2BgcE6taVxFa3F9WkZzQCUeqiaxMgcUiGa3gDUVLTTtswA1pgA%2BniLwGskeMBubKusuwgEEN6aVCDTAKQBAGzXAKwAQpoAHE8AIqTT6MyFV1ud1smgINxe7y%2BPwA7nh0AQkNNtHgAF6Yfo3ADsz2uMgAgtNCbMSNNxNM3NMZAA6Kmw%2BEILE4/FElm/f7MJ7PMkAKmm3BuAWelK%2BNwAzJ9pucqJyedNRQKhQEvtdRUyCazCX9CjLprz%2BbchZwRSqJVKdbz5QbpkbHp8VWqNZr2ea%2BQrpkrbWLTRcXZbBcLbfbcerCddMXb8WGI3jg5MZnMFss1htmFtVrYMHgqHhMOhDsdTlLAfdOZDbT8tcxi8DQeDXh9y9M6Qirsi0Riww6iTRiKTyfNqbS4QjGcHHU3iIQ1sRgN4IGPx0TK6QF4vSSvmWvJT6XrK/YqoavF2bd7q5W6bd8j%2BOT1yz/uA1fNxr%2BkHn1jo1Hg8GE9DJwRp1nZgIErasQTBUsG2%2BUlWwiNEfn8JE3h%2BUAkJ%2BbxW2Q6YqyQjtsSPdgwQgZgEJQ9CO3FaYIE0UUAhAEAqGIVBbGIwYkVo%2BjGOYiBiDYmi6IYpiWOAPiOME7jvH6V9VQIzAwRCPBxByL1sLFCwVQsd19UxTTew0jTrTuLFNOANSDLeYzJTfEM2W1XduW4GkIHEA1OH6ByvipNAQgAT1WLjbFWIQjjGed7gUpTWCpIhVncVZvB8gChAgKTrK/SN8V/f9ALnUCBRrCCISgn5JCRODMAQzCUKqyVMLwrtCT/KdVhnbxgOXddpjIaZgHQn4ZGktUoykQZWGkR55AcWR5FQaQACVtDBIRhlGTABVFTh5AIaQ5Ck0glhAR5gjGqRuEmnaZukeQhBAYJtumwY4FgJA0FsBT2DICgIFe96mloAtWB8vg8FYADiBus4LtIbx9mIHzpE20hXtsMwCAAeVoQGoawRZ1HYbG8GITBDDwAA3TAbum0hMCU5BQXGRG3AAk65BUPBvGIIofKsLAoYISdbARka%2BDoRgWHxngRcEEQJChpQKlUdQQC0HQ2e8G7IEGVAQjySmAForB1vX2HJ1gTQCa6cmJvJTHMdpymp8wekafwKgSSI6Ht%2BJwg92hnfSJpOitkm6BqYovaDqpQ8Kbo3AaAPXb0GOSmsMplC6Oo496QPBmWkYxi4UbxvOqnZqkFWCGQaZ/uAE5Aao3BCBJW4Np%2BKw3pBvx1rc6YFp0LaLr2g6jpUaQztIKbWbL67btIe7dqLqQLYnqHp7nwfBnJ8GohAbggA%3D

This is just the `ImageFormat::Rgb`-case. `make_image_original_light` is supposed to be the original version and `make_image_modified_light` is supposed to represent the version from this PR.

Both versions are bad.

The original version:
- There is a out-of-bounds-check for every single indexing into the slice, i.e. six bounds check per loop iteration.
- The data is really copied around byte by byte.
- For example, just the line `data[x * 4 + 2] = buf[x * 3 + 0];` results in:
```
        # First bounds check
        cmp     r10, rsi
        jae     .LBB0_16
        # Second bounds check
        cmp     r11, rcx
        jae     .LBB0_17
        # Copy [rdi + r9 - 2] to [rdx + r11]
        movzx   eax, byte ptr [rdi + r9 - 2]
        mov     byte ptr [rdx + r11], al
```
So, we have two length checks. Afterwards, a single byte is copied. It looks like the indexing was rewritten by the compiler and thus the offsets are different. At the end of each loop iteration, we have the following, which seems mostly consistent with the expected loop variables:
```
        add     r11, 4
        add     r9, 3
        add     r8, -1
        jne     .LBB0_2
```

"My" version is slightly better.
- There are only three bounds checks per pixel.
- The data is copied per-byte into registers, then put into `ebp` and written to the destination in a single memory access.
- The whole loop is:
```
.LBB1_2:
        # I guess these are three bounds checks for buf[x*3+2], data[x*4] and data[(x+1)*4], but have not checked throughly
        cmp     rax, rsi
        jae     .LBB1_7
        cmp     r10, r11
        je      .LBB1_8
        cmp     r9, rcx
        ja      .LBB1_9
       # Get the three bytes into ebx, ebp, r14d
        movzx   ebx, byte ptr [rdi + rax]
        movzx   ebp, byte ptr [rdi + rax - 1]
        movzx   r14d, byte ptr [rdi + rax - 2]
        # Shift things around
        shl     ebx, 16
        shl     ebp, 8
        # Or everything together
        or      ebp, ebx
        or      ebp, r14d
        # Write the resulting u32
        mov     dword ptr [rdx + 4*r10], ebp
       # Increment counters for the next loop iteration
        add     r10, 1
        add     r9, 4
        add     rax, 3
        # Jump to next loop iteration
        cmp     r8, r10
        jne     .LBB1_2
```

> My only real concern is with ensuring we aren't slowing down the little-endian path,

I think this is right now mostly a theoretical concern. "The code is different, so it might be slower." Should I start working on a benchmark? As I wrote above: 

>> it really comes down to what the generated code looks like,
>
> I haven't looked at any assembly and even if I did, I am not sure I'd draw the right conclusions. I guess one would need some benchmarks instead to compare approaches.
>
> Do you want to go down this route? If so, I think it would make more sense to add some benchmarks in a separate PR and merge them first.

psychon: Or should I just go with your suggested `native_endian_offsets()` function and that's good enough?

cmyr (approved):
Okay, thanks for looking into the codegen. I'm happy enough that this is very unlikely to be a regression, and I'm happy to improve correctness. Thanks!

cmyr:
CI is failing because of the updated rustc, i'll push a fix for that and you can rebase. :)

piet-cairo/src/lib.rs
```diff
@@ -499,3 +519,20 @@ fn compute_blurred_rect(rect: Rect, radius: f64) -> (ImageSurface, Point) {
     let origin = rect_exp.origin();
     (image, origin)
 }
+
+fn write_rgba(data: &mut [u8], offset: usize, x: usize, r: u8, g: u8, b: u8, a: u8) {
+    // From the cairo docs for CAIRO_FORMAT_ARGB32:
+    //  each pixel is a 32-bit quantity, with alpha in the upper 8 bits, then red,
+    //  then green, then blue. The 32-bit quantities are stored native-endian.
```
JAicewizard, line 526: Bit of a nitpick, but why 2 spaces on all subsequent lines? 1 is used everywhere else.

JAicewizard:
CI seems to be blocked by changes in this PR now. Maybe replace `x` with `index` or smth like that.

piet-cairo/src/lib.rs
psychon: Well... it was meant as indication that it is a quote. I now changed it to `> ` instead.

psychon:
Here are updated criterion results (with "latest version of that criterion PR" vs "latest version of that criterion PR merged with this PR"):
```
make_image_2160p_Grayscale                                                                            
                        time:   [16.193 ms 16.198 ms 16.204 ms]
                        change: [-51.808% -51.698% -51.616%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe

make_image_2160p_Rgb    time:   [18.589 ms 18.602 ms 18.623 ms]                                 
                        change: [-48.528% -48.483% -48.423%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  3 (3.00%) high mild
  12 (12.00%) high severe

make_image_2160p_RgbaSeparate                                                                            
                        time:   [40.119 ms 40.125 ms 40.131 ms]
                        change: [-43.341% -43.277% -43.230%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) high mild
  5 (5.00%) high severe

make_image_2160p_RgbaPremul                                                                            
                        time:   [25.172 ms 25.181 ms 25.192 ms]
                        change: [-56.145% -56.124% -56.102%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe
```

cmyr:
Thank you for your thoroughness!

PR #453, opened 2021-06-27
Bump cairo

JAicewizard:
This bumps cairo and friends to the latest version.

Once this is put in a release druid-shell should also be updated to match the versions.


cmyr (commented):
My only concern about this is around client support; I recall at some point in the past we were considering a newer version of cairo, but there were some potential complications. It might have been that the newer cairo didn't work with an older version of gtk, or something? Do you have any insight on this?

piet-cairo/src/lib.rs
```diff
@@ -121,15 +120,15 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         self.set_path(shape);
         self.set_brush(&*brush);
         self.ctx.set_fill_rule(cairo::FillRule::Winding);
-        self.ctx.fill();
+        self.ctx.fill().unwrap();
```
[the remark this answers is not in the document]
JAicewizard: Then I will definitly do that. I didnt look to closely at it `status` just returned `Ok(())` so I didnt think it was of any use.

JAicewizard:
The text part of the `piet-cairo` doesnt seem to be handling errors either, but that should probably be another PR.

I would also like to remind you of `//FIXME: replace this when pango 0.10.0 lands`.

piet-cairo/src/lib.rs
```diff
@@ -503,17 +511,24 @@ fn affine_to_matrix(affine: Affine) -> Matrix {
...
-    let mut data = image.get_data().unwrap();
-    let rect_exp = piet::util::compute_blurred_rect(rect, radius, stride, &mut *data);
-    std::mem::drop(data);
-    let origin = rect_exp.origin();
-    (image, origin)
+    match ImageSurface::create(Format::A8, size.width as i32, size.height as i32) {
+        Ok(mut image) => {
+            let stride = image.stride() as usize;
+            //TODO: This can still panic.
+            let mut data = image.data().unwrap();
```
cmyr, line 520: given that we've changed the signature to return an error, I think we can also address this now?

cmyr (commented):
Cool, this looks good. My only concern now is with support; are there any possible conflicts related to what version of cairo is bundled by default, or anything like that?

piet-cairo/src/lib.rs
JAicewizard: It is sort of complicated. This returns a borrowerror, not a cairo::Error. I tried making the return type more general, but it cant really be anymore general. It would probably have to be a Box, but that wont work in `success`. If you know a way to do this, please let me know how.

cmyr: The only thing I can think of is that we store a `Box<dyn Error>` instead of a `cairo::Error`; we do end up converting to this type anyway when we make the `piet::Error::BackendError`?

JAicewizard: I just tried, but that doesnt work. I looked at the FFI code for `.data()` and it really seems impossible for it to return an error. An error is returned when either: the reference to `image` is dropped (it isnt since its still in scope), there is an error on `image` (there isnt since we havnt used it yet), the pointer to the image is null aka the surface isnt an imagesurface (it is an imagesurface), or the surface is finished (it isnt, we know because we dont finish it).

If someone else wants to remove this unwrap, im fine with it but its not worth it to me.

cmyr: Okay that sounds good, thanks for looking into it. It might be worth adding a little comment explaining that the unwrap shouldn't fail?

cmyr:
@JAicewizard if you can confirm that this will work fine on gtk3 then I'm satisfied. :)

JAicewizard:
This will requite the same bump on druid-shell, for all gtk-related deps. I tested this on my GTK4 branch using updated deps and that works, perfectly fine, I see no reason it wouldn't using GTK3. The rendering is the same after all.

I will create a branch with GTK3, you can merge now or wait for me to test on that.

cmyr:
my main concern is whether the GTK3 that is bundled with most distributions has a dependency on a previous version of cairo.

JAicewizard:
The new `cairo-rs` uses 1.14 by default. Even Ubuntu stable 16.04 supports this. For ubuntu you'd have to go back to 2014 to find a stable version that runs something older. 14.04 is technically still supported by Canonical, but only for Extended Security Maintenance. Not worth supporting IMO. Even the oldest of servers should have updated by now.

JAicewizard:
I check all other dependencies and they all work back to 16.04 from what I can tell

JAicewizard:
I just realised this probably isnt what you meant. 

No the GTK and cairo should always work together, given they are bundled with the same OS/version. Even then, putting in your own cairo still shouldn't break it. These are all backwards compatible, so as long as you target the lowest version it should work for higher versions as well. We just link to whatever the distribution provides WRT GTK/cairo etc, and those should all match each other. I dont know if this happens at compile time or runtime, but either way the picked GTK3/cairo should always be compatible with each other.

As written in my last comment, we should be compatible with cairo all the way back to 2016, so we are wont run into any issues with this.

cmyr (approved):
Okay, let's give it a try. :)

PR #454, opened 2021-06-27
Make gradient stops hashable

### LINKED ISSUE #450: impl Hash+Eq for GradientStop
I'm implementing gradients in piet-gpu, and one of the things I'm running into is being able to determine whether a gradient ramp is new or already in the cache. There are a number of ways to deal with this, but the most direct is to impl `Hash` on `GradientStop`, so it can be used as a key in a hashmap.

This is messy, because a gradient stop contains a float, and also because it means we have to hash color, but I think it's worth it.

Opening an issue first in case people want to discuss.

raphlinus:
Implement Hash and Eq traits on gradient stops, which implies doing the
same for colors. This facilitates creating a cache of baked gradient
ramps.

Closes #450


cmyr (approved):
👍

PR #456, opened 2021-07-12
Standardise font size when less than 0

### LINKED ISSUE #279: replace font sizes <= 0.0 with default font size
yea, this behaviour should be standardized; it seems like both mac and windows use `12.0` if the passed value is <= 0, and I think we should just enforce that.

_Originally posted by @cmyr in https://github.com/linebender/druid/issues/1156#issuecomment-681165023_

JAicewizard:
fixes  #279

### DIFF
--- a/piet/src/util.rs
+++ b/piet/src/util.rs
@@ -167,7 +167,9 @@ impl LayoutDefaults {
     pub fn set(&mut self, val: impl Into<TextAttribute>) {
         match val.into() {
             TextAttribute::FontFamily(t) => self.font = t,
-            TextAttribute::FontSize(size) => self.font_size = if size <= 0.0 { 12.0 } else { size },
+            TextAttribute::FontSize(size) => {
+                self.font_size = if size <= 0.0 { DEFAULT_FONT_SIZE } else { size }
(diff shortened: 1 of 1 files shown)


piet/src/util.rs
```diff
@@ -167,7 +167,7 @@ impl LayoutDefaults {
     pub fn set(&mut self, val: impl Into<TextAttribute>) {
         match val.into() {
             TextAttribute::FontFamily(t) => self.font = t,
-            TextAttribute::FontSize(size) => self.font_size = size,
+            TextAttribute::FontSize(size) => self.font_size = if size <= 0.0 { 12.0 } else { size },
```
cmyr, line 170: you can use the `DEFAULT_FONT_SIZE` const defined elsewhere in this file. :)

cmyr (commented):
Great. We have a `const` for this; otherwise thanks for the fixup. :)

piet/src/util.rs
JAicewizard: done!

cmyr (approved):
thanks!

PR #466, opened 2021-09-28
Build on OpenBSD

klemensn:
The cairo backend works without any platform specific modifications,
just like on Linux.

### DIFF
--- a/piet-cairo/README.md
+++ b/piet-cairo/README.md
@@ -26,6 +26,12 @@ On macOS with Homebrew, the following should work:
 brew install cairo
 ```
 
+On OpenBSD, the library can be installed from official packages:
+```shell
+pkg_add cairo
+```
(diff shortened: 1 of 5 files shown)


cmyr (approved):
Thanks, this makes sense!

klemensn:
Great, now I'm just waiting for a new proper release so software like `psst` can fetch the crate from the registry rather than a specific git checkout.

This is less important for development on personal machines, but environments such as those for official OS/distribution packages often prefetch things prior to build (which has no network) access, and OpenBSD's package infrastructure only knows how to fetch crates from the registry.

I've been working on fetching from git, which already works, but that also has implications to the build process, which are a bit trickier.

tl;dr: it would be nice to see a new release sooner than later to ease downstream development.

cmyr:
@klemensn no worries, happy to do a release, have been meaning to do the proper 0.5 anyway. :)

Issue #383, opened 2021-01-01
piet-coregraphics font weight problems


eulerdisk:
Running test-picture in compare mode agains the snapshots dir for piet-coregraphics gives me a lot of differences. Basically all the tests with fonts are different. Maybe most of them are sampling differences due to the GPU, but in test 14 seems like font weight isn't working at all.

MacOS 10.14.6
MacBook Air mid 2012 (Intel HD graphics 4000)

[test-images.zip](https://github.com/linebender/piet/files/5759439/test-images.zip)

cmyr:
These definitely change based on hardware and OS, because some of these fonts are not shipped on all OSes. For this reason the generated images include a `GENERATED_BY` file that includes the OS version and other relevant details.

It's also likely that your MacBook has different default settings because it doesn't have a 'retina' display; for instance it might be that sub-pixel anti-aliasing is on by default, where it is off in CI.

Example 14 uses a variable font, so that won't work on older OSes, but I *would* expect it work on  10.14.x, so that's a bit of a mystery. 😕 


summing up: I think everything here is expected except for the missing bold on example 14, which I'm confused about.

PR #470, opened 2021-10-31
Always disable font smoothing when creating a CoreGraphicsContext

lord:
This makes all text rendered match the rendering used by Sketch. There's also [this blog post](https://tonsky.me/blog/monitors/#turn-off-font-smoothing) by Tonsky that discusses it.

To me this seems reasonable to just disable for all apps. But could also make changing this a custom option that people can opt-in to? (Although I'd maybe still argue for defaulting to false)

Here are some `@2x` screenshots taken in Pages (which uses font smoothing if it's turned on system-wide) and Sketch (which always turns it off)
Artboard

### DIFF
--- a/piet-coregraphics/src/lib.rs
+++ b/piet-coregraphics/src/lib.rs
@@ -87,6 +87,7 @@ impl<'a> CoreGraphicsContext<'a> {
         text: Option<CoreGraphicsText>,
     ) -> CoreGraphicsContext {
         ctx.save();
+        ctx.set_allows_font_smoothing(false);
         if let Some(height) = height {
             let xform = Affine::FLIP_Y * Affine::translate((0.0, -height));
             ctx.concat_ctm(to_cgaffine(xform));


lord:
This may explain some of #383, but I'm not certain

cmyr:
It looks like this option has finally been removed in monterey? In the back of my head the only possible concern I'd have, here, is whether this makes for a worse experience for folks on non-retina external monitors, but maybe this doesn't concern us.

Did you run into this as a problem, somewhere?

lord:
Non-retina external monitors are a good point! Not sure how this would affect those. I'd be happy to amend this PR to make this an optional function call instead?

I guess I ran into this since I noticed the rendering seemed wrong on one of my projects, and I couldn't figure out a way to call this function without making modifications to Piet. Would be happy to be wrong!

cmyr:
@lord yea, I think it's fine to have platform-specific helpers. I'm not sure where exactly I would hang this method. If you're using `CoreGraphicsContext` directly it could go there, otherwise we might need an extension trait?

PR #472, opened 2021-11-17
Cache CTFont handles on macOS

jpochyla:
This implements caching of `CTFont` handles, not unlike the current font-family cache. This solves a real bottleneck I have in Psst (where creating lot of labels used to be really slow). I've used the `associative-cache` crate, because it is used in the Direct2D backend already (for caching gradients), and also switched the font family cache to it, for consistency.

Thanks!

### DIFF
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -6,7 +6,7 @@ use std::ops::{DerefMut, Range, RangeBounds};
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
-use associative_cache::{AssociativeCache, Capacity1024, HashFourWay, RoundRobinReplacement};
+use associative_cache::{AssociativeCache, Capacity64, HashFourWay, RoundRobinReplacement};
 use core_foundation::base::TCFType;
 use core_foundation::dictionary::{CFDictionary, CFMutableDictionary};
 use core_foundation::number::CFNumber;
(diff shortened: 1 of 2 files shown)


piet-coregraphics/src/text.rs
```diff
@@ -49,7 +50,15 @@ struct SharedTextState {
...
 struct TextState {
     collection: FontCollection,
-    family_cache: HashMap<String, Option<FontFamily>>,
+    family_cache: AssociativeCache<
+        String,
+        Option<FontFamily>,
+        Capacity1024,
+        HashFourWay,
+        RoundRobinReplacement,
+    >,
```
cmyr, line 59: I might suggest a type alias, like:

```rust
type Cache<K, V> = AssociativeCache<K, V, Capacity1024, HashFourWay, RoundRobinReplacement>;
```

cmyr (approved):
This all looks very reasonable, only one small suggestion that you are free to ignore.

piet-coregraphics/src/text.rs
cmyr: Another question here is whether `1024` is the right capacity; in general I would expect far fewer font objects to be in use than that, but I don't have strong opinions. I guess the only downside of a larger cache is potentially using more memory than necessary holding on to a bunch of stale font objects, although this does not seem particularly likely for our current use-cases.

Personally though I think I would pick a smaller capacity, like maybe as small as `64`? It just isn't very likely that there are more than 64 different fonts in use at the same time. 🤷

jpochyla: Those are all very good points, I wasn't completely sure about the size.

jpochyla:
I've added the alias and decreased cache size to 64 for both font-families and ctfonts.

piet-coregraphics/src/text.rs
```diff
@@ -102,6 +107,113 @@ struct Attributes {
...
+impl Eq for CoreTextFontKey {}
+
+impl Hash for CoreTextFontKey {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.font.hash(state);
+        self.weight.hash(state);
+        self.italic.hash(state);
+        self.size.to_bits().hash(state);
+    }
+}
```
cmyr, line 136: just taking a last look, is there a reason we can't derive these?

jpochyla: It's the `to_bits()` on floats :(

cmyr:
great, thanks!

PR #473, opened 2021-11-26
Add piet-common reference to main documentation

luizberti:
Currently there is no direct link to `piet-common`'s documentation from
the main docs page, and that is quite annoying to navigate if you're
referencing both a lot.

This adds that, along with some newbie guidance for how to create
`RenderContext`s for those who are getting started.

### DIFF
--- a/piet/src/lib.rs
+++ b/piet/src/lib.rs
@@ -13,6 +13,11 @@
 //! to be reused in an approximately consistent way. Various such implementations
 //! exist, such as [`piet-cairo`], [`piet-coregraphics`], and [`piet-direct2d`].
 //!
+//! If you are interested in using piet to target multiple platforms,
+//! [`piet-common`] is a utility crate that re-exports an appropriate
+//! backend based on the compile target, and provides convenience
+//! types and methods for setting up a [`RenderContext`].
+//!
 //! [`PostScript`]: https://en.wikipedia.org/wiki/PostScript
 //! [`piet-common`]: https://docs.rs/piet-common
 //! [`piet-cairo`]: https://crates.io/crates/piet-cairo
--- a/piet/src/lib.rs
+++ b/piet/src/lib.rs
@@ -4,13 +4,17 @@
 //! graphics API, in the tradition of [`PostScript`]. It is built on top of
 //! [`kurbo`], a 2D geometry library.
 //!
-//! The main interface is the [`RenderContext`] trait.
+//! The main interface is the [`RenderContext`] trait. Another useful place
+//! to look at if you're getting started would be the `Device` and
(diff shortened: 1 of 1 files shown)


piet/src/lib.rs
```diff
@@ -4,13 +4,17 @@
...
 //!
-//! The main interface is the [`RenderContext`] trait.
+//! The main interface is the [`RenderContext`] trait. Another useful place
+//! to look at if you're getting started would be the `Device` and
+//! `BitmapTarget` documentation on [`piet-common`], to create `RenderContext`
+//! objects and extract rendered pixels from them.
 //!
 //! This API can be implemented on various platforms, allowing drawing code
 //! to be reused in an approximately consistent way. Various such implementations
 //! exist, such as [`piet-cairo`], [`piet-coregraphics`], and [`piet-direct2d`].
```
cmyr, line 14: forgive my micro-management, but I would prefer to have the section mentioning piet-common be at the bottom, after we mention the other crates:

```suggestion
//! exist, such as [`piet-cairo`], [`piet-coregraphics`], and [`piet-direct2d`].
//!
//! If you are interested in using piet to target multiple platforms,
//! [`piet-common`] is a utility crate that re-exports an appropriate
//! backend based on the compile target, and provides convenience
//! types and methods for setting up a [`RenderContext`].
```

piet/src/lib.rs
```diff
@@ -4,13 +4,17 @@
 //! graphics API, in the tradition of [`PostScript`]. It is built on top of
 //! [`kurbo`], a 2D geometry library.
 //!
-//! The main interface is the [`RenderContext`] trait.
+//! The main interface is the [`RenderContext`] trait. Another useful place
+//! to look at if you're getting started would be the `Device` and
+//! `BitmapTarget` documentation on [`piet-common`], to create `RenderContext`
+//! objects and extract rendered pixels from them.
```
cmyr, line 10: I'd discard this in favour of something like the edit below:

cmyr (commented):
Thanks, I absolutely agree that we should mention `piet-common` from this page. I have some thoughts on where that should go, which I've included inline; feel free to make those changes if you are so inclined, otherwise I can open another PR. :)

piet/src/lib.rs
luizberti: Was it supposed to have a code suggestion on this comment too? I already applied your other one, otherwise, I don't follow

cmyr: I don't think there's an easy way for me to 'suggest' discarding an edit? I guess I can just suggest an edit reverting to the previous text 🤔

```suggestion
//! The main interface is the [`RenderContext`] trait.
```

cmyr:
Thanks!

PR #474, opened 2021-11-26
Add warning about ImageFormat and backend compat

luizberti:
Currently, calling either of these functions will simply return `Error::NotSupported`
without any further explanation. This can lead one to believe that the function itself
is not supported on a given platform, rather than just the `ImageFormat`.

This change clarifies that this is the caller's responsibility in the documentation.


cmyr (commented):
Thanks, this seems like a useful addition. While we're doing this, could you add a similar note to `cairo_back.rs` and `direct2d_back.rs`?

luizberti:
Done and done!

cmyr (approved):
Thanks!

PR #475, opened 2021-12-03
Implement text and bitmap images in piet-svg

derekdreery:
The tricky bit is text layout sizing, which I need for my use-case (drawing graphs in svg using piet).  I've taken the approach of loading the font on the host machine, then doing a shaping run with `rustybuzz` and then calculating the dimensions of the run. It doesn't work perfectly but it's OK on my machine. I thought there would be an issue with DPI, but it seems that the SVG renderer (I'm using Firefox) operates at the standard DPI and then scales everything up, so there's no issue. Named font sizing works better than using `sans-serif` or `system-ui`, presumably because font-kit and Firefox differ in their opinion of what the default sans-serif font is.

Anyway I'm PRing this because although it is not perfect, it is a significant improvement on what is currently there (nothing). I'm happy to put some effort into getting this through, so please do suggest changes etc. as necessary.

**Update** The text rendering is more robust now, and I added image rendering as well since it was fairly straightforward (in comparison). Hit tests etc are still not implemented, but I don't think you'll need these often (if ever) on an SVG backend.

### DIFF
--- a/piet-svg/src/evcxr.rs
+++ b/piet-svg/src/evcxr.rs
@@ -6,7 +6,7 @@
 //! [`evcxr`]: https://github.com/google/evcxr
 //! [Jupyter Notebook]: https://jupyter-notebook.readthedocs.io/en/stable/notebook.html
 
-use piet::RenderContext;
+use piet::{kurbo::Size, RenderContext};
 
 impl evcxr_runtime::Display for crate::RenderContext {
     fn evcxr_display(&self) {
@@ -20,8 +20,11 @@ impl evcxr_runtime::Display for crate::RenderContext {
 /// Runs the function `f`, and displays the resulting `SVG`.
 ///
 /// For use within `evcxr_jupyter`.
-pub fn draw_evcxr(f: impl FnOnce(&mut crate::RenderContext)) -> impl evcxr_runtime::Display {
-    let mut ctx = crate::RenderContext::new();
+pub fn draw_evcxr(
+    size: Size,
+    f: impl FnOnce(&mut crate::RenderContext),
+) -> impl evcxr_runtime::Display {
+    let mut ctx = crate::RenderContext::new(size);
     f(&mut ctx);
     ctx.finish().unwrap();
     ctx
--- a/piet-svg/src/evcxr.rs
+++ b/piet-svg/src/evcxr.rs
@@ -1,0 +1,20 @@
+use piet::RenderContext;
+
+impl evcxr_runtime::Display for crate::RenderContext {
+    fn evcxr_display(&self) {
+        evcxr_runtime::mime_type("text/html").text(format!(
+            r#"<div style="display:flex;justify-content:center;">{}</div>"#,
+            self.display()
+        ))
+    }
+}
+
+/// Runs the function `f`, and displays the resulting `SVG`.
+///
+/// For use within `evcxr_jupyter`.
+pub fn draw_evcxr(f: impl FnOnce(&mut crate::RenderContext)) -> impl evcxr_runtime::Display {
+    let mut ctx = crate::RenderContext::new();
+    f(&mut ctx);
+    ctx.finish().unwrap();
+    ctx
+}
--- a/piet-svg/src/evcxr.rs
+++ b/piet-svg/src/evcxr.rs
@@ -1,3 +1,11 @@
+//! Helpers to make the SVG output of `piet-svg` easier to use from within `evcxr_jupyter`.
+//!
+//! [`evcxr`] is a Rust REPL. It also provides a Rust kernel for the [Jupyter Notebook] through
+//! `evcxr_jupyter`.
+//!
+//! [`evcxr`]: https://github.com/google/evcxr
+//! [Jupyter Notebook]: https://jupyter-notebook.readthedocs.io/en/stable/notebook.html
+
 use piet::RenderContext;
 
 impl evcxr_runtime::Display for crate::RenderContext {
--- a/piet-svg/src/lib.rs
+++ b/piet-svg/src/lib.rs
@@ -2,6 +2,8 @@
 
 #![deny(clippy::trivially_copy_pass_by_ref)]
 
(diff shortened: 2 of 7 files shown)


derekdreery:
SVG doesn't have the concept of multiline text or sub-spans of differently styled text, so if we wanted to implement those things, we'd have to render them as a number of separate runs, which would require accurate measuring.

derekdreery:
Now supports images as well (by embedding them as png).

ratmice:
I tried this branch out with something, everything went well with translated/rotated text besides having to fudge the font metrics for characters smaller than an em but that is an entirely separate issue...  

anyhow, thanks for doing this!

derekdreery:
> I tried this branch out with something, everything went well with translated/rotated text besides having to fudge the font metrics for characters smaller than an em but that is an entirely separate issue...
> 
> anyhow, thanks for doing this!

NP! I did test transformed text, but I didn't test clipping. Might be worth someone (e.g. me) doing this, or maybe there is already a clipping test where I just need to check the output.

piet-svg/src/evcxr.rs
```diff
@@ -0,0 +1,20 @@
+use piet::RenderContext;
```
cmyr, line 1: would love a module docstring telling me what `evcxr` is 🤷

piet-svg/src/lib.rs
```diff
@@ -42,13 +48,30 @@ impl RenderContext {
         }
     }
 
+    /// Set the size that the SVG should render to
+    pub fn set_size(&mut self, size: Size) {
```
cmyr, line 52: is it intended/helpful for this to be mutated in the middle of an operation, or should it be set on  create and subsequently be immutable? This would better match the semantics of other backends.

piet-svg/src/text.rs
```diff
@@ -1,114 +1,356 @@
...
+use rustybuzz::{Face, UnicodeBuffer};
 
 type Result<T> = std::result::Result<T, Error>;
 
 /// SVG text (unimplemented)
-#[derive(Debug, Clone)]
-pub struct Text;
+#[derive(Clone)]
+pub struct Text {
+    source: Arc<Mutex<MultiSource>>,
```
cmyr, line 27: Is this a `Mutex` only because we're using `Arc`, or are we actually expecting this to be used from multiple threads?

piet-svg/src/text.rs
```diff
@@ -1,114 +1,356 @@
...
 
     fn trailing_whitespace_width(&self) -> f64 {
-        unimplemented!()
+        // unimplemented
+        0.
     }
 
     fn image_bounds(&self) -> Rect {
-        unimplemented!()
+        Rect::from((Point::from((0., 0.)), self.size()))
```
cmyr, line 245:
```suggestion
        self.size().to_rect()
```

cmyr (approved):
Okay a few little notes inline, but this looks strictly more useful than what we currently have?

I played around with a bunch of the sample images, and there's some weird stuff going on: `clear` doesn't seem to work correctly, there are some weird scaling issues, etc. All that being said, `piet-svg` is definitely a 'best effort' project, so I'm happy to see this merged.

I *do* think it would be helpful to adapt the sample-drawing example from one of the other crates to work with piet-svg: it can write `svg` files to disk, and will accept all the same command-line arguments etc. This will make it easier to test `piet-svg` going forward. :)

piet-svg/src/lib.rs
derekdreery: So the reason I did it like this is because (unlike the other backends), SVG images have a concept of "unsized" - you don't have to set the size or viewBox of an SVG image.

It might be that we want to restrict this to forcing a size (like other backends). In this case I could change as you suggested.

piet-svg/src/text.rs
derekdreery: I chose the type this way because I wanted the text object to be `Clone`, `Send`, and `Sync`. You can think of it as a shared object that contains all fonts and allows you to fetch them. The fact that this might involve loading from disk is abstracted away.

The alternative to this strategy is using `Rc<RefCell>` and having a separate object per thread. I would think this is wasteful if you're loading fonts from disk multiple times. If you're only going to use the text object from the main thread then this isn't an issue.

What would really be nice is a library like `font-kit` that was a bit more high-level, where there is just 1 object for font loading that reference counts any shared state under the hood. That way you could be more clever about avoiding locks and blocking.

I think I'm rambling, so I'll stop. Basically, I can leave it as is or change it to `Rc<RefCell>`, at your preference. I think I need to use interior mutability, because we expect `RenderContext::Text` to be cheap to clone.

piet-svg/src/lib.rs
cmyr: What is the expected behaviour when an svg is unsized?

I'm wondering if this is compatible with the overall design of the crate. For instance, If an SVG is unsized, what should be the behaviour of the `clear` fn?

piet-svg/src/text.rs
cmyr: This is fine, thanks!

piet-svg/src/lib.rs
derekdreery: I vaguely remember another reason for wanting the ability to set size after creation but I can't remember it now. I will make sure all my things work without the `set_size` method, then I propose we get rid of it and make the api like the other backends.

cmyr: still waiting on this, yea? otherwise I think I'm happy to merge :)

derekdreery: done :)

cmyr:
We'll also have to fix the example that's causing clippy to balk

derekdreery:
No worries sorry for the slow reply (Christmas etc.).

derekdreery:
Urgh I'm sure the CI is fine now but its failing due to failed web request.

cmyr:
looks good-to-go, thanks!

Issue #478, opened 2021-12-13
D2D1CreateFactory causes no graceful exit on Windows


wspl:
Background information: https://github.com/linebender/druid/issues/2062

Simply put, `piet_direct2d::D2DFactory::new()` will cause the program to no longer exit normally via Ctrl+C when it reaches the `D2D1CreateFactory` method.

wspl:
I found the cause of the problem and tried to fix it: https://github.com/linebender/piet/pull/479

PR #479, opened 2021-12-13
Set D2D1Factory debug mode to NONE

wspl:
Related issues:
https://github.com/linebender/druid/issues/2062
https://github.com/linebender/piet/issues/478

Initialize `D2D1Factory` with `debugLevel=D2D1_DEBUG_LEVEL_WARNING` will prevent programs from exiting with Ctrl+C under Windows.

Disable the debug mode of D2D1Factory can solve this problem.


cmyr:
Thanks! I don't have enough windows knowledge to know offhand that this is correct, can you point me to some documentation or something that explains what this is doing?

raphlinus:
This doesn't seem right to me. If there are problems that surface when debug warnings are enabled, those may be real problems, and if so it's not a good idea to sweep them under the rug. In that case, it would be much better to figure out why the warnings are being produced and solve those.

One thing that I would be open to is setting warnings only in debug mode, especially if there is a measurable performance impact. But that's not a substitute for resolving what's causing the warnings.

raphlinus:
I've read the linked issue. Hmm, it looks like maybe Direct2D is taking over the signal handling for Ctrl-C when this flag is set? That might be something to avoid.

wspl:
@raphlinus It looks like this is a normal behavior of the debug mode of win32 apis/programs, and probably unavoidable (or does it require an exit signal sent via windbg? I don't know much about this).

Enabling the debug mode of D2D1Factory via command line arguments may be a better option. For this approach, see Node.js, which provides a `--inspect` parameter to allow it communicate with the v8 debugger.

wspl:
@cmyr The relevant documents I found: https://docs.microsoft.com/en-us/windows/win32/Direct2D/direct2ddebuglayer-overview

The Direct2D Debug Layer is really needed in some scenarios. So we need to find a way to turn it on conditionally, rather than by default.

avitex:
Perhaps a suitable solution to this is using features, with something [similar to how the `log` crate can handle log levels](https://github.com/rust-lang/log/blob/2f7e96f16ced35244f25d0ef453ef9e1a91c6dbe/src/lib.rs#L1640).

```toml
[features]
release_level_error = []
release_level_warn = []
release_level_info = []
level_off = []
level_error = []
level_info = []
```
Perhaps defaulting to `D2D1_DEBUG_LEVEL_NONE` for release builds and `D2D1_DEBUG_LEVEL_WARNING` for debug builds. This means tests will run with warnings.

From the [docs](https://docs.microsoft.com/en-us/windows/win32/Direct2D/direct2ddebuglayer-overview):

> If no factory options are specified or a debug level of "none" is specified, the debug layer is not invoked. **The debug layer should never be active in the release version of an application.**

If an application (not library) author wanted to disable for debug builds due to the original issue then they could set with:

```toml
[target.'cfg(target_os="windows")'.dependencies]
piet-direct2d = { version = "=0.5.0", features = ["level_off"] }
```

PR #482, opened 2022-01-02
Add iOS as a CoreGraphics operating system to piet-common

lord:
Confirmed text drawing and simple shapes work on iOS:

IMG_94837D379BD8-1


cmyr (approved):
Looks good, I would expect maybe *some* differences here, like iOS graphics contexts are y-down by default? But I'm sure we can work around that.

Issue #483, opened 2022-01-18
Add `Clone` impl to `TextAttribute`


Fenex:
I have a `'static` const array with some types of text styles:

```Rust
const SEPARATOR_STYLE: &'static [TextAttribute; 1] = &[
    TextAttribute::TextColor(Color::BLUE)
];
const ERRORS_STYLE: &'static [TextAttribute; 2] = &[
    TextAttribute::TextColor(Color::YELLOW),
    TextAttribute::Strikethrough(true)
];
```

I would be happy to clone values from its, however `TextAttribute` doesn't implement a `Clone` trait. (PR #484)

I research all variants of `TextAttribute` and all its variats implements `PartialEq` and `Debug`, so `TextAttribute` could to derive this additional traits too. (PR #485)

<details>
  <summary>more detail code example isssue</summary>
  
```Rust
#[derive(Default)]
struct StyledRanges {
    separator: Vec<Range<usize>>,
    errors: Vec<Range<usize>>,
}

impl StyledRanges {
    const SEPARATOR_STYLE: &'static [TextAttribute; 1] = &[
        TextAttribute::TextColor(Color::BLUE)
    ];
    const ERRORS_STYLE: &'static [TextAttribute; 2] = &[
        TextAttribute::TextColor(Color::YELLOW),
        TextAttribute::Strikethrough(true)
    ];
}

let (ranges, text): (StyledRanges, String) = logger.get_text_with_errors();

let mut attrs: Vec<(Range<usize>, &[TextAttribute])> = std::iter::empty()
    .chain(&mut ranges.separator.into_iter().zip(std::iter::repeat(&StyledRanges::SEPARATOR_STYLE[..])))
    .chain(&mut ranges.errors.into_iter().zip(std::iter::repeat(&StyledRanges::ERRORS_STYLE[..])))
    .collect();

// sort all types by `start` key due avoid assert
// https://github.com/linebender/piet/blob/8f7bc70a097d585205573442b970ec087d4e8f3b/piet-cairo/src/text.rs#L265-L268
attrs.sort_by_key(|(r, _)| r.start);

let mut text_layout: PietTextLayoutBuilder = ctx // &mut PaintCtx
    .text() // &mut PietText
    .new_text_layout(text) //PietTextLayoutBuilder
    .max_width(rect.width());

for (range /* Range<usize> */, attrs /* &[TextAttribute] */) in attrs {
    for attr /* &TextAttribute */ in attrs {
[...]

PR #484, opened 2022-01-18
feat(piet): add Clone impl to TextAttribute

Fenex:
Issue #483 (conflict with PR #485)

The PR adds `Clone` derive to `TextAttribute`.


cmyr:
let's take #485 instead. thanks!

PR #485, opened 2022-01-18
feat(piet): add Debug,Clone,PartialEq impls to TextAttribute

Fenex:
Issue #483  (extended, conflict with PR #484) 

The PR adds  `Debug`, `Clone` and `PartialEq` derives to `TextAttribute`.

### DIFF
--- a/piet/src/text.rs
+++ b/piet/src/text.rs
@@ -141,6 +141,7 @@ impl std::ops::Deref for dyn TextStorage {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
 /// Attributes that can be applied to text.
 pub enum TextAttribute {
     /// The font family.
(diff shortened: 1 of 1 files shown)


cmyr:
@Fenex just out of curiosity, do you have any particular need for `PartialEq`? I'm not sure if it's particularly helpful, so I might keep this to just `Debug` and `Clone` unless you have a particular case in mind. :)

Fenex:
Mmm. In real I need `Clone` impl only. (:

piet/src/text.rs
```diff
@@ -141,6 +141,7 @@ impl std::ops::Deref for dyn TextStorage {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
```
cmyr, line 144:
```suggestion
#[derive(Debug, Clone)]
```

PR #487, opened 2022-01-20
Release 0.5.1

derekdreery:
I wonder if you'd consider releasing a new version of piet. It contains support for outputting bitmaps on cairo. There are very few changes other than that - this would be a patch release even if piet were stable.

If you prefer you could close this PR and use cargo-workspaces to do the releases.

piet@0.5.1
piet-cairo@0.5.1
piet-common@0.5.1
piet-coregraphics@0.5.1
piet-direct2d@0.5.1
piet-svg@0.5.1
piet-web@0.5.1

Generated by cargo-workspaces

### DIFF
--- a/piet-cairo/Cargo.toml
+++ b/piet-cairo/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "piet-cairo"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 description = "Cairo backend for piet 2D graphics abstraction."
 license = "MIT/Apache-2.0"
@@ -11,7 +11,7 @@ keywords = ["graphics", "2d"]
 categories = ["rendering::graphics-api"]
 
 [dependencies]
-piet = { version = "=0.5.0", path = "../piet" }
+piet = { version = "0.5.1", path = "../piet" }
 
 cairo-rs = { version = "0.14.0", default-features = false } # We don't need glib
 pango = { version = "0.14.0", features = ["v1_44"] }
--- a/piet-common/Cargo.toml
+++ b/piet-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "piet-common"
-version = "0.5.0"
+version = "0.5.1"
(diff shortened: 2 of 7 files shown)


piet-cairo/Cargo.toml
```diff
@@ -11,7 +11,7 @@ keywords = ["graphics", "2d"]
 categories = ["rendering::graphics-api"]
 
 [dependencies]
-piet = { version = "=0.5.0", path = "../piet" }
+piet = { version = "0.5.1", path = "../piet" }
```
cmyr, line 14: you'll notice in all the intra-workspace dependencies we specify them exactly, with `= "=version"`.

...and now I forget why we do this, exactly. It had something to do with the pre-releases for 0.5, where there could be breaking changes between pre-releases and we wanted to make sure anything that specified a pre-release would not be moved from that release by the resolver (since iirc `0.5` is considered compatible with `0.5-pre1` for the purposes of semvar?) 🤔

derekdreery: Maybe it's easier for you to just close this PR and do the release, since `cargo ws` automates the whole process.

cmyr: the main blocker for a release is figuring out what to do about #488 😒

derekdreery: Aah OK cool. If I get chance I'll have a look at how much work would be required to move to the `windows` crate. It's surely something that will need doing anyway at some point, since it's the one supported by Microsoft.

derekdreery: Now that we've decided to ignore the clippy lint, could we do a release now? Worst case scenario, we could yank it, and I don't think that will be necessary.

cmyr: What exactly were you wanting to get in this release? when I diff master against v0.5, I don't see any changes in piet-cairo? (want to make sure I'm not overlooking something)

derekdreery: ack I think I made a mistake - it was because I didn't enable the png feature rather than the version. Apologies!

PR #489, opened 2022-01-25
mac: Create gradients with DeviceRGB colorspace

jpochyla:
This fixes the gradient coloring on Mac. I'm not exactly sure why just switching to `CGColorSpace::create_device_rgb()` doesn't help, but I had to also avoid `CGColor` entirely and compose the gradient from the color components.


cmyr:
You'll need to update the piet-snapshots repo to include the new snapshots, there should be instructions in the CONTRIBUTING file (in this repo). :)

PR #496, opened 2022-02-06
Update `foreign-types` to `0.5.0`

avitex:
Last of my deps update spam. Left this separate from the others :)

### DIFF
--- a/piet-coregraphics/Cargo.toml
+++ b/piet-coregraphics/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["rendering::graphics-api"]
 [dependencies]
 piet = { version = "=0.5.0", path = "../piet" }
 
-foreign-types = "0.3.2"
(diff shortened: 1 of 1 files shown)


cmyr:
Unfortunately iirc we can't update this dep, since one of our dependencies is stuck on the old version.

cmyr:
@avitex you're welcome to try, but I suspect it will be a lot of work and I'm not sure it's the best use of time. My ultimate conclusion was that we have a solution that works, and it wasn't worth making more work for ourselves. You are free to draw different conclusions of course!

I'm going to close this for the time being, but feel free to open again if you believe that there are no upstream conflicts. :)

avitex:
I made a even more upstream PR https://github.com/sfackler/foreign-types/pull/17
Will reopen this if I manage to make it out of that hole :)

cmyr:
good luck!

PR #502, opened 2022-02-09
Move samples out of `piet` and into local crate

avitex:
- Cargo manifests have been cleaned up a bit
- `piet` now no longer includes fonts within the crate published, making it smaller
- `piet::Error::InvalidSampleArgs` is removed, which is specific to samples


avitex:
I might end up moving `samples` to another folder while working on a more comprehensive test suite, but I'll leave it like this for now

avitex:
Note the fonts are uploaded in the crate: https://docs.rs/crate/piet/latest/source/

avitex:
Awesome, this is the line of thinking I'm going down too. I'll move it into `piet-test` now and build out some more test stuff in there in following PRs

avitex:
The reason I named it samples for the time being is it is close enough to examples, which I would look at for a "sample" of what I can output with the library. Moving all the stuff into `piet-test` hides it away a little more

cmyr:
That's fair, but I think it's going to cause more confusion, since it isn't clear how to run these samples, and people have opened issues about that before. I'd prefer to avoid that confusion.

(I'd also like to move the `test-picturer.s` file out of `examples` and into `bin` for each backend that runs the snapshots; I'd also rename it to "generate_snapshots.rs" or something)

avitex:
Quick note perhaps for further discussion. If the `sample_x`/`generate_snapshots` was to be moved from `examples` to `bin` this then means `piet-test` would have to be a published crate to be an optional dependency referenced within `required-features` for the bin in `Cargo.toml`. I instead choose to put them each in `piet-test` for the time being to remove that dependency.

xStrom:
@avitex any interest in reviving this or should we close it?

PR #503, opened 2022-02-13
Remove unspecific comment

avitex:
This comment used to point out unneeded `unsafe` which has been removed.

On a side note, I had a look at the hacks for RTL and it doesn't look like the original problem has been fixed yet (tested with pango `1.50.3`). I did start going down the path of fixing pango, but I need to find some more time for that task as it is a bit more convoluted than I thought it would be. I'll make a PR if I do get around to fixing it.


cmyr (approved):
Thanks!

PR #504, opened 2022-03-13
Build on FreeBSD.

goyox86:
Hello there! 

I was trying to play with Druid on FreeBSD and saw Piet works on OpenBSD so followed #466 everything builds and the example at least works flawlessly 😲 

No changes needed, the Mondrian example `cargo run --example mondrian --features png` generates a beautiful picture.


cmyr (approved):
Thanks!

PR #512, opened 2022-04-15
piet-svg: fix missing transform, clip when drawing image

flxzt:
Unless I misread how piet works, I think there was a missing apply of the transform and clip path when drawing an image. This should fix it. :)


derekdreery:
Sort answer: yes.

Longer answer: I can't remember all the details of my implementation, but this PR certainly looks reasonable. Let's merge it and if there are issues they'll surface as people use the lib. Sorry I don't have more time to look at it more thoroughly.

PR #519, opened 2022-06-22
Improve test image scale support.

xStrom:
The current test image code has a constant size which is hardcoded to expect a scale factor of `2.0`.

This PR changes the test image code to specify the logical size that the code actually uses, and then the platform code that actually generates the image and chooses the scale factor will also choose the proper image size.

Results remain unchanged for the scale factor of `2.0` but this reduces confusion and helps enable a future potential of doing snapshot testing with multiple different scale factors.

### DIFF
--- a/piet-cairo/examples/test-picture.rs
+++ b/piet-cairo/examples/test-picture.rs
@@ -9,7 +9,8 @@ use cairo::{Context, Format, ImageSurface};
 use piet::{samples, RenderContext};
 use piet_cairo::CairoRenderContext;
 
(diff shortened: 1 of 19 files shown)


xStrom:
Some test images have even more hardcoded `2.0` expectations. I'll need to address those as well.

cmyr (approved):
Thanks, happy to have this be a bit more coherent. There's no good explanation behind the original approach, it was basically carried forward from the very first sample code in raph's first cut of piet, uh, *several* years ago. 😬

PR #522, opened 2022-06-28
Add runtime scale configuration to the test picture generator.

xStrom:
This PR further generalizes the test picture code by making it much more scale-aware. It is now possible to specify a `--scale 5` option at runtime to get the test pictures at 500% scale. The generated filenames now contain the scale information and the comparison functions are aware of this.

Additionally I also updated the snapshots testing configuration to do the tests at both 1x and 2x scale. While reviewing #513 I've found that this points out weird behavior that would go unnoticed with just the old 2x scale testing.

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -27,11 +27,18 @@ jobs:
           profile: minimal
           override: true
 
-      - name: generate and compare (ubuntu+cairo)
+      - name: generate and compare (1.00) (ubuntu+cairo)
         uses: actions-rs/cargo@v1
         with:
           command: run
-          args: --manifest-path=piet-cairo/Cargo.toml --example=test-picture -- --all --out=cairo_samples --compare=./piet/snapshots/cairo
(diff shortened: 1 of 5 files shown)


xStrom:
The snapshots have been updated in [piet-snapshots#30](https://github.com/linebender/piet-snapshots/pull/30).

One slightly annoying thing with having two separate invocations of `test-picture` is that when submitting a new test image, the first CI run will only provide 1x scale artifacts, because the 2x scale generation is never run due to 1x failure. So the snapshots ref has to be updated twice, as seen in this PR here.

One solution to that would be to support repeated `--scale` for a single invocation, and effectively run the code in a loop. Might be worth doing if snapshot updates become frequent here. Until then, I think this double invocation situation does the job for now.

cmyr (approved):
Just a thought: one reasonable approach here might be to not have separate versions of all images, but instead just having a couple specific images where we generate both?

That is kind of annoying, though, so I do think this is reasonable, especially given that I do not anticipate us adding a bunch of new images.

xStrom:
In the interest of time prioritization I'll merge as is. We can come back to this if we're sufficiently annoyed by any of this in practice.

PR #524, opened 2022-07-06
Derive `Copy` for `Color`

fwcd:
A small patch that derives the `Copy` trait for `Color` making it more convenient to use as a value and also permits embedding it in other `Copy` structures.

### DIFF
--- a/piet-common/examples/mondrian.rs
+++ b/piet-common/examples/mondrian.rs
@@ -80,7 +80,7 @@ impl Mondrian {
     fn generate(&self, size: Size, ctx: &mut impl RenderContext) {
         // Start with single rect over whole picture
         let mut rects = vec![ColorRect {
-            color: self.white.clone(),
+            color: self.white,
             rect: Rect::new(0., 0., size.width, size.height).inset(-self.border * DPI),
         }];
         let mut rng = thread_rng();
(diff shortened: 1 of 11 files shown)


cmyr:
We declined to do this because it turns what is currently an implementation detail (that a color is just an int) into part of the actual API, which may not be desirable. For instance, in the future a color might hold a reference to some color space object, which might not implement `Copy`.

It now seems unlikely that piet is going to evolve in that particular direction, but this was the original motivation, and I still sort of like this way of thinking. Is the absence of `Copy` causing you a real headache?

xStrom:
YAGNI seems prudent. I think this is a nice change for `Color` as it stands now. If complex color is added in later, a major version increase can help deal with the API breakage.

To continue with YAGNI let's fix the clippy errors by changing the method signatures to pass `Color` by value.

fwcd:
I've updated the methods to pass `Color` by value where possible, note that this is a breaking change of the API, however.

xStrom:
We've got `v0.6.0` coming up soon so this is the perfect time for an API change. The leaner API makes sense in light of `Color` likely remaining lightweight for now.

Also, there seem to be a few remaining compilation issues.

cmyr (approved):
works for me!

@xStrom feel free to merge if you're satisfied.

xStrom (approved):
Thanks for your effort! Let's get this merged now and see how it goes.

PR #525, opened 2022-07-07
adding const fn with_a8() to Color

flxzt:
similar to `with_alpha()`, which unfortunately can't currently be `const fn` due to floating point arithmetic's, I am suggesting to add `const fn with_a8()`.

My personal use case for it: I have a palette of predefined colors (think Gnome Colors), which I use as the basis for other default colors definitions (which is why I need 'const' ).

Related: [the discussion](https://github.com/linebender/piet/pull/524) that Color probably won't be anything else than u32 so the `const` is unlikely to become a problem in the future.

With the same thought, `as_rgba_u32()` can then also converted to be const.

I could also add analogous methods `with_r8()`, `with_g8()` and `with_b8()` to the PR for completeness sake, if the idea is approved.

### DIFF
--- a/piet/src/color.rs
+++ b/piet/src/color.rs
@@ -187,30 +187,30 @@ impl Color {
         Color::from_rgba32_u32((self.as_rgba_u32() & !0xff) | a)
     }
 
-    /// Change just the red value of a color
+    /// Change just the red value of a color.
     ///
-    /// The `r` value represents red as a u8 from 0 to 255.
+    /// The `r` value represents red as a `u8` from 0 to 255.
     pub const fn with_r8(self, r: u8) -> Color {
         Color::from_rgba32_u32((self.as_rgba_u32() & !0xff000000) | (r as u32) << 24)
     }
 
-    /// Change just the green value of a color
+    /// Change just the green value of a color.
     ///
-    /// The `g` value represents green as a u8 from 0 to 255.
+    /// The `g` value represents green as a `u8` from 0 to 255.
     pub const fn with_g8(self, g: u8) -> Color {
         Color::from_rgba32_u32((self.as_rgba_u32() & !0xff0000) | (g as u32) << 16)
     }
 
-    /// Change just the blue value of a color
+    /// Change just the blue value of a color.
     ///
-    /// The `b` value represents blue as a u8 from 0 to 255.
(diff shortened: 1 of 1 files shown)


xStrom:
That discussion is about `Copy` and `f64` implements `Copy`. If floats cause issues for `const` then this here is a different decision.

The code itself looks good.

flxzt:
Sorry, poorly worded. What I meant: a more complex Color could be an issue when introducing const fn's.

piet/src/color.rs
```diff
@@ -187,8 +187,36 @@ impl Color {
...
+    pub const fn with_b8(self, b: u8) -> Color {
+        Color::from_rgba32_u32((self.as_rgba_u32() & !0xff00) | (b as u32) << 8)
+    }
+
+    /// Change just the alpha value of a color
+    ///
+    /// The `a` value represents alpha as a u8 from 0 to 255.
+    pub const fn with_a8(self, a: u8) -> Color {
+        Color::from_rgba32_u32((self.as_rgba_u32() & !0xff) | a as u32)
+    }
```
xStrom, line 216: Minor style changes for the docs.

```suggestion
    /// Change just the red value of a color.
    ///
    /// The `r` value represents red as a `u8` from 0 to 255.
    pub const fn with_r8(self, r: u8) -> Color {
        Color::from_rgba32_u32((self.as_rgba_u32() & !0xff000000) | (r as u32) << 24)
    }

    /// Change just the green value of a color.
    ///
    /// The `g` value represents green as a `u8` from 0 to 255.
    pub const fn with_g8(self, g: u8) -> Color {
        Color::from_rgba32_u32((self.as_rgba_u32() & !0xff0000) | (g as u32) << 16)
    }

    /// Change just the blue value of a color.
    ///
    /// The `b` value represents blue as a `u8` from 0 to 255.
    pub const fn with_b8(self, b: u8) -> Color {
        Color::from_rgba32_u32((self.as_rgba_u32() & !0xff00) | (b as u32) << 8)
    }

    /// Change just the alpha value of a color.
    ///
    /// The `a` value represents alpha as a `u8` from 0 to 255.
    pub const fn with_a8(self, a: u8) -> Color {
        Color::from_rgba32_u32((self.as_rgba_u32() & !0xff) | a as u32)
    }
```

flxzt: done!

Issue #481, opened 2021-12-24
`Wrong size` error when rendering PNG via CoreGraphics


lily-mara:
I'm trying to replicate the results of the example file in `piet-coregraphics`. I'm unable to render an empty PNG file. 

```
use eyre::{Context, Result};
use piet::kurbo::Size;
use std::fs::File;
use std::io::BufWriter;

use core_graphics::color_space::CGColorSpace;
use core_graphics::context::CGContext;

use piet::{Color, RenderContext};
use piet_coregraphics::CoreGraphicsContext;

fn main() -> Result<()> {
    let size = Size {
        width: 1.0,
        height: 1.0,
    };

    let mut cg_ctx = make_cg_ctx(size);
    let mut piet_context = CoreGraphicsContext::new_y_up(&mut cg_ctx, size.height, None);

    draw(&mut piet_context).wrap_err("failed to draw file")?;

    piet_context
        .finish()
        .map_err(|e| eyre::eyre!("{}", e))
        .wrap_err("failed to finalize piet context")?;

    std::mem::drop(piet_context);
    let mut data = cg_ctx.data();
    let file = File::create("out.png").wrap_err("failed to create output file")?;
    let w = BufWriter::new(file);

    dbg!(data.len());

    let mut encoder = png::Encoder::new(w, dbg!(size.width as u32), dbg!(size.height as u32));
    encoder.set_color(png::ColorType::Rgba);
    encoder.set_depth(png::BitDepth::Eight);
    let mut writer = encoder
        .write_header()
        .wrap_err("failed to write png header")?;

    piet_coregraphics::unpremultiply_rgba(&mut data);
    writer
        .write_image_data(&data)
        .wrap_err("failed to write image data")?;

    Ok(())
}

fn make_cg_ctx(size: Size) -> CGContext {
    dbg!(&size);
    let cg_ctx = CGContext::create_bitmap_context(
        None,
        size.width as usize,
        size.height as usize,
        8,
        0,
        &CGColorSpace::create_device_rgb(),
        core_graphics::base::kCGImageAlphaPremultipliedLast,
    );

    cg_ctx
}

fn draw(rc: &mut impl RenderContext) -> Result<()> {
    rc.clear(None, Color::WHITE);

    Ok(())
}
```

I'm using these crates:

```
core-graphics = "0.22.3"
eyre = "0.6.5"
piet = "0.5.0"
piet-coregraphics = "0.5.0"
png = "0.17.2"
```

Running the above example, I get the following output:

[...]

PR #526, opened 2022-07-16
Modify capture_image_area to work with druid direct2d surfaces

longmathemagician:
Using capture_image_area on a druid paint context currently fails on windows/direct2d for the following reasons:

1) Druid paint contexts have layer masks applied to the render context, which causes the call to [ID2D1Bitmap::CopyFromRenderTarget](https://docs.microsoft.com/en-us/windows/win32/api/d2d1/nf-d2d1-id2d1bitmap-copyfromrendertarget) to fail.

2) Druid's internal surface format is BGRA and piet_direct2d::D2DRenderContext::create_blank_bitmap hardcoded RGBA, which also causes the call to ID2D1Bitmap::CopyFromRenderTarget to fail.

This PR fixes the above two issues so that capture_image_area works inside druid widget paint loops. The caveat is that as the returned image format now matches the druid surface format, you can receive captured images which are not RGBA. That's not an issue for my uses as I'm drawing the captured image back to the paint context, but it probably could be an issue for some.

### DIFF
--- a/piet-direct2d/src/d2d.rs
+++ b/piet-direct2d/src/d2d.rs
@@ -784,7 +784,7 @@ impl DeviceContext {
         &mut self,
         width: usize,
         height: usize,
-        alpha_mode: D2D1_ALPHA_MODE,
+        dpi_scale: f32,
     ) -> Result<Bitmap, Error> {
         // Maybe using TryInto would be more Rust-like.
         // Note: value is set so that multiplying by 4 (for pitch) is valid.
@@ -794,18 +794,17 @@ impl DeviceContext {
             width: width as u32,
             height: height as u32,
         };
-        let format = D2D1_PIXEL_FORMAT {
-            format: DXGI_FORMAT_R8G8B8A8_UNORM,
-            alphaMode: alpha_mode,
-        };
-        let props = D2D1_BITMAP_PROPERTIES1 {
-            pixelFormat: format,
-            dpiX: 96.0,
-            dpiY: 96.0,
-            bitmapOptions: D2D1_BITMAP_OPTIONS_NONE,
-            colorContext: null_mut(),
-        };
+
         unsafe {
+            let pixel_format = self.0.GetPixelFormat();
+            let props = D2D1_BITMAP_PROPERTIES1 {
+                pixelFormat: pixel_format,
+                dpiX: dpi_scale * 96.0,
+                dpiY: dpi_scale * 96.0,
+                bitmapOptions: D2D1_BITMAP_OPTIONS_TARGET,
+                colorContext: null_mut(),
+            };
+
             let mut ptr = null_mut();
             let hr = self
                 .0
--- a/piet-direct2d/src/lib.rs
+++ b/piet-direct2d/src/lib.rs
@@ -481,14 +481,25 @@ impl<'a> RenderContext for D2DRenderContext<'a> {
         let mut target_bitmap = self.rt.create_blank_bitmap(
             device_size.width as usize,
             device_size.height as usize,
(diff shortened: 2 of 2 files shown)


cmyr:
Sorry about the delay here!

I don't currently have access to a windows machine, but the linked docs do clearly state that you need to pop clips before calling this method. I am slightly worried, however, that this could have unexpected consequences? To my eye, it feels like the intention of this method should be to capture the current state of the context, including any masking? Although maybe I'm misunderstanding, if removing the clip layers doesn't in any way change the visual appearance, then my concern is irrelevant.

longmathemagician:
Hey no worries timing wise, my forks of druid and piet & PRs are a bit of a mess at the moment anyway since I've primarily been focused on supporting a single application.

Regarding your concern, that's certainly valid. My use of masks in piet is minimal so I haven't noticed any issues with this, but I can throw together a test case later and see how it plays out.

Unfortunately I'm not sure if there's a simple way to capture a region of a dxgi surface render target without popping off the layers. I suspect an alternate method would involve duplicating the surface's command list (which is problematic in its own ways). If you (or anyone else 'round here) have an alternate path that looks viable I'm happy to try implementing it, but I'm not familiar enough with direct2d to have any concrete fix for this in mind.

longmathemagician:
@cmyr Sorry for the delay. I finally had time to look into this a bit more. It does capture the entire (unclipped) canvas, e.g. if you clip to a centered circle and capture the canvas you receive a bitmap of the entire canvas and not one with just the centered circle filled in and the rest transparent. This `capture_image_area` captures the canvas exactly as it is rendered on screen (at least in all of the tests I did), with all layers essentially flattened.

IMO this is currently fine for handling a `PietImage`\*. When you want just a portion of the captured image, clip (or just leave clipped, this restores existing masks) the context before the draw call and it works exactly as expected. This is consistent with how `capture_image_area` works in `piet-cairo` and `piet-coregraphics`<sup>†</sup>.

But, if what this _should_ be doing is instead capturing only the intersection of the union of layer masks and the provided source rect then it'll need to be fixed.

Doing that may or may not be tricky. A potential approach is using [ID2D1DeviceContext::GetTarget](https://learn.microsoft.com/en-us/windows/win32/api/d2d1_1/nf-d2d1_1-id2d1devicecontext-gettarget) to get an `ID2D1Image` (rather than an `ID2D1Bitmap` from [ID2D1Bitmap::CopyFromRenderTarget](https://learn.microsoft.com/en-us/windows/win32/api/d2d1/nf-d2d1-id2d1bitmap-copyfromrendertarget)) and then drawing it to the `target_bitmap`. Another possibility is to apply the cached layers+masks to either the `target_bitmap`'s rendertarget or an intermediate to ensure that only the desired region is painted. Either way it will probably incur an additional performance penalty.

Both of those approaches ended in failure in the short amount of time I had to try them. They're probably viable paths though, either if I spend more time troubleshooting or if someone who is better with Direct2D wants to take over.

\* There is a conversation to be had concerning handling of a `PietImage` with regard to exporting it out of the active context or encoding them to some standard image format. Currently this doesn't really work, the image ends up in one of a variety of pixel formats depending on the platform and how you created your context (i.e. whether you're using druid+piet or just piet), not to mention that IIRC an `ID2D1Bitmap` lives in GPU memory. From #481 I take it this is not a high priority task.

<sup>†</sup> `capture_image_area` in `piet-coregraphics` under druid is problematic, linebender/druid#2246 makes it functional but it's still pretty bad.

longmathemagician:
@jneem unless @cmyr disagrees this can be merged as well.

cmyr:
I think this is probably an improvement! 💁

PR #527, opened 2022-07-26
Blur methods for PietImages

longmathemagician:
This is a draft implementation of a quick-ish blur effect. It was written to support overlay UI elements in druid, but could in theory be used anywhere you want to blur a PietImage. As an example, this is what I'm using it for:

Example image showing the a blurred toolbar

Working backends & notes:
- Direct2D (hardware accelerated). The device context should probably be cached instead of recreated each time the method is called, and it would be nice if the unsafe block wasn't so large.
- Cairo (software implementation, very approximate dual-kawase-inspired algorithm). Not as efficient as it could be, currently creates a duplicate surface and render context to get around Cairo's reference count limit.

To do:
- CoreGraphics backend implementation
- SVG backend implementation
- Web backend implementation
- Snapshots


derekdreery:
Note to self: this should also be possible for piet-svg using filters. I should look at this if/when this PR lands.

PR #528, opened 2022-07-29
Improve piet-coregraphics

longmathemagician:
CoreGraphicsContext::draw_image_area() doesn't account for coregraphics' reversed y-axis, 086e8e613dfc951f869b6f6ebb9ebcba012ab25a fixes that by passing the cropped image to CoreGraphicsContext::draw_image() which does draw correctly.

### DIFF
--- a/piet-coregraphics/src/lib.rs
+++ b/piet-coregraphics/src/lib.rs
@@ -382,8 +382,7 @@ impl<'a> RenderContext for CoreGraphicsContext<'a> {
     ) {
         if let CoreGraphicsImage::NonEmpty(image) = image {
             if let Some(cropped) = image.cropped(to_cgrect(src_rect)) {
-                // TODO: apply interpolation mode
-                self.ctx.draw_image(to_cgrect(dst_rect), &cropped);
+               self.draw_image(&CoreGraphicsImage::NonEmpty(cropped), dst_rect, _interp);
             }
         }
     }
--- a/piet-coregraphics/src/lib.rs
+++ b/piet-coregraphics/src/lib.rs
@@ -48,6 +48,7 @@ pub struct CoreGraphicsContext<'a> {
     // by CTContextGetCTM. Instead we maintain our own stack, which will contain
     // only those transforms applied by us.
     transform_stack: Vec<Affine>,
+    y_down: bool,
 }
 
 impl<'a> CoreGraphicsContext<'a> {
@@ -64,7 +65,7 @@ impl<'a> CoreGraphicsContext<'a> {
         height: f64,
         text: Option<CoreGraphicsText>,
     ) -> CoreGraphicsContext {
-        Self::new_impl(ctx, Some(height), text)
+        Self::new_impl(ctx, Some(height), text, false)
     }
 
     /// Create a new context with the y-origin at the bottom right corner.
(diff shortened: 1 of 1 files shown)


xStrom:
Please rebase on `master` to fix clippy errors and then also run `cargo fmt`.

piet-coregraphics/src/lib.rs
```diff
@@ -382,8 +382,7 @@ impl<'a> RenderContext for CoreGraphicsContext<'a> {
     ) {
         if let CoreGraphicsImage::NonEmpty(image) = image {
             if let Some(cropped) = image.cropped(to_cgrect(src_rect)) {
-                // TODO: apply interpolation mode
-                self.ctx.draw_image(to_cgrect(dst_rect), &cropped);
+                self.draw_image(&CoreGraphicsImage::NonEmpty(cropped), dst_rect, _interp);
```
xStrom, line 385: Sorry I missed this earlier, but please rename the `_interp` variable to `interp`. The underscore means that it is not used, but that is no longer true.

After that you can mark this PR as ready and I'll approve it. Thanks for your patience!

longmathemagician: Quite right, thank you. Fixed in 4f7dc7115e0ae71abf9312a9285ff880b76c1d57.

I had originally hoped to address linebender/druid#2245 with this branch as well but it seems that'll have to be done within druid so I'll mark this PR as ready now. Thanks for getting it merged!

longmathemagician:
Added one last (hopefully) commit to this branch: 5df4eec2dbfe26fdad0c675c489210e33daea782 fixes capture_image_area for y-up contexts by tracking the y-direction in CoreGraphicsContext and translating the context's origin if necessary.

xStrom:
I haven't looked into it deeper yet, but the test image is definitely broken now.

coregraphics-test-16-1 00
.

longmathemagician:
Dang, this actually broke a lot of stuff that wasn't in my test case. Getting à la cart piet snapshot testing set up locally and then harmonizing y-direction handling across druid with and without bitmap caching (linebender/druid#2246) may take me a day or two.

If you want I can just shelve the commit, it's not likely to help anyone who isn't also using bitmap render caching of some form in druid.

xStrom:
No rush, we can see what else comes up.

longmathemagician:
Oops, I was using [`as_ref`](https://github.com/linebender/piet/runs/7910660375?check_suite_focus=true) in another branch and didn't catch that with clippy. Some form of unwrapper is needed for sharing assets across contexts with differing y-directions, so I just made `as_ref` public and axed `pub fn copy_image` since it didn't actually copy yet (a weirdly difficult thing for coregraphics it seems).

For the snapshots, I don't know if the standard here is to have more comprehensive tests or faster/simpler ones. If simpler is better I can just drop be1da570da12fca60bb23b4463d5cf0401e31805.

longmathemagician:
Pulled the changes to test image 16, I'll throw that in a separate PR. Took care of the last issue I had with coregraphics in f5abbb15391aa585440f6c5c0725c2251a4f432f, as long as I didn't break anything else in the process this branch is good to go.

jneem:
@longmathemagician can I go ahead and merge this? Or is it waiting on the updated test image 16?

longmathemagician:
@jneem you're good to merge it, thank you. There are still a lot of issues with piet-coregraphics I never got around to tackling but this is at least an improvement over the current state.

PR #530, opened 2022-08-19
Optimized set_stroke for piet_web

Cl00e9ment:
I implemented the  `TODO(performance)`.

It's hard to tell if there is really a big performance increase, but there is definitely less calls to the Canvas API.


cmyr (approved):
Thanks, this looks reasonable :)

PR #536, opened 2022-10-28
Improve text position in SVG.

derekdreery:
Because SVG uses the baseline to position text, while we use the
top-left of the text, we need to convert between the two.

Before this PR, the whole font height was used, which pushes the text
down too far.

After this PR, we tell SVG to use the top of the ascender as our anchor
point (`dominant-baseline: hanging`), with an additional fiddle factor
of 0.04 * font height because we actually want the top of the line, not
the ascender. The fiddle factor was worked out by eye looking at the
default sans-serif font, comparing the svg to a png generated by
piet-cairo.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -31,8 +31,6 @@ pub struct CairoRenderContext<'a> {
     error: Result<(), cairo::Error>,
 }
 
-impl<'a> CairoRenderContext<'a> {}
-
 #[derive(Clone)]
 pub enum Brush {
     Solid(u32),
--- a/piet-svg/src/lib.rs
+++ b/piet-svg/src/lib.rs
@@ -272,9 +272,17 @@ impl piet::RenderContext for RenderContext {
             .unwrap()
             .insert(layout.font_face.clone());
 
+        // We use the top of the text for y position, but SVG uses baseline, so we need to convert
+        // between the two.
+        //
+        // `dominant-baseline` gets us most of the way (to the top of the ascender), so we add a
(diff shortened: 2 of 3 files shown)


derekdreery:
OK I take what I said back - there is a `dominant-baseline` attribute that makes the text use the top-left as the draw point. Updated the PR to use this.

piet-cairo/src/lib.rs
```diff
@@ -31,8 +31,6 @@ pub struct CairoRenderContext<'a> {
     error: Result<(), cairo::Error>,
 }
 
-impl<'a> CairoRenderContext<'a> {}
```
raphlinus, line 34: This seems unrelated to the issue at hand but seems like a reasonable cleanup.

piet-svg/src/lib.rs
```diff
@@ -272,9 +272,17 @@ impl piet::RenderContext for RenderContext {
             .unwrap()
             .insert(layout.font_face.clone());
 
+        // We use the top of the text for y position, but SVG uses baseline, so we need to convert
+        // between the two.
+        //
+        // `dominant-baseline` gets us most of the way (to the top of the ascender), so we add a
+        // small fiddle factor in to cover the difference between the top of the line and the top
+        // of the ascender (currently 6% of the font height, calcuated by eye).
+        let y = pos.y + 0.06 * layout.size().height;
```
raphlinus, line 281: I generally find that fudge factors like this are a sign that the implementation isn't *quite* doing the right thing. That said, I had a skim through SVG docs and don't see any better way to do this, so I'll approve.

raphlinus (approved):

PR #518, opened 2022-06-15
Prepeare for releasing v0.6.0

xStrom:
This PR does two things:

#### Update dependencies explicitly to the latest compatible versions

These are all semver compatible changes. These are versions that Cargo resolves to anyway, which also means they're the versions we actually test with. For the curious I've written more about the rationale for doing this in the Druid [CONTRIBUTING.md](https://github.com/linebender/druid/blob/master/CONTRIBUTING.md#dependencies).

#### Increase Piet version to v0.6.0

Straightforward, with the only question being if it should be v0.5.1 or v0.6.0. When I went over all the changes it seemed like the choice of v0.6.0 was clear. One example is #494 which updated `image` from `0.23` to `0.24`. Some of the Piet API exposes this crate and so by extension the Piet API changed.

PR #540, opened 2022-11-23
Bump version to 0.6.0

jneem:
This is a rebase of #518, bumping all the versions to 0.6.0.

### DIFF
--- a/piet/Cargo.toml
+++ b/piet/Cargo.toml
@@ -12,12 +12,12 @@ categories = ["rendering::graphics-api"]
 include = ["src/**/*", "Cargo.toml", "snapshots/resources/*"]
 
 [dependencies]
-image = { version = "0.24.0", optional = true, default-features = false }
-kurbo = "0.8.2"
-pico-args = { version = "0.4.0", optional = true }
-png = { version = "0.17", optional = true }
-os_info = { version = "3.0.0", optional = true, default-features = false }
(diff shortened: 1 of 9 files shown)


piet/Cargo.toml
```diff
@@ -12,12 +12,12 @@ categories = ["rendering::graphics-api"]
...
 
 [dependencies]
-image = { version = "0.24.0", optional = true, default-features = false }
-kurbo = "0.8.2"
-pico-args = { version = "0.4.0", optional = true }
-png = { version = "0.17", optional = true }
-os_info = { version = "3.0.0", optional = true, default-features = false }
-unic-bidi = "0.9"
+image = { version = "0.24.2", optional = true, default-features = false }
+kurbo = "0.8.3"
```
raphlinus, line 16: Do we want to pull off the bandaid and jump to 0.9 now that's been released? I don't *think* there are any client changes that will be impacted, and I confirm the example compiles and runs without modification.

jneem: May as well give it a try. I guess it has the downside of requiring a recent `rustc` but I don't think we've ever cared about supporting old compilers, right?

cmyr: I think we're happy to be aggressive about msrv.

cmyr (approved):
Looks good!

PR #542, opened 2022-12-08
Impl Default for piet-svg Text and remove `new` method

matthunz:
This removes the need for `#[allow(clippy::new_without_default)]`

### DIFF
--- a/piet-svg/src/text.rs
+++ b/piet-svg/src/text.rs
@@ -32,10 +32,9 @@ pub struct Text {
     pub(crate) seen_fonts: Arc<Mutex<HashSet<FontFace>>>,
 }
 
-impl Text {
-    #[allow(clippy::new_without_default)]
-    pub fn new() -> Self {
-        Text {
+impl Default for Text {
+    fn default() -> Self {
+        Self {
             source: Arc::new(Mutex::new(MultiSource::from_sources(vec![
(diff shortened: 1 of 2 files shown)


jneem:
Thanks for the PR! I don't have a strong opinion about the presence of `default` vs `new`, although I wonder if whoever added that `allow` option had a reason for it...

Given that removing `new` is a breaking change, though, I'd advocate for keeping it there and just adding the `default`.

cmyr:
I don't actually remember the rationale for wanting to avoid default, here, but in any case we definitely want to keep the `new` method, although I do not see any harm in also adding an impl of `Default`.

matthunz:
Makes sense to me! I added `new` back that just uses the default implentation in [cf03df5](https://github.com/linebender/piet/pull/542/commits/cf03df5643cf9691b9df205ff9ef0d6255cf8858)

piet-svg/src/text.rs
```diff
@@ -32,17 +32,22 @@ pub struct Text {
...
                 Box::new(MemSource::empty()),
             ]))),
             seen_fonts: Arc::new(Mutex::new(HashSet::new())),
         }
     }
+}
+
+impl Text {
+    pub fn new() -> Self {
+        Seld::default()
```
xStrom, line 49: `Seld` -> `Self`

PR #546, opened 2023-01-14
Fix `docs.rs` generation by building on Windows for D2D and on Mac for Coregraphics

benthillerkus:
Adds [`docs.rs` metadata](https://docs.rs/about/metadata) instructions.

I haven't tested if this is enough to actually build, but considering that [`winapi`](https://docs.rs/crate/winapi/latest/source/Cargo.toml), `win-rs` and [`core-x`](https://docs.rs/crate/core-foundation/0.9.3/source/Cargo.toml) are doing this, it should work.


jneem (approved):
Thanks!

benthillerkus:
Yeah, I think the only way to update the docs is to make a release.

I published my fork and it seems to work: [docs.rs/definitely-not-piet-direct2d-please-ignore](https://docs.rs/definitely-not-piet-direct2d-please-ignore/0.6.0/definitely_not_piet_direct2d_please_ignore/index.html).

PR #550, opened 2023-01-30
Re-export the `image` crate and prepare v0.6.2.

xStrom:
Druid v0.8 ended up shipping with a different major version of `image` than Piet, which causes severe issues due to their incompatibility. The ultimate solution to this is to just have Piet re-export the `image` crate and `druid-shell` depend on that. Synchronization bliss!

This PR re-exports the `image` crate and bumps the version to v0.6.2.

### DIFF
--- a/piet-common/src/lib.rs
+++ b/piet-common/src/lib.rs
@@ -2,7 +2,7 @@
 //!
 //! This crate reexports the [piet crate][piet], alongside an appropriate backend
 //! for the given platform. It also exposes [kurbo], which defines shape and
-//! curve types useful in drawing.
+//! curve types useful in drawing, and [image] for image manipulation.
 //!
 //! The intention of this crate is to provide a single dependency that handles
 //! the common piet use-case. If you have more complicated needs (such as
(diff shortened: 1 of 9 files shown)


piet-common/src/lib.rs
```diff
@@ -31,6 +32,10 @@ pub use piet::*;
 #[doc(hidden)]
 pub use piet::kurbo;
 
+#[doc(hidden)]
+#[cfg(feature = "image")]
+pub use piet::image_crate;
+
```
jneem, line 38: I see this is following the example of kurbo above, but why have the re-exports be hidden? IIUC the point of the re-export is so that downstream crates can easily use the re-export instead of bringing in their own version, but then don't we want downstream users to know that the re-export is available?

jneem (approved):

piet-common/src/lib.rs
xStrom: I agree with your thinking and thought the same. The reason that I ended up leaving it was that not having the `#[doc(hidden)]` actually causes it to be documented twice. This removes it from the "Re-exports" section, but leaves it in the modules section which is two lines below.

Now that you brought this up again I looked more closely, and it seems that it's not just the documentation that is double. The actual export seems to be redundant too.

```rust
pub use piet::*; // This line exposes piet::image (and kurbo) as modules.
```

I now removed these lines are everything that depends on them seems to keep compiling.

PR #554, opened 2023-03-10
Add `make_image_with_stride`

### LINKED ISSUE #553: More flexible `RenderContext::make_image`
I am trying to create a druid widget that renders a 3d scene using WGPU. To do this, I render the scene into a texture, get the data as a `&[u8]` and paint it using piet's [`draw_image`](https://docs.rs/piet/0.6.2/piet/trait.RenderContext.html#tymethod.draw_image).  
Unfortunately, each row in the texture has a stride with an alignment of [`COPY_BYTES_PER_ROW_ALIGNMENT`](https://docs.rs/wgpu/0.15.1/wgpu/constant.COPY_BYTES_PER_ROW_ALIGNMENT.html) bytes so, in order to use it with [`make_image`](https://docs.rs/piet/0.6.2/piet/trait.RenderContext.html#tymethod.make_image), I have to convert it first. This is quite annoying and feels unnecessary.

I think this situation could be common in other situations such as showing images/videos from raw data. Maybe it could be useful to create another function, for example `RenderContext::make_image_flat` that accepts a row pitch, just like [`image::FlatSamples`](https://docs.rs/image/latest/image/flat/struct.FlatSamples.html) that stores the layout information in [`SampleLayout`](https://docs.rs/image/latest/image/flat/struct.SampleLayout.html)

LucaCiucci:
The `make_image_with_stride` function will allow to create images from a buffer with a specific row stride. See #553 for more info.

Progress:
 - [x] add `make_image_with_stride` to `RenderContext`
 - [x] add documentation
 - [x] implement for `NullRenderContext`
 - [x] implement for `piet-cairo`
 - [x] implement for `piet-coregraphics`
 - [x] implement for `piet-direct2d`
 - [x] implement for `piet-svg`
 - [x] implement for `piet-web`
 - [x] check use of ~`std::ptr::copy_nonoverlapping`~ `copy_from_slice` when possible
 - [x] size checks
 - [x] **more tests**?

Resolves #553.

### DIFF
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -238,7 +238,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         // Confident no borrow errors because we just created it.
         let image_stride = image.stride() as usize;
         {
-            if buff.len() < piet::util::expected_image_buffer_size(image_stride, height, stride) {
+            if buf.len() < piet::util::expected_image_buffer_size(image_stride, height, stride) {
                 return Err(Error::InvalidInput);
             }
 
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -238,7 +238,7 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         // Confident no borrow errors because we just created it.
         let image_stride = image.stride() as usize;
         {
-            if buf.len() < piet::util::expected_image_buffer_size(image_stride, height, stride) {
+            if buf.len() < piet::util::expected_image_buffer_size(width * format.bytes_per_pixel(), height, stride) {
                 return Err(Error::InvalidInput);
             }
 
--- a/piet-cairo/src/lib.rs
+++ b/piet-cairo/src/lib.rs
@@ -238,7 +238,13 @@ impl<'a> RenderContext for CairoRenderContext<'a> {
         // Confident no borrow errors because we just created it.
         let image_stride = image.stride() as usize;
         {
-            if buf.len() < piet::util::expected_image_buffer_size(width * format.bytes_per_pixel(), height, stride) {
+            if buf.len()
+                < piet::util::expected_image_buffer_size(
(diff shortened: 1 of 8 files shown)


LucaCiucci:
I think the implementation is at a good point so I'm marking it as "ready for review". Some clippy are failing but I think this is related to #555.

xStrom:
Yes #555 resolved CI errors in `master` and is now merged. You can now rebase your fork on `master` to get those fixes here too.

LucaCiucci:
Done, I also corrected a mistake I made on cairo. Unfortunately, I cannot test piet-cairo and piet-coregraphics on my system.

LucaCiucci:
I still have to implement some checks but I'm unsure about what kind of error to return when the given `buf` is smaller than `bytest_per_pixel * stride * height`. Maybe `InvalidInput` or `BackendError`?

xStrom:
I think `InvalidInput` is a good choice.

LucaCiucci:
I think I've done. Let me know if you are satisfied with the changes, especially with the fact that `make_image` is now a provided method.

xStrom:
Looks like that error is now being triggered.

```
Running `target/debug/examples/test-picture --all --scale=1 --out=cairo_samples --compare=./piet/snapshots/cairo`
error in sample 2: 'Invalid input'
error generating sample: Invalid input
```

I have't looked deeper to see what's going on.

LucaCiucci:
67af29b should fix this, but I cannot test it at the moment. Can you run the workflows again?

xStrom (commented):
Looks good in general. Just a few nits for which I left inline comments.

Also, I think there is room for optimizations here by piping the stride info all the way to the platform APIs instead of creating the temporary tightly packed buffers. However that doesn't need to be part of this PR as it might not be trivial, depending on which strides the platform APIs will accept due to alignment. So I'm just making a note of it here, no need to do it.

xStrom (approved):
Great, thanks for your effort on this!

PR #557, opened 2023-03-31
feat: Add a feature that allows piet to be used in no_std environments

notgull:
As [below](https://github.com/linebender/kurbo/pull/242), so above.

The main hiccup here is that I can't use `Box<dyn std::error::Error>` on `no_std`, so I just make it so the `BackendError` variant of `Error` isn't there on `no_std`. There is probably a more elegant solution to this problem.


xStrom:
I haven't looked at the changes in any depth, but the CI failures should be resolved.

notgull:
I can't replicate the failures locally

xStrom:
Is this still something you would like to see merged?

If so, then the current problem is that the CI seems to be stuck in 'waiting'. I think the best approach here is for you to rebase on `master` and force push to your branch to re-trigger the CI for this PR. Thanks!

notgull:
I still want this, it's just that I've been busy with more pressing issues.

notgull:
I have no idea why the piet-core-graphics implementation is failing

ratmice:
@notgull #562 has a fix for the macos failures, but also bumps some dependencies.

I got that other failure once about local connections, but another time it went green so I'm not sure if that check is spurious or why exactly it finally went green for me.

xStrom:
Alright, some significant time has passed now. The macOS dependencies have been updated and we have a new ARM based macOS runner for CI. The `master` branch is passing CI nicely right now.

It would be nice to get this landed. Please rebase on `master` and let's see if we can get the CI to pass for this PR. 🤞

DJMcNab:
[`core::error::Error`](https://doc.rust-lang.org/stable/core/error/trait.Error.html) is stable as of early Septermber. In current Linebender practise, this would be a sufficient reason to raise the MSRV. Not saying that's the right solution for piet as-is.

PR #558, opened 2023-06-20
Make samples_main take an Fn trait

notgull:
I'm currently using a [hacky workaround](https://github.com/notgull/piet-hardware/blob/9bb5e1cdf58bdc8ad47fd7ea3906fbc72da9a034/crates/piet-wgpu/examples/test_picture_wgpu.rs#L46-L69) in `piet-hardware` to test samples, and it doesn't look like there is a reason why it's an `fn` pointer instead of an `Fn` trait?


derekdreery (approved):
LGTM

PR #562, opened 2023-12-08
Update kurbo and svg dependencies, fix a clippy warning.

ratmice:
This updates to the latest kurbo, I could have a version which is compatible with both piet and vello... 

* Svg had a future incompatibility warning, upgrade to fix that.
* Clippy noticed that `font_kit::sources::MultiSource` is `!Send !Sync`, So replace `Arc<Mutex<MultiSource>` with `Rc<RefCell<_>>`.

Not sure if this last bit is a wanted change, or if it'd be preferred to just silence the warning, or ignore it entirely?

### DIFF
--- a/piet-coregraphics/src/text.rs
+++ b/piet-coregraphics/src/text.rs
@@ -1,10 +1,10 @@
 //! Text related stuff for the coregraphics backend
 
+use std::cell::RefCell;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::ops::{DerefMut, Range, RangeBounds};
 use std::rc::Rc;
-use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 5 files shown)


jneem (approved):
I'm not sure whether we *want* `MultiSource` to be `Send` or `Sync`, but it appears to be `!Send` and `!Sync` on all platforms, so the change here should have no visible effect.

This does seem to be a breaking change, since the re-export is public.

ratmice:
Yeah, perhaps it is better if I drop the version bumps to avoid the breaking changes?
I didn't really end up using the library I wanted this version bump for, but it'd be good to get the lints updated regardless.

jneem:
I don't think it's so bad to have the version bump, just as long as we're aware that it needs one. I think it makes sense to have a version of piet that's compatible with the latest version of kurbo, after all.

PR #569, opened 2024-04-16
fix: allow drawing after copying raw pixels

### LINKED ISSUE #568: Can't resume drawing after calling `BitmapTarget::copy_raw_pixels`.
On Windows specifically, it's not possible to make further draws to a BitmapTarget after calling `BitmapTarget::copy_raw_pixels`. I believe this is because `self.context.end_draw()?;` is called during aforementioned method, without a corresponding `begin_draw()` call. It fails with an error code of `BackendError(hresult 88990001)`.

Example:
```rs
use piet::{kurbo::Rect, Color, ImageFormat};
use piet_common::{Device, RenderContext};

fn main() {
    let mut piet = Device::new().unwrap();
    let mut bitmap = piet.bitmap_target(256, 256, 1.0).unwrap();
    {
        let mut rc = bitmap.render_context();
        let rect = Rect::new(10.0, 10.0, 100.0, 100.0);
        rc.fill(rect, &Color::RED);
        rc.finish().unwrap();
    }
    let mut data = vec![0; 256 * 256 * 4];
    bitmap
        .copy_raw_pixels(ImageFormat::RgbaPremul, &mut data)
        .unwrap();

    {
        let mut rc = bitmap.render_context();
        let rect = Rect::new(50.0, 50.0, 150.0, 150.0);
        rc.fill(rect, &Color::GREEN); // ERROR!
        rc.finish().unwrap();
    }
}
```

doinkythederp:
Fixes #568

### DIFF
--- a/piet-common/src/direct2d_back.rs
+++ b/piet-common/src/direct2d_back.rs
@@ -207,6 +207,8 @@ impl<'a> BitmapTarget<'a> {
                 std::ptr::copy_nonoverlapping(src, dst, self.width * 4);
             }
         }
+
+        self.context.begin_draw();
         Ok(size)
     }
 


xStrom (approved):
Thanks!

Not having a matching `begin_draw` definitely causes the described issue and this PR solves it nicely for the happy path.

We will still have the same issue if an error happens and we don't reach the end of the method where the `begin_draw` is called. However resolving those edge cases can be follow-up work, for which I will submit a PR myself shortly.

PR #572, opened 2024-09-15
Changing 'alloc' from piet::StrokeDash use Arc<[f64]> instead of Rc<[f64]> 

anesthetice:
Changes alloc in piet::StrokeDash use Arc instead of Rc, allowing structs containing piet::StrokeStyle to implement Send + Sync


xStrom:
The `master` branch has been updated to get the CI passing again. So please rebase on `master` so that we can get the CI to pass for this PR as well.

Also, how useful is the feature? Is there still a need for the `Rc` version? I'm thinking that perhaps we could just have only `Arc` and skip the feature.

xStrom:
I understand the value of the `Arc` version. What I'm skeptical of is the `Rc` version. Do you have a need for it? Because if you don't need it, then just having `Arc` would be better. Adding a whole feature is quite complex and only makes sense when both options give a lot of value.

anesthetice:
> I understand the value of the `Arc` version. What I'm skeptical of is the `Rc` version. Do you have a need for it? Because if you don't need it, then just having `Arc` would be better. Adding a whole feature is quite complex and only makes sense when both options give a lot of value.

Not personally no, I could remove the feature flag and simply switch Rc to Arc if that works for you

xStrom:
I don't see any significant downside with it, so yeah do that.

anesthetice:
Done, just the macos-latest check seems to fail now..

xStrom:
The PR doesn't have the latest submodule from `master`. The following should do the trick:

```sh
git fetch origin
git rebase origin master
git push --force
```

That is assuming that the remote repo here is named as `origin` in your git settings.

anesthetice:
`git submodule update --remote` seems to have done the trick

xStrom (approved):
Thanks for pushing this forward!

I hope it helps with your feature implementation.

PR #583, opened 2024-10-22
Update more deps.

waywardmonkeys:
These are semver changes, but that don't require changes to the Piet source code.


waywardmonkeys:
This leaves `svg` and `base64` as the remaining out of date dependencies. The `svg` changes required are simple. The `base64` are up for anyone that wants to do them.

xStrom (approved):
Thanks!

PR #584, opened 2024-10-23
piet-svg: Update `svg`

waywardmonkeys:


waywardmonkeys:
Confirmed by looking at the output of `cargo run --example basic-svg` and seeing that text nodes are still correct.

xStrom (approved):
Let's hope it's an improvement. 🤞

PR #588, opened 2024-11-04
Update to the latest Linebender CI standard.

xStrom:
* The CI script gets a big update to bring it more in sync with the latest standard.
* `piet-common` loses the `web` feature, which was functionally broken anyway as shown by the new CI.
  The `piet-web` backend was and still is automatically chosen when targeting `wasm32`.
* Enable the `doc_auto_cfg` feature for docs.rs which will show a little tip next to feature gated functionality informing of the crate feature flag.
* Piet version is now configured in the workspace `Cargo.toml` instead of all over the place.
* `Cargo.lock` is now committed.
* Piet now has an explicit MSRV of 1.77 and it's documented.
* Each published package now has its own copy of the license files.


piet-cairo/Cargo.toml
```diff
@@ -1,17 +1,24 @@
 [package]
 name = "piet-cairo"
-version = "0.7.0"
+version.workspace = true
 authors = ["Raph Levien <raph.levien@gmail.com>"]
```
DJMcNab, line 4: We should probably move to the AUTHORS file, I presume that would be a follow-up?

piet-cairo/Cargo.toml
```diff
@@ -1,17 +1,24 @@
...
-edition = "2021"
-keywords = ["graphics", "2d"]
+description = "Cairo backend for Piet 2D graphics abstraction."
 categories = ["rendering::graphics-api"]
+keywords = ["graphics", "2d"]
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+readme = "../README.md"
```
DJMcNab, line 12: ... That's not great

DJMcNab (approved):
Happy to land this. Obviously I wasn't previously aware of this PR

piet-cairo/Cargo.toml
xStrom: Might as well do it here as it's quite straightforward. Added the commit.

PR #591, opened 2024-12-07
Simplify per-platform package exclusion in CI.

xStrom:
The per-platform exclusions were getting a bit crazy, with a lot of step duplication. This new approach uses a JSON object with dynamic access to reduce the number of steps back down to sanity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
           components: rustfmt
 
       - name: cargo fmt
-        run: cargo fmt --all --check
+        run: cargo fmt ${{ fromJson(env.EXCLUDE_PKGS)['Foobar'] }} --all --check
 
       - name: install ripgrep
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -118,29 +118,11 @@ jobs:
...
-
-      - name: cargo clippy (auxiliary) (macos)
-        run: cargo hack clippy --workspace ${{ env.NO_MACOS_PKGS }} --locked --optional-deps --each-feature --tests --benches --examples -- -D warnings
-        if: contains(matrix.os, 'macos')
-
-      - name: cargo clippy (ubuntu)
-        run: cargo hack clippy --workspace ${{ env.NO_UBUNTU_PKGS }} --locked --optional-deps --each-feature -- -D warnings
-        if: contains(matrix.os, 'ubuntu')
+      - name: cargo clippy
+        run: cargo hack clippy --workspace ${{ fromJson(env.EXCLUDE_PKGS)[runner.os] }} --locked --optional-deps --each-feature -- -D warnings
```
DJMcNab, line 122: What happens if we get a `null` from here? My first thought is that I want an error, but thinking about it further, the empty string would be better - CI would fail if we had a typo, because the package wouldn't be excluded.

DJMcNab (approved):
This seems reasonable. It's a big frustrating that we can't embed a YAML string, because embedding JSON is ugly - at least not without a (hypothetical syntax) `env[NO_ ## runner.os ## PKGS]`)

But this is definitely an improvement for this repo. I presume we'd leave it out for the platform agnostic workspaces (Vello?)

.github/workflows/ci.yml
xStrom: I tested it and referencing an undefined property converts into an empty string. So yeah, the CI would fail because by design it would only pass if there is no mistake.

xStrom:
> I presume we'd leave it out for the platform agnostic workspaces (Vello?)

Well, no. The dream is to converge on a single CI script that handles all repos.

For now I don't have plans on generalizing any of the LFS/snapshot stuff, but for everything else I'll have solutions later this week - as the prototypes already exist locally.

PR #592, opened 2024-12-21
piet-next: Use released Peniko 0.3.0

waywardmonkeys:


raphlinus (approved):
Thanks! At some point we'll need to figure out the bureaucracy for collaborating on this branch. If we think existing Piet is over, then maybe the best answer is just merge to main and work there.

PR #595, opened 2025-08-10
Bump cairo, pango versions, repo chores

flxzt:
Version bumps seems to be enough.

Changelog can be found here: https://github.com/gtk-rs/gtk-rs-core/releases


flxzt:
Ugh, fixing the clippy warning caused a chain of repo chores that need to be done.
For one, I updated the `Cargo.lock` because otherwise clippy would complain about the usage of the macro from wasm-bindgen-test which seems to be fixed in the latest versions.
Then cfg-attr demanded to bump to edition 2024 for it's "edition2024" (which seemingly breaks the promise that one's codebase edition can be independent in it's edition?).
With the update to the 2024 edition, new formatting needed to be applied.

And now new lints need to be fixed, which I'll do later..

PR #604, opened 2026-04-01
Update dependencies and stable Rust to 1.95.

xStrom:
* Stable Rust to 1.95
* `checkout` to 6
* `typos` to 1.45
* `png` to 0.18
* `kurbo` to 0.13
* other deps to their latest semver compatible versions

PR #605, opened 2026-04-09
Direct2D: enable support for colorful emojis

leksiso:
Direct2d draws emojis in black and white by default. This PR sets the flag that enables colorful emojis.
See https://learn.microsoft.com/en-us/windows/win32/directwrite/color-fonts for more information.


xStrom (approved):
Thanks!

Will merge after #604 lands, which will fix CI.

